PR #781, opened 2015-01-14
FilterGroupsByAge: Remove or set ExcludedByFilter for people whose age is unknown

dcs619:
Currently if the LoadAll flag in LoadGroups is set, people without an age/birthdate aren't filtered out.

This makes sure only people with valid ages are allowed to check into groups with an AgeRange attribute.


dcs619:
I'm able to work around this bug by setting an age restriction of 101 on our Special Needs and generic groups.  I don't really like that option so we may revisit in the future.

PR #813, opened 2015-02-01
Save viewstate for all levels of groups & labels

### LINKED ISSUE #806: CheckIn Configuration won't save any new Groups & Areas
Using the latest version of develop, I can't add a new GroupType and new Group(s) at the same time. I can only add one Group and then I have to save.

See https://www.dropbox.com/s/v7hhj6w9kau4avt/CheckinConfig.mov?dl=0

Related to #729 but now it's broken worse than it was before.

dcs619:
Fixes #806.  Currently only groups for the top-level areas (group types) are being saved in ViewState.  This adds all N-child-level groups to ViewState.

I couldn't find a way to get N-levels deep using standard LINQ, so I also added a non-recursive IEnumerable<Flatten> extension method.  See comments below.


mikepetersonccv:
Looks good! Thanks!

PR #826, opened 2015-02-04
Fix new REST key not confirmed on create 

### LINKED ISSUE #825: Unexpected behavior when adding a new API user to security group
Using the latest code in develop, I created a new API user but Rock won't let me immediately add it to a security role. See https://www.dropbox.com/s/ubtyg6vso776auj/NewAPIUser.mov?dl=0

Rock sees the new Person/User but won't let me see the User person or add the User to a group.  However, If I restart my build, I'm able to navigate to the PersonDetail and add the User.

I thought maybe a password had to be set, but that doesn't seem to be the case.

dcs619:
Also fixes UserDetail block showing incorrectly that it was. Closes #825.

### DIFF
--- a/RockWeb/Blocks/Security/RestKeyDetail.ascx.cs
+++ b/RockWeb/Blocks/Security/RestKeyDetail.ascx.cs
@@ -51,7 +51,7 @@ namespace RockWeb.Blocks.Security
             }
         }
 
-        #endregion Control Methods
+        #endregion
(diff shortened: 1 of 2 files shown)


dcs619:
@mikepetersonccv @azturner Can one of you merge this in?

azturner:
@dcs619 , it's been merged

PR #869, opened 2015-03-02
Provide helper classes for plugin developers to use theme colors

dcs619:
@edmistj Current theme colors are included on control styles, but aren't accessible by themselves.  This adds a couple classes for plugin developers to use themes without using the exact control style. 

It looks like _helpers.less was designed to do something similar, but that's not included anywhere & it didn't compile when I added it.

Side note: the latest Web Essentials is stable and auto-recompiles on save.

### DIFF
--- a/RockWeb/Styles/_checkin-core.less
+++ b/RockWeb/Styles/_checkin-core.less
@@ -78,28 +78,6 @@ body {
   }
 }
 
-// theme colors
-.primary-color {
-    color: @bg-color;
-}
-
-.secondary-color {
-    color: @text-color;
-}
-
-.background-primary {
-    background-color: @bg-color;
-}
-
-.background-secondary {
-    background-color: @text-color;
-}
-
-.base-font {
-    font-family: @font-family-base;
-}
-
-
 // 3. Components
 // -------------------------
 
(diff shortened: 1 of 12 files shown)


edmistj:
I’d like to get the _helpers.less working for check-in (there’s a few other classes that would be nice to have). I’m wondering if you can use the @accent-xxxx-colors instead of the .primary colors. That would make it more consistent with the core themes. I pushed a change that adds this. Let me know if you have questions.

dcs619:
@edmistj accent-xxxx colors are only set to the button color; we need the background and text colors as well.  

Can I add classes of bg-color and text-color so they're consistent?

edmistj:
Is this what you were thinking?

.bg-color {
    background-color: @body-bg;
}

.text-color {
    color: @text-color;
}

dcs619:
@edmistj Yep!  However, Rock themes use @body-bg for background color, and Check-in themes use @bg-color.  Can we make that variable name consistent?  I'll update the PR.

dcs619:
Whoops, just saw you already made those consistent.  Thanks!

edmistj:
OK, let me know if there is anything else you need.
-Jon



[jon-sig]

Issue #874, opened 2015-03-03
Issue with saving Binary Files


edmistj:
From this Q&A: http://www.rockrms.com/page/278?question=448
More details are on this post.

This may be a bug or it may be specific to my setup. Is anyone else having trouble uploading pdf's of previous background checks? To clarify the pdf will actually upload without a problem but it refuses to let me save the block once its finished. It will let me update the information under the background check block and save but never when there's a pdf involved. Is there a size limit that i'm missing?

saconger:
I think we have this fixed.  It's a deadlock issue.  I'll do a pull request.

PR #919, opened 2015-03-27
Deadlock issue with saving Binary File

saconger:
There was a deadlock issue with the background check document upload on the person extended attributes.

This fixes this bug:
https://github.com/SparkDevNetwork/Rock/issues/874


azturner:
@saconger , First, thanks for debugging this issue and finding a fix. However, to fix this, we're just going to change the AttributeValues.ascx.cs block so that it does not wrap the saving of all attributes and updating of history in one db transaction. This allows us to not have to update the base field type class and all the other field types that use a rockcontext when formatting the values. We made the fix for v3.1 (https://github.com/SparkDevNetwork/Rock/commit/54e494e53f9a23607d37142c5408423b5227b175)

saconger:
@azturner , Sounds good.  Thanks for the update.

PR #922, opened 2015-03-30
Add a binary file picker attribute

benwiley:
This adds an attribute that can be used to show a binary file field on a workflow action.  I made the binaryFileTypeGuid required since, without it being set, the field has no options and doesn't make sense.


benwiley:
image

dcs619:
To clarify, this screenshot is using the CheckinLabel.Guid to display instances of CheckinLabel files.  

For an email workflow, you would probably send out a Background Check document for someone to fill out and send back.

PR #959, opened 2015-04-19
Return the created entity after a REST POST

benwiley:
This will improve the efficiency of 3rd party software that uses the REST endpoint.  This is a very common requirement to need the Id (among other data points) when creating an entity through a REST client.

Consider this real world example of NewSpring's current usage:
1.) Create person locally
2.) Be sure to assign guid for reference
3.) 1st API request POST to Rock to create person there
4.) 2nd API request GET person back from Rock by Guid
5.) Update local person with Rock's Id and other default data (guid, primaryAliasId, etc) assigned upon creation

And with this pull request:
1.) Create person locally
2.) 1st API request POST to Rock to create person there
3.) Update local person with Rock's Id and other default data assigned upon creation

http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5 - I believe the POST specification in the link allows for returning the entity in this verbiage: "contain an entity which describes the status of the request and refers to the new resource"

The primary change is to lines generating the responses: 
return ControllerContext.Request.CreateResponse( HttpStatusCode.Created, person );

### DIFF
--- a/Rock.Rest/ApiController.cs
+++ b/Rock.Rest/ApiController.cs
@@ -128,7 +128,7 @@ namespace Rock.Rest
             System.Web.HttpContext.Current.Items.Add( "CurrentPerson", GetPerson() );
             Service.Context.SaveChanges();
 
-            var response = ControllerContext.Request.CreateResponse( HttpStatusCode.Created );
+            var response = ControllerContext.Request.CreateResponse( HttpStatusCode.Created, value );
 
             // TODO set response.Headers.Location as per REST POST convention
             //response.Headers.Location = new Uri( Request.RequestUri, "/api/pages/" + page.Id.ToString() );
--- a/Rock.Rest/ApiController.cs
+++ b/Rock.Rest/ApiController.cs
@@ -128,7 +128,7 @@ namespace Rock.Rest
             System.Web.HttpContext.Current.Items.Add( "CurrentPerson", GetPerson() );
             Service.Context.SaveChanges();
 
-            var response = ControllerContext.Request.CreateResponse( HttpStatusCode.Created, value );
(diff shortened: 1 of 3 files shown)


MrUpsideDown:
I may be misunderstanding the purpose of this change, but I'm not sure that this should be the default behavior when POSTing a new entity?
Isn't this potentially creating a lot of extra network traffic where it may not be necessary, by returning mostly the same data that was sent in the initial POST? For very large entities (such as the Person record) this could be quite inefficient where the new data is not needed.

If there is a need to return the new entity in its entirety in the circumstances described, my suggestion would be to implement this using a different operation (ie. not the default POST).

The POST specification says that the entity returned from the post "describes the status and ... *refers* to the new resource" - not that it *is* the new resource.
I think an example of this would be a response object that contains a status code (success/failure), a status message and the Guid *and* Id of the new resource. (The Guid can be used by the client as a local reference to the supplied data, and the Id is the server internal reference used in a subsequent GET).

Feel free to ignore - this is just my 0.02 worth!

benwiley:
After feedback from @dcs619 and @MrUpsideDown (thank you both!), I changed it to return only the Id of the new entity.  Guid can be assigned by the third party before sending to Rock for creation.  Other data is not as commonly needed.  So for the sake of speed in most common usage scenarios, I'll commit that change.

benwiley:
@azturner Not sure what to do with this controller.. https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Rest/Controllers/WorkflowsController.Partial.cs#L82-90

Do you want it updated to return an HttpResponse, while overriding the base.Post method?  It just seems inconsistent.

PR #989, opened 2015-04-26
Data View filter condition on Attribute does not return all matching Entities

MrUpsideDown:


mikepetersonccv:
Looks good! Much better :)

mikepetersonccv:
FYI, had to make a small tweak see https://github.com/SparkDevNetwork/Rock/commit/98341c5d959a9c444cbeec9122db0bef3a88e8df
:)

MrUpsideDown:
Nice catch! I have a feeling there will be a few other areas that will need to be tweaked for this inversion logic in the future. For example, what happens where an Attribute has a Default Value that satisfies the filter condition (e.g. Filter = Value < 10, Default Value = 9). Is the Default Value actually stored for each record, or just imputed? I have no spare time for testing this scenario at the moment, but I'm sure it will come up at some point in the future so I'll make a note of it here for future reference.

PR #1009, opened 2015-04-30
Changes to allow more compact display of Attribute values

MrUpsideDown:


edmistj:
Would it be possible to get before/after screenshots?  If it's a lot of work don't worry about it we can look at it after pulling the merge.

MrUpsideDown:
Screenshots as requested...
Before (empty field labels visible):
rock-group-viewer-before
After (empty field labels hidden):
rock-group-viewer-after

edmistj:
Ah, I see. Very nice!

nairdo:
Yes, that's great and it fits the spirit of things (not showing empty stuff).

MrUpsideDown:
I also have a suggestion about a possible future improvement in this area. Perhaps instead of the "Show In Grid" flag, we could have a more general "Detail Level" property associated with Custom Attributes? The setting could be something like "Low/Medium/High", and it could be used in a more general way to determine which items should be displayed in different blocks.
I think this would also fit nicely with a responsive UI which can scale the level of detail shown in blocks/grids to suit the client device. For example, on a mobile device only "High" detail items are displayed, on a large screen you might want to display "Medium" items by default and then show a "More..." button to display "Low" detail items if desired.
Needs some work-shopping, but what do you guys think of this idea in principle?

edmistj:
I like the idea. If we could, lets hold on to it and revisit it later in the year. We're focusing all our energy right now on getting the core features complete. Perhaps add it to the black book though so we don't lose it.

MrUpsideDown:
Done.

PR #1014, opened 2015-05-02
Google Authentication

arranf:
This is my first pull request so I'm not sure if I've missed anything. I've been testing using Rockit rather than a local version of the develop branch if that makes a difference.

This is a Google OAuth2 authentication which heavily borrows from the existing Facebook authentication service. It currently requests first name, last name, email, id, and gender. It requires a client secret and client ID from the [Google Developers Console](http://console.developers.google.com), the process of how to obtain those is detailed [here](https://developers.google.com/identity/protocols/OpenIDConnect#getcredentials).

I made the decision not to request Google+ circles (given how little Google+ is used and the fact you'd need to request an additional scope for that) and a user picture (given the likelihood it's the default).

I haven't used CodeMaid or any other tools to help with code formatting as I'm using VSE.

Feedback more than welcome.

### DIFF
--- a/Rock.Rest/Controllers/AuthController.cs
+++ b/Rock.Rest/Controllers/AuthController.cs
@@ -86,7 +86,7 @@ namespace Rock.Rest.Controllers
         /// <exception cref="System.Web.Http.HttpResponseException"></exception>
         [HttpPost]
         [System.Web.Http.Route("api/Auth/GoogleLogin")]
-        public void FacebookLogin( [FromBody]Rock.Security.ExternalAuthentication.Google.GoogleUser googleUser )
+        public void GoogleLogin( [FromBody]Rock.Security.ExternalAuthentication.Google.GoogleUser googleUser )
         {
             string userName = Rock.Security.ExternalAuthentication.Google.GetGoogleUser(googleUser);
             if ( !string.IsNullOrWhiteSpace(userName) )
--- a/Rock.Rest/Controllers/AuthController.cs
+++ b/Rock.Rest/Controllers/AuthController.cs
@@ -79,6 +79,26 @@ namespace Rock.Rest.Controllers
             }
         }
 
+        /// <summary>
+        /// Use this to Login a user and return an AuthCookie which can be used in subsequent REST calls
+        /// </summary>
+        /// <param name="loginParameters">The login parameters.</param>
+        /// <exception cref="System.Web.Http.HttpResponseException"></exception>
+        [HttpPost]
+        [System.Web.Http.Route("api/Auth/GoogleLogin")]
+        public void FacebookLogin( [FromBody]Rock.Security.ExternalAuthentication.Google.GoogleUser googleUser )
+        {
+            string userName = Rock.Security.ExternalAuthentication.Google.GetGoogleUser(googleUser);
+            if ( !string.IsNullOrWhiteSpace(userName) )
+            {
+                Rock.Security.Authorization.SetAuthCookie(userName, false, false);
+            }
+            else
+            {
+                throw new HttpResponseException(HttpStatusCode.Unauthorized);
+            }
+        }
+
     }
 
 }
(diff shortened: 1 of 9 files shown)


Rock/Security/Authentication/Google.cs
```diff
@@ -0,0 +1,363 @@
...
+        /// </summary>
+        /// <param name="user">The user.</param>
+        /// <param name="oldPassword">The old password.</param>
+        /// <param name="newPassword">The new password.</param>
+        /// <param name="warningMessage">The warning message.</param>
+        /// <returns></returns>
+        /// <exception cref="System.NotImplementedException"></exception>
+        public override bool ChangePassword( UserLogin user, string oldPassword, string newPassword, out string warningMessage )
+        {
+            warningMessage = "not supported";
```
dcs619: This looks a little sketch, not to mention the user experience if they actually tried it.  I believe there is a Password Change Not Supported clause in the Active Directory provider that you'd want to model it after.

Rock/Security/Authentication/Google.cs
```diff
@@ -0,0 +1,363 @@
...
+
+                // If no user was found, see if we can find a match in the person table
+                if ( user == null )
+                    {
+                    // Get name/email from Google login
+                    string lastName = googleUser.family_name.ToString();
+                    string firstName = googleUser.given_name.ToString();
+                    string email = string.Empty;
+                    try { email = googleUser.email.ToString(); }
+                    catch { }
```
dcs619: What's the purpose of this try/catch?  If it's just that the user's email might be null, there is a ToStringSafe method.

Rock/Security/Authentication/Google.cs
```diff
@@ -0,0 +1,363 @@
...
+                            if ( person != null )
+                            {
+                                PersonService.SaveNewPerson(person, rockContext, null, false);
+                            }
+                        }
+
+                        if ( person != null )
+                        {
+                            int typeId = EntityTypeCache.Read(typeof(Google)).Id;
+                            user = UserLoginService.Create(rockContext, person, AuthenticationServiceType.External, typeId, userName, "goog", true);
```
dcs619: I would make the "goog" parameter a static string as part of your class.  Is the wording "goog" a requirement or did you just pick that?

arranf: "goog" is just arbitrary. It's the password parameter for the UserLogin create method. It is unclear.

Rock/Security/Authentication/Google.cs
arranf: Thanks for looking over this David.
This portion was borrowed from the existing Facebook authentication. If that is the reasoning behind it I agree with you it's not needed. Thinking about it though, I'm not sure there'd be a situation where the user's email would be null. I don't think the user can choose scopes to accept (i.e. accept the profile scope but not the email scope) nor do I think it's possible to have a Google account without having a Gmail address or providing an existing email address.

Rock.Rest/Controllers/AuthController.cs
```diff
@@ -79,6 +79,26 @@ public void FacebookLogin( [FromBody]Rock.Security.ExternalAuthentication.Facebo
...
 
+        /// <summary>
+        /// Use this to Login a user and return an AuthCookie which can be used in subsequent REST calls
+        /// </summary>
+        /// <param name="loginParameters">The login parameters.</param>
+        /// <exception cref="System.Web.Http.HttpResponseException"></exception>
+        [HttpPost]
+        [System.Web.Http.Route("api/Auth/GoogleLogin")]
+        public void FacebookLogin( [FromBody]Rock.Security.ExternalAuthentication.Google.GoogleUser googleUser )
+        {
```
azturner: Might want to rename this method :)

arranf: Done

Rock/Security/Authentication/Google.cs
dcs619: Yep, I looked at the Facebook auth and you were just following the pattern.  Sorry about that :)  

@azturner Should the Facebook auth be updated to use ToStringSafe instead of the generic try/catch?

PR #1041, opened 2015-05-18
Fix "campu" typo in Check-in Location model

dcs619:
screenshot 2015-05-18 12 23 53

### DIFF
--- a/Rock/CheckIn/CheckInLocation.cs
+++ b/Rock/CheckIn/CheckInLocation.cs
@@ -38,13 +38,13 @@ namespace Rock.CheckIn
         public Location Location { get; set; }
 
         /// <summary>
-        /// Gets or sets the campu identifier.
+        /// Gets or sets the campus identifier.
         /// </summary>
         /// <value>
-        /// The campu identifier.
+        /// The campus identifier.
         /// </value>
         [DataMember]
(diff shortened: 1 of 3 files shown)


azturner:
@dcs619 Thanks for catching this. However, so that it's not a 'breaking change', can you add back the old CampuId property ( so it has both the wrong name and your new name ), but add the [obsolete] attribute.  For example...

    [DataMember]
    [Obsolete("Use CampusId property instead")]
    public int? CampuId
    {
        get { return CampusId; }
        set { CampusId = value; }
    }

-David

dcs619:
@azturner Did you mean to close this?  

I can add that property, but since CheckInLocation is a temporary model mapped to Rock.Location it wasn't referenced anywhere else...our check-in wasn't using it anywhere either.

azturner:
I was hoping you'd make that change and submit a new pull request. We just always try to avoid anything that could possibly be a breaking change. While there's no code in core (or your check-in) that references that field, it is possible ( although at this point unlikely ) that a third party app/plugin referenced that field, or even someone's lava on a check-in label referenced that property.

dcs619:
Gotcha.  

I made that change, but since this PR is closed it's no longer tracking our branch.  I'll open another.

PR #1043, opened 2015-05-18
Fix "campu" typo in Check-in Location model

dcs619:


nairdo:
This should show everyone how serious we are about backwards compatibility. I'm not saying we always get it perfect... I'm just sayin`. :)

PR #1051, opened 2015-05-19
Fix typo that was causing a "multiple controls with ID xxxxxx" message

dcs619:
also some whitespace cleanup @mikepetersonccv.

ylscqnqjwueazdaik-gipnqv4yybvkgbe4wugvsjsks


mikepetersonccv:
Looks good. BTW, I would suggest doing whitespace/stylecop cleanup as the only thing in a commit, then follow it up with the code changes in a 2nd commit. :)

dcs619:
@mikepetersonccv will do.  I have a couple more updates coming your way :)

PR #1052, opened 2015-05-20
Add group ancestry block attribute

dcs619:
This allows the attendance reporting to ignore however deep a group might be. 

In our case groups are at least 3 levels deep and are getting pushed out of the column they should live in:
screenshot 2015-05-19 16 57 22
.


azturner:
@dcs619 , I merged this request, and then made a [small change](https://github.com/SparkDevNetwork/Rock/commit/b88b1c6d2cb429f8a1ea879f1d0338f44935b768) so that GetAttributeValue() didn't need to be called on each iteration of the AddGroupControls() method.

dcs619:
@azturner Awesome, thanks!  I started out with it that way until I realized AddGroupTypeControls was also a recursive method.  The other option was to make it a global variable and pass it to both methods, but it looked too messy in the pull request so I ditched it.

PR #1054, opened 2015-05-20
Fix inherited grouptypes and groups in Attendance Reporting

dcs619:
@mikepetersonccv I stole all of @azturner's brilliant work on CheckInConfiguration to update this block.

This change displays groups at their current grouptype/area instead of at the top-level grouptype/area (see previous PR's screenshot for incorrect behavior).  It also allows groups to have a parent group and still be displayed.

screenshot 2015-05-20 10 35 18

One remaining issue whenever switching to the attendee view; there are too many columns for the size of the page and I would occasionally get a timeout.   

screenshot 2015-05-20 11 09 43

The timeout occurred when trying to load a single Sunday morning for a single campus (208 check-ins).  I'm guessing the timeout is due to the location query -- maybe we should add a Show/Hide Location block attribute?

### DIFF
--- a/RockWeb/Blocks/CheckIn/AttendanceReporting.ascx.cs
+++ b/RockWeb/Blocks/CheckIn/AttendanceReporting.ascx.cs
@@ -51,6 +51,9 @@ namespace RockWeb.Blocks.CheckIn
         private List<DateTime> _possibleAttendances = null;
         private Dictionary<int, string> _scheduleNameLookup = null;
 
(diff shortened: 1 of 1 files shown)


azturner:
@dcs619 , Is this still functionality you need?  If so, can you do another pull request that we can merge ( We're no longer able to automatically merge this request ).

dcs619:
@azturner No, since we refactored our group structure the requirements for this are different.  There was an issue with groups showing twice but I confirmed this morning with @frankgrand that it's fixed.

PR #1072, opened 2015-06-03
+ Selecting certain FieldTypes in AttributeEditor no longer causes ex…

### LINKED ISSUE #894: Selecting GroupTypeGroup Field-Type for Workflow attribute results in Exception
When constructing Workflows, I have run into a problem that is reproduce-able every time.
1. Add an attribute to workflow
2. Select 'Group Type Group' as field type
3. Error message is shown in on-screen notification about an unknown error

When checking the exception log I see this:

System.Web.HttpUnhandledException	System.Web	Exception of type 'System.Web.HttpUnhandledException' was thrown.	 Show Stack Trace
System.Collections.Generic.KeyNotFoundException	System.Web	The given key was not present in the dictionary.	 Show Stack Trace

Newest stable release of Rock

### LINKED ISSUE #1058: Group and Role field type causes exception
If you try to add an attribute with a field type of "Group and Role" an exception occurs...

image

tcavaletto:
…ceptions (Fixes #894, Fixes #1058).

### DIFF
--- a/Rock/Web/UI/Controls/AttributeEditor.cs
+++ b/Rock/Web/UI/Controls/AttributeEditor.cs
@@ -769,7 +769,7 @@ namespace Rock.Web.UI.Controls
         /// <param name="e">The <see cref="EventArgs"/> instance containing the event data.</param>
         void _ddlFieldType_SelectedIndexChanged( object sender, EventArgs e )
         {
-            Qualifiers = new Dictionary<string, ConfigurationValue>();
+            // When someone changes the field type, we clear the default value since it's no longer relevant.
             DefaultValue = string.Empty;
         }
 


Rock/Web/UI/Controls/AttributeEditor.cs
```diff
@@ -769,7 +769,7 @@ protected void cvKey_ServerValidate( object source, ServerValidateEventArgs args
         /// <param name="e">The <see cref="EventArgs"/> instance containing the event data.</param>
         void _ddlFieldType_SelectedIndexChanged( object sender, EventArgs e )
         {
-            Qualifiers = new Dictionary<string, ConfigurationValue>();
+            // When someone changes the field type, we clear the default value since it's no longer relevant.
```
tcavaletto: @azturner could you verify that this won't cause any problems?

Issue #1011, opened 2015-05-01
SMS: Not Automatically Enabled When A New Person is Added


arranf:
The check box for 'sms' is not automatically checked for new people who are added via the Add Person block.

If this is by design it'd be useful to have a block setting to allow this to be checked when people are added.

arranf:
Using Rockit I've implemented a per block setting if that's a route you guys would want to go down. I was also thinking it might be worth having a checkbox UI element that could be toggled off.

arranf:
Closed with #1078

PR #1078, opened 2015-06-06
Add a SMS checkbox to add family contact info pane

arranf:
My church was finding that adding a large volume of data via the add family screen was forcing users to then edit a new persons information after they were added.

This pull adds a Rock checkbox to the add family contact info pane which toggles SMS enabled on or off for the cell number input.

I'm not sure if is a route you guys want to go down (I can see the potential for creep). This would fix issue #1011 which I raised.


azturner:
@arranf , are you up to making the default value for this new checkbox a block setting on the the AddFamily block and defaulting it to false so that current behavior is the same?

arranf:
Yep that should be doable!

arranf:
@azturner Done

PR #1079, opened 2015-06-06
Use Defined Values for Family Pane Labels

### LINKED ISSUE #941: New Family Screen shows "Cell Phone" instead of the Defined Type Value
In the New Family Screen (Page 2) when entering contact information for family members, the "Home Phone" and "Cell Phone" field labels do not show the correct values from the Defined Type "Phone Type". The labels seem to be hard-coded to "Home Phone" and "Cell Phone".

arranf:
Fixes #941

### DIFF
--- a/Rock/Web/UI/Controls/NewFamily/ContactInfo.cs
+++ b/Rock/Web/UI/Controls/NewFamily/ContactInfo.cs
@@ -68,6 +68,9 @@ namespace Rock.Web.UI.Controls
         {
             if ( this.Visible )
             {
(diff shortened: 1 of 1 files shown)


azturner:
I merged this and then [made a few small changes](https://github.com/SparkDevNetwork/Rock/commit/caf484805ba4a7e8d9277e722dff5bbc78807050) to check for null, update the textbox placeholder text, and add ' Phone' to end of label if the defined-value didn't include it...

image

Let me know if you think it's an issue to have ' Phone' appended to the end of value.

arranf:
That's a much better commit!
Appending phone isn't a problem at all.

PR #1102, opened 2015-06-20
Auto Fill Mode

treyhendon:
Option to have the form not auto populate data just because there is a
logged in user.  Allows the form to go in the Group Toolbox as a method
for leaders to add new members.


treyhendon:
I think I got my new machine all caught up and ready.  Sorry to not do this sooner.

PR #1112, opened 2015-06-25
Add Redirect Page Setting

treyhendon:
Allows the Login Block to have a default redirect page if there is no URL param for 'returnurl'.  Example use: redirect a user on the "My Account" page if no 'returnurl' is in query string.


azturner:
@treyhendon , Thanks for the pull requests. When you make your commits, can you use [these guidelines](https://github.com/SparkDevNetwork/Rock/wiki/Committing-Your-Code). Thanks.
-David

PR #1114, opened 2015-06-25
+ Sites can now be deleted via an extra confirmation (Fixes #937).

### LINKED ISSUE #937: Unable to delete Sites
From @BelAirIT (Jeremy Turgeon):
Sites cannot be deleted.  They display a “This Site has a Layout which is used by a page.”

tcavaletto:


azturner:
@tcavaletto , Still get an exception trying to delete a site if there are any pageview records associated with a page that is being deleted

nairdo:
Should probably cascade delete pageviews on delete of page in model, right David?

PR #1116, opened 2015-06-26
Add Register Button Alt Text

treyhendon:
Allows the Register button to use alternate text like "Sign Up" or "Contact Leader".


RockWeb/Blocks/Groups/GroupRegistration.ascx.cs
```diff
@@ -574,6 +575,11 @@ private bool CheckSettings()
 
             tbEmail.Required = _autoFill;
 
+            string registerButtonText = GetAttributeValue( "RegisterButtonAltText" );
+            if ( string.IsNullOrWhiteSpace( registerButtonText ) )
+                registerButtonText = "Register";
```
azturner: @treyhendon, just a [coding-standard](https://github.com/SparkDevNetwork/Rock/wiki/Coding-standards) note, we always use braces on our if/else statements even if it's just one-line :). I'll update this one after the merge. Thanks again for jumping in and doing the pull requests.
-David

PR #1123, opened 2015-06-29
Feature billing info

benwiley:
@edmistj I've made the following changes:

1.) Billing Address Location Id 
2.) Expiry Month * 
3.) Expiry Year * 
4.) Name on Card *
  * should use a field that is 2-way encrypted in the DB

I am thinking that it might make sense to add a FinancialPersonSavedAccountId to the ScheduledTransaction model.  Thoughts?  Rather than duplicate fields onto the scheduled model, would it make sense to associate a schedule with a saved account?

Also I was not sure what methods to add to the Gateway class.  The gateways are already handling billing address, name on card, and expiry.  The original issue was that Rock was simply not storing this information after it was sent along to the gateway (payment processor).

I tested the migration by adding functionality to the demo give page that comes with Rock, but removed the code because I wasn't sure you actually wanted it.  The removal is part of the second commit, but I can easily add it back if wanted.

Thanks for your guidance on this!


azturner:
@benwiley ( and @dcs619 ) Looks like we had another migration added between the time you branched from develop and submitted this pull request so we can't automatically merge. 

Also, in looking at this a little closer to try and answer your questions, I think we should take a different approach to modifying the models. Instead of adding these fields to the PersonSavedAccount model and the scheduledTransaction ( or adding foreign key to saved account). We should probably do the following instead...

Add a new FinancialPaymentDetail table with following fields:
Id, Guid, MaskedAccountNumber, CurrencyTypeValueId, CreditCardTypeValueId, NameOnCardEncrypted, ExpiryMonthEncrypted, ExpiryYearEncrypted, BillingLocationId.

Make following changes to the FinancialTransaction, FinancialPersonSavedAccount, and FinancialScheduledTransaction tables:
Add FinancialPaymentMethodId field ( foreign key to new table )
Remove CurrencyTypeValueId and CreditCardTypeValueId fields ( since they are now on new table )

The FinancialPaymentDetail table would then be more of a log table of the details that existed when a payment was made. When user wants to update their information, they could do so, but a new entry would be added to this table, and the foreign key on the SavedAccount and/or ScheduledTransaction table would be updated to point to new value, so that future FinancialTransactions generated from saved account or scheduled transaction would then use the new values.

The migration would also need to create records in the new table for all the existing FinancialTransaction, FinancialPersonSavedAccount, and FinancialScheduledTransactions and copy the existing CurrencyTypeValueId and CreditCardTypeValueId values from those tables before they are removed.

Because this is more complex (and includes a migration), thinking it would be easiest for us to make and coordinate this change along with the block/UI updates to allow user to make changes. Is this something you guys need before your golive date? or can it wait until shortly after golive?

-David

dcs619:
@azturner @benwiley that change seems a lot more complicated and would push back other timelines for launch (for you or for us).  Can you help us understand why splitting SavedAccount into two tables is a better solution?

Considering SavedAccount (with the new fields) is structurally the same as SavedAccount + PaymentDetail:

1) Could we add an IsActive flag to SavedAccount, with the expectation that only new SavedAccounts entries are created (and Transactions are updated to reference the new SavedAccount)?  This would become the audit log you needed.

2) Are you worried SavedAccount is too large/wide to handle any new fields?  With FK's, the 4 new fields & IsActive, SavedAccount would have 25 columns.

3) Does the name SavedAccount incorrectly reflect what it does?  If so, maybe we should rename FinancialPersonSavedAccount to FinancialPaymentDetail.

4) Could we make the changes to SavedAccount now and do a refactor post-launch/post 4.0?  We wouldn't need the Rock blocks updated since we're building a native app for giving.

azturner:
SavedAccount records only exist for a subset of scheduled transactions and is only used when the user elects to save the cc/ach info that they used when creating a one-time or scheduled transaction.  If they don't elect to do so, a SavedAccount record is not created (and we wouldn't want it to be created). PaymentDetails would exist for every transaction. Because of this, SavedAccount (with new fields) isn't quite the same as SavedAccount + Payment Details.

For this same reason, it also really doesn't make sense to add a foreign key to ScheduledTransaction to the SavedAccount table, since not every ScheduledTransaction would have corresponding SavedAccount record. 

If the idea is to let anyone with an existing scheduled transction be able to update their CC info, then you'd need to either add those same fields to the ScheduledTransaction table, or create a new table that can be used by both the SavedAccount and ScheduledTransaction. Adding the new table rather than duplicating those fields on both SavedAccount and ScheduledTransaction would be better. Especially when you consider that the FinancialTransaction table could then use it also (although that is just a bonus and not a requirement).

Does that make sense?

edmistj:
Wanted to jump in to let you know we have an extensive native mobile app we're working on that integrates with Rock. It might be helpful to your needs. 

Jered on our team has been writing it for almost a year now using Xamarin.
https://github.com/SparkDevNetwork/Rock.MobileApp

One other thing to note that we've run into with out current app is Apple's rules about donations inside a mobile app. Wanted to make sure you were aware of them:
https://developer.apple.com/app-store/review/guidelines/

See section 21.2.  We have talked to Apple to ensure we understood the limitations (which are pretty clear) and they confirmed them. You might end up having to use the blocks at least for the Apple users.

dcs619:
@azturner +@jbaxleyiii We can wait for the refactor to support payment edits and display the expiration date.

In order to match our current functionality by July 19th, we need a way to store a Billing Address for people.  Is creating a Person Attribute of type Location the best route to take?

dcs619:
@azturner Any chance you have an ETA on this?  

If not, Billing Address is the biggest question mark before we launch.  Is there a good way to assign a  Billing Location to people?

azturner:
@dcs619 , I'll get those model changes done this weekend.

azturner:
This has been merged into the development branch with this commit: https://github.com/SparkDevNetwork/Rock/commit/f50e703aa1fc8f052d73c46437181674cc3ea407. The models and blocks have been updated, but there has not been any functionality added to actually allow user to update their information on a pre-existing saved account or scheduled transaction.

PR #1130, opened 2015-07-01
Fix issue with FilterBySpecialNeeds checking for a null attributevalue

dcs619:
SpecialNeeds was checking for an attribute value prior to person.Attributes being loaded from the database.  

AbilityLevel was checking for an attribute value but always loading person.Attributes (maybe it was the first action in the stock workflow?).

Now attribute values are available regardless of the workflow action order.


dcs619:
Also, I thought there was an extension HasAttributes() on IEntity that would taken the place of ( Attributes == null ) but I didn't see it.

dcs619:
Changed commit message per dev guidelines

PR #1139, opened 2015-07-09
- Improved performance of Bulk Update Add/Remove Group to avoid timeo…

### LINKED ISSUE #1136: Bulk Update - Add To Group fails with timeout when processing large number of records
When trying to add a large number of people (1000+) to a Group, the update fails with a Timeout error.
The process seems to be taking far too long for a relatively simple update.

This also applies when trying to remove people from a Group

MrUpsideDown:
…ut failures with large data sets. (Fixes #1136)

### DIFF
--- a/RockWeb/Blocks/Crm/BulkUpdate.ascx.cs
+++ b/RockWeb/Blocks/Crm/BulkUpdate.ascx.cs
@@ -1044,17 +1044,27 @@ namespace RockWeb.Blocks.Crm
                     {
                         var groupMemberService = new GroupMemberService( rockContext );
 
-                        var existingMembers = groupMemberService.Queryable( "Group" )
-                            .Where( m =>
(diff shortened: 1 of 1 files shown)


MrUpsideDown:
I had to make a minor tweak to this, hence the second commit - can these commits be squashed together into the develop branch, or do I need to submit a new pull request?

arranf:
David, I had a similar situation to you last week. I've written up
instructions on how to deal with it and they're linked to from the Pull
Request wiki page.

You need to squash them on your branch and the change will be reflected in
the pull.

MrUpsideDown:
Arran, thanks for the advice. I've squashed on the local branch before, but I ran into trouble this time because I had already pushed the commits to my remote github fork. Something went wrong when I tried to revert  the fork commits, then squash, then resubmit to create a single commit. So I gave up and resubmitted instead!

PR #1150, opened 2015-07-13
Added public description html and is public fields to accounts

benwiley:
Added fields to FinancialAccount:

1. Public Description: Allows users to associate HTML content with an account
2. Is Public: A flag to indicate if the fund should be shown on an external give app


benwiley:
To explain the desire for these fields a little bit more: 

1.) We are thinking along the lines of CMS and externally facing sites that might use Rock.  We want our congregation to see a rich explanation (the HTML within PublicDescription) of an account that they might choose to give to.  

2.) We also have accounts that are active (IsActive = true) but used internally (IsPublic = false) - so we would not want them to show up on the external giving site.

azturner:
FYI, before merging I looked at the source repo, and your migration was listed after our latest migration 'RegistrationEmails' in the project file, so figured it was good for merge...

image

However, after merging, I realized you created your migration a while ago and it wasn't based on our latest migration. Because of this, I needed to delete your migration and recreate it from the latest migration we have. While this doesn't adversely affect us (besides having to redo the migration), it may affect any databases you have that you've already run that migration against.

PR #1153, opened 2015-07-14
+ Added default campus setting to Add Family block

arranf:
After staring for too long at the block settings for the Add Family block I thought it made sense to have a default campus for churches operating as multi-service rather than multi-campus.


arranf:
Thinking about this this morning. Instead of having a default campus setting (to avoid cluttering the block settings) perhaps it would be best to default to the only campus listed in the case of there only being one campus? That would fit the use case I was thinking of and would remove the temptation for churches with a main and plant campus(es) to set the default setting as the main campus which would like become a source of data error.

edmistj:
I like your thought of not using a block setting and using the only campus if there's is only one.

arranf:
In that case I'll submit a new pull.

PR #1157, opened 2015-07-22
- Improved performance of Bulk Update Add/Remove/Update Group to avoi…

### LINKED ISSUE #1136: Bulk Update - Add To Group fails with timeout when processing large number of records

MrUpsideDown:
…d timeout failures with large data sets. (Fixes #1136)


azturner:
@MrUpsideDown Very clever solution!

MrUpsideDown:
@azturner Thanks :-)

PR #1173, opened 2015-07-29
+ Added Report Field "Related People" to show a list of people with s…

MrUpsideDown:
…pecified family or known relationships to the subject person.

### DIFF
--- a/Rock/Reporting/DataSelect/Person/RelatedPeopleSelect.cs
+++ b/Rock/Reporting/DataSelect/Person/RelatedPeopleSelect.cs
@@ -1,0 +1,468 @@
+﻿// <copyright>
+// Copyright 2013 by the Spark Development Network
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+// http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System;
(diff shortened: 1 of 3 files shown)


azturner:
@MrUpsideDown , a free tool that we use for adding xml comments to public properties/methods is [GhostDoc](http://submain.com/products/ghostdoc.aspx). It is a VS plugin that allows you to right-click a property or method and select "Document This" and it will generate and add xml comments to the property or method.  If you compile in release mode, you'll get warnings for any properties and methods that are missing these comments.

MrUpsideDown:
Ok, thanks. I know the comments are sparse, but my habit is to not add xml comments for overrides unless the override does something unusual. Otherwise, I find the comment is usually just adding bloat by repeating the comment already supplied for the base method. Is the preference to have absolutely all public methods and properties commented? If so, I will be sure to do that.

azturner:
Yep, that's the preference, especially since the compiler by default will generate warnings for any that don't.

MrUpsideDown:
Ok, noted - I'll mend my ways for future commits!

Rock/Reporting/DataSelect/Person/RelatedPeopleSelect.cs
```diff
@@ -0,0 +1,468 @@
...
+                                                                                                                        } ) );
+
+            return relatedPeopleQuery;
+        }
+
+        private const string _CtlFormat = "ddlFormat";
+        private const string _CtlKnownRelationshipType = "cblKnownRelationshipType";
+        private const string _CtlFamilyRelationshipType = "cblFamilyRelationshipType";
+
+        public const string FamilyRelationshipParentGuid = "A498E9DB-4BFF-4C52-A6C0-7510EECDF6E7";
```
dcs619: @MrUpsideDown Are these existing GUID's in the Rock database or new GUID's that you're creating for the report?

MrUpsideDown: @dcs619 These values are only used to persist the settings for this report.

PR #1179, opened 2015-08-03
+ Added option for Filter Group Condition = "True/False" to implement…

MrUpsideDown:
… logical "NOT" operations for Filter Groups. This allows existing Filter Conditions to be negated.

### DIFF
--- a/Rock/Model/DataViewFilter.cs
+++ b/Rock/Model/DataViewFilter.cs
@@ -248,6 +248,7 @@ namespace Rock.Model
                     return null;
 
                 case FilterExpressionType.GroupAll:
+                case FilterExpressionType.GroupAnyFalse:
 
                     Expression andExp = null;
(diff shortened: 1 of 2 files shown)


MrUpsideDown:
Another area where this is particularly useful is where we have a Data View that identifies a specific set of people, and we want to exclude these people from the results of a second Data View (possibly with combined with other conditions). Without the True/False modification, the only alternative is to re-create the inverse of the first Data View, which is tedious and introduces the potential for error if the logic of the first Data View is somewhat complex.

arranf:
This is ace. Thank you David!

azturner:
Sounds good. We'll pull and take a look.

MrUpsideDown:
@arranf You're welcome - I hope this change simplifies a few queries for you, just as it has for us. :-)

PR #1186, opened 2015-08-11
Tweak UI overhaul

sheck:
@edmistj here are the changes we discussed earlier

### DIFF
--- a/RockWeb/Styles/_grid.less
+++ b/RockWeb/Styles/_grid.less
@@ -1,6 +1,6 @@
 ﻿/*
     Used to manage styles related to Rock's internal grid component
-    
+
     1. Bootstrap Overrides
     2. Core Table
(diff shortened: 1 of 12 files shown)


edmistj:
I cherry picked most of the commits. I hesitated to move the panel-info and pill styles out of core since some of the core blocks use these. Everything else looked great though. Thanks for the fixes.

sheck:
Is there anyway we can scope those styles, then? Writing styles in custom themes to undo this and return to the bootstrap defaults feels a bit backwards.

dcs619:
+1 

We've been working on a themed external workflow page to replace Wufoo and noticed a significant number of styles needed just to get back to bootstrap defaults.  Or, the alternative, don't use core styles at all (increasing maintenance).

PR #1187, opened 2015-08-12
+ Added a Copy command for Data Views.

MrUpsideDown:
Adds a copy button to duplicate an existing Data View. If the approach taken in this PR is OK, I intend to do the same for Reports.

### DIFF
--- a/Rock/Model/CodeGenerated/DataViewService.cs
+++ b/Rock/Model/CodeGenerated/DataViewService.cs
@@ -21,6 +21,7 @@
 // </copyright>
 //
 using System;
+using System.Data.Entity;
 using System.Linq;
 
 using Rock.Data;
@@ -82,6 +83,57 @@ namespace Rock.Model
                 return false;
             }  
             return true;
+        }
+
+        /// <summary>
+        /// Create a new non-persisted Data View using an existing Data View as a template. 
+        /// </summary>
+        /// <param name="dataViewId">The identifier of a Data View to use as a template for the new Data View.</param>
+        /// <returns></returns>
+        public DataView GetNewFromTemplate( int dataViewId )
+        {
+            var item = this.Queryable()
+                           .AsNoTracking()
+                           .Include( x => x.DataViewFilter )
+                           .FirstOrDefault( x => x.Id == dataViewId );
+
+            if ( item == null )
+            {
+                throw new Exception( string.Format( "GetNewFromTemplate method failed. Template Data View ID \"{0}\" could not be found.", dataViewId ) );
+            }
+
+            // Deep-clone the Data View and reset the properties that connect it to the permanent store.
+            var newItem = (DataView)( item.Clone( true ) );
+
+            newItem.Id = 0;
+            newItem.Guid = Guid.NewGuid();
+            newItem.ForeignId = null;
+
+            newItem.DataViewFilterId = 0;
+
+            this.ResetPermanentStoreIdentifiers( newItem.DataViewFilter );
+
+            return newItem;
+        }
+
+        /// <summary>
+        /// Reset all of the identifiers on a DataViewFilter that uniquely identify it in the permanent store.
+        /// </summary>
+        /// <param name="filter">The data view filter.</param>
+        private void ResetPermanentStoreIdentifiers( DataViewFilter filter )
+        {
+            if ( filter == null )
+                return;
+
+            filter.Id = 0;
+            filter.Guid = Guid.NewGuid();
+            filter.ForeignId = null;
+
+            // Recursively reset any contained filters.
+            foreach ( var childFilter in filter.ChildFilters )
+            {
+                this.ResetPermanentStoreIdentifiers( childFilter );
+            }
         }
     }
 
--- a/RockWeb/Blocks/Reporting/DataViewDetail.ascx.cs
+++ b/RockWeb/Blocks/Reporting/DataViewDetail.ascx.cs
@@ -142,6 +142,47 @@ $(document).ready(function() {
         }
 
         /// <summary>
+        /// Handles the Click event of the Copy button control.
+        /// </summary>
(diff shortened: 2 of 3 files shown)


arranf:
You beat me to the punch on this one!

MrUpsideDown:
Sorry about that... I should have logged this in GitHub with a note to say I am working on it, but I can find no way to label issues as "Enhancement". We've been using this mod for a while in production, but I finally found the time to replicate it for Reports which prompted me to put together PR's for both.

arranf:
No worries ;) I don't tend to flag an issue for something I'm working on
either. Perhaps we should as good practice?

MrUpsideDown:
I think so, yes. I have created issues for bug fixes in the past, but the enhancements are tricky - I can't label them as such in GitHub (as far as I know), but they're not really "issues" either. If only the "black book" were more transparent, I'd put them in there... ;-)

arranf:
I'll mention @azturner and @edmistj because they might have an idea for how to tackle this (perhaps there's a way to make the enhancement and in progres' label available for us?) but in the meantime If we start creating issues and prefixing them with [EN] then that'll make it fairly plain that it's something we're working on?

mikepetersonccv:
@MrUpsideDown Wow, looks great.  Very nice solution!

Rock/Model/CodeGenerated/DataViewService.cs
```diff
@@ -83,6 +84,55 @@ public bool CanDelete( DataView item, out string errorMessage )
...
+        /// <param name="dataViewId">The identifier of a Data View to use as a template for the new Data View.</param>
+        /// <returns></returns>
+        public DataView GetNewFromTemplate(int dataViewId)
+        {
+            var item = this.Queryable()
+                           .AsNoTracking()
+                           .Include( x => x.DataViewFilter )
+                           .FirstOrDefault( x => x.Id == dataViewId );
+
+            if (item == null)
```
dcs619: @MrUpsideDown Small thing, but you probably want to include brackets around single-line if statements.  Noted in the Rock documentation here: https://github.com/SparkDevNetwork/Rock/wiki/Coding-standards

RockWeb/Blocks/Reporting/DataViewDetail.ascx.cs
```diff
@@ -142,6 +142,47 @@ protected void btnEdit_Click( object sender, EventArgs e )
...
+        /// Handles the Click event of the Copy button control.
+        /// </summary>
+        /// <param name="sender">The source of the event.</param>
+        /// <param name="e">The <see cref="EventArgs" /> instance containing the event data.</param>
+        protected void btnCopy_Click( object sender, EventArgs e )
+        {
+            // Create a new Data View using the current item as a template.
+            var id = int.Parse( hfDataViewId.Value );
+            
+            var service = new DataViewService( new RockContext() );
```
dcs619: @MrUpsideDown I'd suggest naming this variable what it is (dataViewService).  Specific variable names are easier to maintain and read than generic ones.

MrUpsideDown: @dcs619 I agree in principle - but in this case, the variable is only used once in the very next line? There's probably a better argument for just removing it altogether, but I prefer not to nest too many object creation statements in a row.

Rock/Model/CodeGenerated/DataViewService.cs
MrUpsideDown: Yeah, I missed that one - sorry. I know the standard, it just doesn't come naturally to me.

MrUpsideDown: @dcs619 Perhaps you can help me here. On the topic of minor amendments to PRs, what is the best method for submitting changes like this? In the past, I have tried simply pushing a new commit to the branch in my forked repo, but this doesn't squash the commits in the PR, just adds a second commit (which is not preferred as I understand it). Is there any way to do this easily, because otherwise it requires me to close and re-create the PR (which is time-consuming)?

MrUpsideDown:
@mikepetersonccv Thanks :-)

Rock/Model/CodeGenerated/DataViewService.cs
arranf: @mrupsidedown I wrote about that precise topic here as a reminder to myself
http://blog.arranfrance.com/arrans-guide-to-fixing-a-pull-request/

MrUpsideDown: @arranf Thanks, I'll give that a try!

MrUpsideDown:
@dcs19 Changes made as requested... thanks for the pointers.

azturner:
Opening an issue for an enhancement that your working on is fine. I would recommend mentioning @edmistj or I so that we have an opportunity to see it and give you some feedback (i.e. let you know if we've already started something similar or it potentially might not be something for core ). We can add the Enhancement label then too.

arranf:
Thanks David.

MrUpsideDown:
@azturner This sounds like a good approach, it will go some way to making sure we're not treading on each other's toes. I will also submit another PR to replicate the Copy feature for Reports.

MrUpsideDown:
@arranf Thanks for your help with squashing commits in a remote. I read your blog post and got a good understanding of the process so I could apply it to my use of SmartGit, and... success!

RockWeb/Blocks/Reporting/DataViewDetail.ascx.cs
dcs619: @MrUpsideDown Sorry, I've been out for a few days.  I agree with not nesting too many creates in a row.  I generally create the RockContext separately and instantiate services on the fly when needed.  A lot of the service methods take an optional RockContext so you can pass in the already-created context for a (tiny) performance benefit.

PR #1188, opened 2015-08-12
Suggested layout changes to bio block, added breakpoint namespaced alignment helper classes

briankalwat:
image

### DIFF
--- a/RockWeb/Blocks/Crm/PersonDetail/Bio.ascx
+++ b/RockWeb/Blocks/Crm/PersonDetail/Bio.ascx
@@ -16,32 +16,7 @@
         </div>
 
         <div class="row">
-            <div class="col-sm-6">
-                <h1 class="title name">
-                    <asp:Literal ID="lName" runat="server" /></h1>
-            </div>
-            <div class="col-sm-6 labels">
-
-                <Rock:PersonProfileBadgeList ID="blStatus" runat="server" />
-
-                <ul id="ulActions" runat="server" class="nav pull-right">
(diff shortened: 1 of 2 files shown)


dcs619:
@edmistj This fixes some weird alignment issues we were seeing between mobile/desktop breakpoints.  We made some other changes in our theme but figured you'd want the alignment fixed in core.

edmistj:
We'll pull it in and take a look at it.

edmistj:
Now that we have the pull in we've been able to look at it with production data. I like many of the concepts, but viewing it with more data there still are a few challenges. 

With the social icons the 'Action' drop-down gets lost. Also with a blank background it seems like having the contact info under the demographic section leaves a large empty area. When implemented with a background image like your site it works, but with a solid color it feels a little unused.

I also worry a bit about all the videos, documentation and collateral that we'd need to change. Long-term we never want these types of things to hold back innovation but right now we're struggling day/night/weekends just to keep our head above water.

Let me know your thoughts about these ideas. I've attached a before/after of the change in our environment.  Do know that we like to collaboration here, just sharing some of the other facets that go with the decisions.

bio-options

edmistj:
I made a few changes to the bio block to try to blend the two designs.  If you want to go back to a single column for the demographics and contact info you just need to override one style:

.demographics {
      @media @md { 
            float: inherit;
        }
}

The only other real change from yours is the actions being back in the upper right. The proposed change had them under the picture where it felt lost especially with social icons. We're open to other ideas for it's placement.

PR #1192, opened 2015-08-15
+ Added a Copy command for Reports.

MrUpsideDown:
Uses the same method as #1187 for copying Data Views.

### DIFF
--- a/Rock/Model/CodeGenerated/ReportService.cs
+++ b/Rock/Model/CodeGenerated/ReportService.cs
@@ -21,6 +21,7 @@
 // </copyright>
 //
 using System;
+using System.Data.Entity;
 using System.Linq;
 
 using Rock.Data;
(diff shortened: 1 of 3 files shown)


Rock/Model/CodeGenerated/ReportService.cs
```diff
@@ -53,6 +54,43 @@ public bool CanDelete( Report item, out string errorMessage )
             errorMessage = string.Empty;
             return true;
         }
+
+        /// <summary>
+        /// Create a new non-persisted Report using an existing Report as a template. 
+        /// </summary>
+        /// <param name="reportId">The identifier of a Report to use as a template for the new Report.</param>
+        /// <returns></returns>
+        public Report GetNewFromTemplate( int reportId )
```
azturner: @MrUpsideDown , I didn't catch this when I pulled this, but you want to avoid modifying any of the files in the codegenerated files. We periodically regenerate these files and any changes to them would get lost. They are all partial classes so anytime you need to add methods to them (i.e. service classes ), you can create a new partial class ( see this commit that fixes this request for an example: https://github.com/SparkDevNetwork/Rock/commit/27ae302b1a8841c626bcfd59d98ad4c3ce0debe1 )

MrUpsideDown: @azturner Sorry, I should have been alert to the fact that it was in a folder named "CodeGenerated"! Thanks for picking this up.

dcs619:
@azturner Will these changes be in 4.0?  I don't see a branch split from develop yet...

azturner:
Yes, they will be in 4.0. We plan to branch for 4.0 at the end of August. So this also means that if anyone has any pull requests that they'd like to be in v4.0, they should plan on having them submitted by the end of the month. After then, we'll just be doing testing and bug fixing in preparation for the 4.0 beta release.

@dcs619, @MrUpsideDown, @arranf, @nairdo, @treyhendon, @briankalwat, @benwiley,  @swedd, @saconger

arranf:
Thanks for the heads up.

dcs619:
@azturner Sounds good.  @MrUpsideDown I don't suppose you'd have time to add a CopyWorkflow action?  I can't get to it this week but would use your same pattern to get that added before the end of the month.

PR #1207, opened 2015-08-21
Fix non-default medium components not populating

### LINKED ISSUE #1172: InitializeFromSender not called for second Communication Medium in Simple Communication Block

arranf:


azturner:
@arranf , this actually caused the templates in the Communication Entry block to stop working (content was not loaded in HTML area when a template was selected ). I backed out this change and made another change that I think fixes what you were trying to accomplish. We don't have any non-default mediums though, so you may want to confirm.

arranf:
@azturner Sorry for that, I didn't test templates thoroughly enough.

Just to clarify for non-default I meant mediums that weren't the default medium when you load the block.

I've just tested your fix which works for the initial load however if I was to switch to a different medium and then switch back the fields are no longer populated.

azturner:
@arranf which fields are no longer populated? When login as Ted Decker for example, and select new communication, the From Name and From Address have Ted's name and email address. If I switch to the SMS medium and then back to Email, it still shows Ted's name and email. Are you not seeing that behavior?

arranf:
@azturner I've figured it out. It wasn't your fix at all. The issue is occurring because of Mediums's MediumData having overlapping keys. 

The default SMS and Email mediums don't shared medium data keys but our third medium is an alphanumeric SMS medium and so it shares keys with the default SMS medium.

arranf:
@azturner As an aside. Is LoadTemplates() needed on line 308? BindMediums on line 304 calls it.

azturner:
@arranf , nope, not needed... removed. Thx.

PR #1227, opened 2015-09-02
+ Added "Applies To" and "Data View" fields to read-only details of Report

MrUpsideDown:


MrUpsideDown:
This change eliminates the inconvenience of having to Edit a Report to see which Data View is used to supply its data.

azturner:
@MrUpsideDown , We'd rather keep the display of the report fairly clean with just the description of the report showing. What the report applies to should be obvious from the description of the report (and category grouping). For easily determining the data view [we just added](https://github.com/SparkDevNetwork/Rock/commit/21a0fd596a08f3730f99f4d680236b46233af364) a "Data View" link next to the security button.that is visible to those with edit rights to the report. It has a tooltip that shows the name of the data view when hovering over the link, and clicking it actually takes you to the data view. Hopefully that accomplishes what you were trying to here.

MrUpsideDown:
@azturner I actually chose this presentation method to preserve consistency with the Data View detail display! I must say, our users have found this mod really helpful when sifting quickly through reports or when they are frequently re-using the same report with different Data Views.
I like the idea of linking to the Data View, but I don't think the tooltip will be as convenient when moving quickly between reports because it requires a lot more mouse movement to reveal the information.
Perhaps a better option for reducing the size of the Report Detail header (and perhaps the Data View header also) might be using a "Show/Hide Details" toggle in the header, similar to the Results section?

MrUpsideDown:
@azturner As a further note to this, although the new "Data Views" and "Reports" sections are a very useful addition to the Data View Detail block, in our case it often adds 20+ lines of information to the header which pushes the results secrion off the page. Perhaps a tabbed interface would work better here? Description could be always visible, and then have a set of tabs for "Details" (displays Applies To/Category/Filter), "Usages" (displays Data Views/Reports) and "Results". That would deliver a compact header section, and allow the additional information to be lazy-loaded depending upon the active tab. Thoughts?

PR #1237, opened 2015-09-06
Enhancement - Enables Url Query Filters For My Workflow Block

edolyne:
Overview:
---
This code change allows the My Workflow block to be filtered via query string.

Example:
---
Using the following url [http://localhost:6229/?RoleFilter=false&StatusFilter=true](http://localhost:6229/?RoleFilter=false&StatusFilter=true) on page load (not on a post back) the My Workflow block shows only workflows that are assigned to the current user and active.

Use Case:
---
We are building a notification block which shows the current number of active workflows assigned to the current user.  Now with the help of the url query string when the user clicks on the notification icon the user is taken directly to a page with the My Workflow block already filtered to show only the workflows that are assigned to them and currently active.


dcs619:
@edolyne You had some duplicate code sections so I moved them out of the if/else.

PR #1246, opened 2015-09-10
googleMapsIsLoaded

sheck:
I'm running into a scenario where I have to dynamically load the google maps API on certain pages. Some blocks (GroupDetailLava.ascx for example) add the google maps API on page load and do not use the rock util for it. The problem I'm having occurs when I am using pages with these blocks. Calling the rock util to load the gmaps api neither loads the api nor fires the loaded event — leaving my scripts waiting for the event.

To fix this I could either 1. check for the gmaps object before trying load in my own scripts, or 2. we could have the util fire the loaded event if the api is already loaded. Here's an implementation of option 2.


mikepetersonccv:
Looks good!

Issue #1261, opened 2015-10-07
Pickers: html element Id collision


demersus:
We created a custom workflow action that uses a Workflow Type attribute.  Unfortunately, the way the ItemPicker class is coded, we cannot have multiple of these in the same workflow editor.  

I will post a merge request to here. 

 I discovered that the PersonPicker does not inherit from ItemPicker.  It uses this.ClientID for the element & javascript picker call.  However, the WorkflowTypePicker does inherit from ItemPicker.  

The ItemPIcker code makes the html element use the ID of the c# class instead of ClientID.  This was causing issues because the jquery ui widget would only initialize on one of the picker elements.  (all html elements had the same html id attribute).

Anyway,  you'll see what I mean from my merge request.

azturner:
Fixed with this commit: https://github.com/RealLifeMinistries/Rock/commit/ab79ca88ab56a5804b4326f077b8d8c01f0794d0

PR #1262, opened 2015-10-07
Change ItemPicker:  ClientID instead of ID

demersus:
See issue #1261


mikepetersonccv:
@demersus Yep, this is definitely a bug. Can we fix it a slightly different way? Basically, instead of using ClientId for the control prefixes, etc, how about just changing RegisterJavaScript() and RenderControls() use the .ClientID instead of the .Id

demersus:
That sounds good. I will update the merge request tomorrow.

mikepetersonccv:
@demersus Looks good! Thanks for finding that issue and fixing it.

demersus:
@mikepetersonccv  You're welcome.  Thanks for building this system!

PR #1265, opened 2015-10-11
Add Group Member report fields

MrUpsideDown:


Rock/Reporting/DataSelect/GroupMember/GroupLinkSelect.cs
```diff
@@ -0,0 +1,126 @@
...
+            bool showAsLink = this.GetAttributeValueFromSelection( "ShowAsLink", selection ).AsBooleanOrNull() ?? false;
+
+            var memberQuery = new GroupMemberService( context ).Queryable();
+
+            IQueryable<string> groupLinkQuery;
+
+            if ( showAsLink )
+            {
+                // Return a string in the format: <a href='/group/{groupId}'>Group Name</a>
+                // Prepend with a HTML comment to ensure that items are sorted correctly.
```
mikepetersonccv: @MrUpsideDown you can use the new SortProperties( string selection ) override to avoid doing this trick for sorting.  Take a look at PersonLinkSelect.cs

Rock/Reporting/DataSelect/GroupMember/GroupMemberAttributeSelect.cs
```diff
@@ -0,0 +1,419 @@
...
+        private const string _CtlProperty = "ddlProperty";
+
+        /// <summary>
+        /// Creates the child controls.
+        /// </summary>
+        /// <param name="parentControl"></param>
+        /// <returns></returns>
+        public override Control[] CreateChildControls( Control parentControl )
+        {
+            var pnlGroupAttributeFilterControls = new Panel();
```
mikepetersonccv: @MrUpsideDown assuming you are disabling ViewState to avoid the ControlTree issue, you could use the Rock.Web.UI.Controls.DynamicControlsPanel which is designed for this specific scenario.  Also, a lot of our Attribute Controls won't work if ViewState is disabled.  You can see an example of DynamicControlsPanel in PropertyFilter.cs

Rock/Reporting/DataSelect/GroupMember/PersonLinkSelect.cs
```diff
@@ -0,0 +1,142 @@
...
+            int displayOrder = this.GetAttributeValueFromSelection( "DisplayOrder", selection ).AsIntegerOrNull() ?? 0;
+            
+            var memberQuery = new GroupMemberService( context ).Queryable();
+
+            IQueryable<string> personLinkQuery;
+            
+            if ( showAsLink )
+            {
+                // Return a string in the format: <a href='/person/{personId}'>LastName, NickName</a>
+                // Prepend with a HTML comment to ensure that items are sorted correctly.
```
mikepetersonccv: @MrUpsideDown yep, another place where you could use the SortProperties override

Rock/Reporting/DataSelect/GroupMember/GroupMemberAttributeSelect.cs
MrUpsideDown: @mikepetersonccv Nothing so complicated, just ignorance on my part! I copied this from some other Rock code, wasn't aware of the alternatives. So I should use DynamicControlsPanel instead of Panel for pnlGroupAttributeFilterControls?

Rock/Reporting/DataSelect/GroupMember/GroupLinkSelect.cs
MrUpsideDown: @mikepetersonccv Thanks - the reason I used this older method is because I had to implement this in our live system, which is running v3. Can I amend this code when we move to v4?

Rock/Reporting/DataSelect/GroupMember/GroupMemberAttributeSelect.cs
mikepetersonccv: yep :)

Rock/Reporting/DataSelect/GroupMember/GroupLinkSelect.cs
mikepetersonccv: @MrUpsideDown I suspected you might be making this for v3.  However, since you are doing a pull request into v4, you can safely make the change now, but feel free to make the call on that.

MrUpsideDown: @mikepetersonccv  I can do it now - just let me know when you are done reviewing and I will submit a new PR with the suggested changes.

mikepetersonccv: yep, everything else looks good

MrUpsideDown:
@mikepetersonccv I've made the changes as suggested. I also added some error handling to GetPropertyType to help resolve some issues I encountered when implementing the SortProperties method.

PR #1313, opened 2015-11-19
Feature/add middle name to person bio

demersus:
We had the requirement to display people's middle names for security reasons.  I added a setting to the bio block (default false), which will enable the middle name to display in title on the Person Bio block.

Is this something you would like in Core?  If not, we'll create a custom block for our installation.


demersus:
This pull request got all kinds of messed up...   Odd.  I will close it and reopen one that is clean.

PR #1320, opened 2015-11-25
When assigned placement group is not in the drop down list

demersus:
This adds the assigned placement group to the dropdown list if it is missing.

I didn't want the assigned placement group unset if the connection request was opened and edited, if it wasn't in the dropdown list.

There are two scenarios that this could be used.
1.) A connection request was opened;  the opportunity type is edited, eliminating the placement group from the list; the connection request is edited afterwards.
2.) A custom workflow action is used to set the placement group.

In our case we will be doing scenario 2, because we need more logic for selecting placement groups beyond a simple drop down.  (We have lots of home groups, and we want to select them based on geolocation).

### DIFF
--- a/RockWeb/Blocks/Connection/ConnectionRequestDetail.ascx.cs
+++ b/RockWeb/Blocks/Connection/ConnectionRequestDetail.ascx.cs
@@ -1330,6 +1330,9 @@ namespace RockWeb.Blocks.Connection
                 }
                 catch
                 {
+                    // Group not in list, let's add it. It must have been assigned elsewhere.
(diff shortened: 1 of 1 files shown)


demersus:
I'm open to discussion on this.
Does the problem I'm trying to solve make sense?
Would you like me to change the implementation...?

azturner:
@demersus , Sorry didn't get back to you sooner. I had just noticed in this pull request that there were some issues with the code around what you were changing (unrelated to your change) that really needed to be updated. For example, there was a DataBind on that dropdownlist even though items were being added manually. It also was not using our SetValue extension method that handles setting a value that is not in the item list.

I added functionality for this and cleaned up the surrounding code in this commit: https://github.com/SparkDevNetwork/Rock/commit/0f83514340e5f444dd5f384e75cbc9f737570744

demersus:
Awesome!  Thank you.  This will allow us to move forward with adopting the new connections system.

- Nik Petersen

PR #1325, opened 2015-12-02
Tweaks/fixes for event list by audience block

sheck:
- If a campus is selected, also show the all campus events, not just the events for only that campus.
- Do not pass inactive events to the lava


sheck:
Please do not merge yet, I have another change I'd like to suggest as well.

PR #1328, opened 2015-12-03
- Add a DateRange context setter to core

dcs619:
@azturner @edmistj I built this block for the dashboards we're using and matched it to the Analytics work you've been doing.

screenshot 2015-12-03 17 18 10

Two deviations from the other context blocks:
- Stores the date range in the current user's preferences (requires the user to be logged in)
- padding: 10px currently hardcoded in the block but should be added to theme.css

If you don't want this in core feel free to close it.


dcs619:
@edmistj any feedback on this PR?

edmistj:
@dcs619 Looks good, we'll pull it in the next day or so.

azturner:
@dcs619 FYI, I removed the site/page setting since it wasn't being used (and would be hard to implement since it's not for an entity being loaded by framework).

PR #1359, opened 2016-01-25
Make the new family control to add a person match the edit family control to add a person

benwiley:
We noticed that users were often overlooking the "Add Person" control on the "Add Family" page.  I made a change to make that control resemble the control used on the "Edit Family" page.

It looks like this:
image


dcs619:
Screenshot from "Edit Family" block:

image

edmistj:
Looks good!

PR #1360, opened 2016-01-25
Person Info Icon in Group Member List

edolyne:
During our testing we found that if a person didn't have a picture associated with their profile the spacing of their name within the group member list seemed off.  After researching we found that it was because there was no image associated with the person's profile, however when you hovered there was a popover that would display additional information about that person.

This change removes the check for a photo and replaces the photo icon with an info icon so that it's clear to the user that they can hover on the icon for more information about that person.

### DIFF
--- a/RockWeb/Blocks/Groups/GroupMemberList.ascx.cs
+++ b/RockWeb/Blocks/Groups/GroupMemberList.ascx.cs
@@ -1018,7 +1018,7 @@ namespace RockWeb.Blocks.Groups
                         m.Person.NickName,
                         m.Person.LastName,
                         Name =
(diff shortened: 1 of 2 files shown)


edmistj:
The goal of adding it was to help staff know who had a photo and who didn't. I don't like the current look either, but would like to keep the ability to know who does have a photo. Would like to make it show a tiny thumbnail (in circle) if they have a photo and the current icon if they don't. 

Haven't had time to implement this. Do you have some bandwidth to add that? If not let me know and I'll try to.

edmistj:
I have a decent start on the real photo solution (would have finished it but my battery died on my flight last night...)

edmistj:
How does this look? 
chrome_2016-01-27_11-03-08

edolyne:
@edmistj Thanks for the clarity on the intended functionality of the icon.  The solution looks fantastic!  It's great to have it always visible, and then inserting the person's image when one is present!

edmistj:
OK, I'll check this into v4.2 as it may appear to be a UI bug to others. It will be moved to develop too so you'll have it in the pre-alpha branch.

edolyne:
@edmistj Thanks so much for doing that!

edmistj:
It's been bothering me too for a while. The fact that it bothered you too was the motivation to just get 'r done.

PR #1362, opened 2016-01-26
Registration - NickName is null and throws a null reference exception

benwiley:
`CurrentPerson.NickName` is null in our registration scenario and causes a null reference exception on the `CurrentPerson.NickName.Trim()` within the `if` statement.

Because `NickName` is not a required field, I opted to assign it to the value of `FirstName` - though there are many other ways this null reference exception could be remedied.

### DIFF
--- a/RockWeb/Blocks/Event/RegistrationEntry.ascx.cs
+++ b/RockWeb/Blocks/Event/RegistrationEntry.ascx.cs
@@ -1438,6 +1438,11 @@ namespace RockWeb.Blocks.Event
 
             if ( newRegistrar )
             {
+                if( CurrentPerson != null && CurrentPerson.NickName == null )
+                {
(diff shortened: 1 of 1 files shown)


azturner:
@benwiley , The core UI does the same exact thing when saving a person, so this wouldn't be an issue for anyone created through Rock, but is a good check for any people that might have been imported without a nick name (hint, hint) :).

https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Model/Person.cs#L1439

dcs619:
@azturner Just double-checked :+1: 

Excavator sets the [nickname to the firstname](https://github.com/NewSpring/Excavator/blob/master/Excavator.FellowshipOne/Maps/People.cs#L310) if it doesn't exist.

The only "people" in our database without nicknames are businesses, but I can't think of a scenario where a business could register for an event.

PR #1366, opened 2016-01-27
Order Groups and Schedules by name in Context Selectors

dcs619:
After getting our production server up, the Order of Groups and Schedules was all over the place.  

We could've changed the order on all our Groups and Schedules but it's easier to have the selector order by Name.

Here's an example of previous behavior:

image


dcs619:
@azturner any feedback on this?

azturner:
Nope, no feedback looks good.

PR #1378, opened 2016-02-03
Update the check-in manager to exclude people who have checked out

dcs619:
Per https://rockrms.slack.com/archives/check-in/p1448843537000032:
>
David Turner [7:32 PM] 
@mrupsidedown: You have several existing fields on the attendance record that you can use to track this type of information. Which fields you choose sort of depends on how you want attendance to be reported for those people that haven't yet shown up at the classroom.  The existing SaveAttendance action in the out-of-the-box workflow updates the StartDateTime, and DidAttend flags whenever a person checks in.  You could clone this action, and update it so that it does not set one or both of these fields until the child actually shows up at the classroom. And there is already an EndDateTime field that can be used to track when a child "checks out" of the classroom.  In addition to those fields, there is a also a QualiferValueId field that you could use to indicate a custom check-in status for each check-in in this scenario. This field is provided for exactly this type of scenario where a church may need to additionally qualify their attendance records using a custom workflow action or block.

This update excludes people who have an EndDateTime from showing up in the Check-in Manager.

Before:

screenshot 2016-02-03 16 55 50

After: 

screenshot 2016-02-03 17 03 05

### DIFF
--- a/RockWeb/Blocks/CheckIn/Manager/Locations.ascx.cs
+++ b/RockWeb/Blocks/CheckIn/Manager/Locations.ascx.cs
@@ -239,6 +239,7 @@ namespace RockWeb.Blocks.CheckIn.Manager
                         a.DidAttend.HasValue &&
                         a.DidAttend.Value &&
                         a.StartDateTime > today &&
+                        !a.EndDateTime.HasValue &&
                         activeScheduleIds.Contains( a.ScheduleId.Value ) &&
                         groupIds.Contains( a.GroupId.Value ) )
                     .Select( a =>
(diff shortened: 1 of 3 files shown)


MrUpsideDown:
@dcs619 Could this be a switchable filter rather than an automatic exclusion? If I am a volunteer in charge of a room using the Check-in Manager, I would also like to know who has recently left the class - for example, if someone is trying to locate a child that recently checked out, how do I know if they ever attended?

dcs619:
@MrUpsideDown I could update this to include recent people in the search box, would that help?  That way they wouldn't show up in the graph but the volunteer could search the child's name and see where they recently checked in.  Alternatively, could your volunteers access the main Rock site to check child history?

I think this code will be revamped when Spark adds check-out functionality to core. I noticed the `Current` tag that shows on a child applies to any check-in with the current date, not just within the current service window.

edmistj:
@dcs619 @MrUpsideDown I think we should only show those that are actively checked in. From the check-in manager you are able to search for anyone and see their recent check-in history which would help to find the missing child.

Issue #1281, opened 2015-10-27
Possible Error w/ Age Filters and Birthdays in Check-In


arranf:
Possible error with check-in very close to a child's birthday if they are about to age out of a group. I'll attempt to reproduce.

http://www.rockrms.com/Rock/Ask/Troubleshooting/Questions/579

dcs619:
@arranf Did you ever reproduce?  We had a user report this just this week.

arranf:
@dcs619 I did try but couldn't. Didn't spend a whole lot on this could definitely use further testing.

edmistj:
@dcs619 can you provide a update on this issue? Have you been able to reproduce it. Any insight our repeat steps would be helpful.

dcs619:
@edmistj Yes, there's more context at #1385 and 0b4c7e1041cf6591e7771e311a47a1013cdb77c3#r22149095.  It's hard to reproduce the edge case since it's specific to the datetime being calculated.

lpeterson1:
This fix has changed the behavior of the age ranges in a way that is likely to catch some users off guard. Previously, instead of entering 3.999.... as the end date of a range, some have entered "4", which was interpreted by Rock as 4.000.... With this fix, 4 now _includes_ 4 yos, so the range is actually 3 to 4.999... If these users aren't aware of that (and don't change their age ranges) unexpected choices will be presented on check-in (causing confusion and delay, and kids will potentially end up in wrong rooms).

3.000000 to 4.000000 seems to eliminate 4 yos, but 3.0 to 4.0 includes at least a month's worth of 4 yos.

I don't have any idea how widespread this is likely to be, but I know of some cases. It seems like this should at least be looked at, since it could cause significant weekend disruption for some users.

mikejed:
For posterity, someone mentioned in Slack that the correct solution is to set the range to `3 to 3` rather than using 4.000000000 to eliminate 4 year olds. However, to Lee's point, this is a change that people will have to actively look for in their installations.

lpeterson1:
To add more confusion, I was working with someone yesterday who had a group with age range 1.5 to 2. I tested, and this range will now allow all 2-year-olds into the room. A range of "1.5 to 1" works as expected, but certainly is counterintuitive. "1.5 to 2.0" allows about .1 year's worth of 2-year-olds to check in. I'm sure 2.0000... (or 1.999....) would end the range as expected, but that's what we were trying to avoid.

PR #1385, opened 2016-02-08
- Fix issue with check-in "randomly" not allowing check-ins

dcs619:
See the following;
- https://github.com/SparkDevNetwork/Rock/issues/1281
- http://www.rockrms.com/Rock/Ask/Troubleshooting/Questions/579
- https://github.com/NewSpring/Rock/issues/151

TL;DR in certain edge cases, an AgePrecise of 2.9955191256830603 would get filtered out of a group with an AgeRange of 2.99.  Fixed with PR.

image

I tracked down the issue to the AgePrecise value.  Since double is stored as a fractional number in C#, the precision is only limited by the data type and was being calculated with 16 degrees of precision.  Conversely, most of our Groups were using an AgeRange precision of 0 (and usually a max precision of 2).  

In order to consistently compare the ages, I adjusted the local AgePrecise data type to a decimal and limited the precision to whatever the group AgeRange is using.  This means a group can be as flexible as needed for the MinAge *or* the MaxAge:

image

I'm also using [MidpointRounding.ToEven](https://msdn.microsoft.com/en-us/library/system.midpointrounding%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) so that a child of 2.9999562 would still be included in a group that allows 2.99.

### DIFF
--- a/Rock/Workflow/Action/CheckIn/FilterGroupsByAge.cs
+++ b/Rock/Workflow/Action/CheckIn/FilterGroupsByAge.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.ComponentModel.Composition;
+using System.Globalization;
 using System.Linq;
 
 using Rock.Attribute;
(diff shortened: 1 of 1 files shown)


arranf:
Dude that is some impressive work tracking that down.

Issue #1386, opened 2016-02-09
Image / File API Design


jbaxleyiii:
After a few days of toying around with some ideas and test, I've determined using ~/ImageUploader and ~/FileUploader aren't possible from remote applications.

All in all, I think thats a great thing. Since they aren't part of the API, if they worked from remote locations, anyone could hijack your application and start uploading whatever they wanted which would be awful. We still want the benefit of users being able to choose their file / image storage methods so here is what I'm thinking might work. I'm 1000% open to ideas, this may be the wrong direction.

# Use Case
#### Client
An application (native or web), that posts an image or file to update a users avatar, add a file, upload a video (in the future), etc.

#### Server
An application with image / file manipulation abilities (like resizing, bluring, etc) uploads corresponding binary as a spoofed client side multipart form

## Endpoint
| Method | Relative Path |
|:---------- |:------------------ |
| POST | `/api/BinaryFiles/Upload?fileTypeGuid={fileTypeGuid}&fileId={fileId}` |

## Sample Client Post
```javascript
var input = document.querySelector('input[type="file"]')

var data = new FormData()
data.append('file', input.files[0])

fetch('https://rock.newspring.cc/api/BinaryFiles/Upload?fileTypeGuid={guid}&fileId={currentUsersPhotoId}', {
  method: 'post',
  headers: { "Authorization-Token": "client-token" },
  body: data
})
  .then((response) => {
    return response.json()
   })
  .then((id) => {
    console.log(id)
  })
```

## Sample Server Post
```javascript
// this would look the same as the client on Rock's end, but can be done using the methods explored here https://github.com/rfrench/poster
var poster = require('poster');

var options = {
  uploadUrl: 'https://rock.newspring.cc/api/BinaryFiles/Upload?fileTypeGuid={guid}&fileId={currentUsersPhotoId}',
  method: 'POST'
};

poster.post('file.jpg', options, function(err, data) {
  if (!err) {
    console.log(data);
  }
});
```

cc @dcs619 @benwiley @edmistj @azturner @mikepetersonccv

dcs619:
@edmistj Any feedback on this?  We're going to look at implementing a solution next week.

edmistj:
@dcs619  Looks good to me.

dcs619:
@jbaxleyiii you should be able to close this issue now :boom:

edmistj:
Thanks for the great addition to the API!

jbaxleyiii:
[...]

PR #1404, opened 2016-02-22
File upload endpoint for BinaryFilesController

benwiley:
- Add upload endpoint to binary files controller.  The file type can be passed as an id or guid.  The file is uploaded and an id is returned if successful. There are various error messages for invalid types, invalid files, etc and I tried to be specific with the messages.

- Move EXIFextractor DLL from RockWeb to Rock for use in new ImageUtilities class

- Created static ImageUtilities class within Rock.Utilities.  This has the functionality previously in ImageUploader that was in RockWeb.  The move allows use of this code both in RockWeb and Rock.Rest for the upload endpoint of binary files controller.

- Reduce ImageUploader to nothing but a call to the ImageUtilities since that code was moved.

- Ignore files directory.  When debugging locally, I was setting people's profile images.  Without this ignore setting, the profile images could be committed.

Would appreciate feedback to ensure this is right before merging in!

With this implementation, steps to set a profile image:
1.) POST to api/BinaryFiles/Upload to upload file and get it's id
2.) PATCH to api/People/X to set PhotoId
3.) DELETE to api/BinaryFiles/X to delete old profile photo (if applicable)

cc @jbaxleyiii @dcs619


dcs619:
Implements #1386

jbaxleyiii:
@dcs619 @benwiley can we cherry-pick this and deploy it to alpha?

Rock.Rest/Controllers/BinaryFilesController.Partial.cs
```diff
@@ -0,0 +1,166 @@
...
+                var binaryFileType = new BinaryFileTypeService( rockContext ).Get( binaryFileTypeId );
+
+                if ( uploadedFile == null )
+                {
+                    GenerateResponse( HttpStatusCode.BadRequest, "No file was sent" );
+                }
+
+                if ( binaryFileType == null )
+                {
+                    GenerateResponse( HttpStatusCode.InternalServerError, "There is no person image file type" );
```
azturner: Probably should be more generic (i.e. "Invalid binary file type")

Rock.Rest/Controllers/BinaryFilesController.Partial.cs
```diff
@@ -0,0 +1,166 @@
...
+                }
+
+                var currentUser = UserLoginService.GetCurrentUser();
+                var currentPerson = currentUser != null ? currentUser.Person : null;
+
```
azturner: The base ApiController class has a GetPerson() method that should be used to get the person.

azturner:
@benwiley , I Like that you moved the contents of the GetFileContentStream method in the ImageUploader to a helper class, but it looks like the ValidateFileType override got deleted too. Don't we still need that?  

Noticed too that the AppVeyor build failed for this too due to missing EXIFextractor.dll. might need to add that dll (and modify a .gitignore so it gets included).

benwiley:
@azturner I made those changes.  Thanks for your feedback!

PR #1409, opened 2016-02-26
Filters for the Dynamic Data Block

benwiley:
appliedfilters
filtercontrols

- Adds an option to Dynamic Data Block's options that turns filters on or off (off by default to maintain current functionality)

- When enabled, a filter is available for each column within the dynamic data.  Boolean columns are filterable with a drop down list that presents blank (no filter), yes (show only true values), or no (show only false values).  Date columns use a date range picker control.  All other columns use a text box.  String fields filter as a "begins with" filter.  All other fields filter for an exact match of the text box contents.

- Adds `DateRangePicker.TryParse` that mimics the functionality of `DateTime.TryParse` but for delimited values generated from the `DateRangePicker`

- Fixes issue where calling `GridFilter.DeleteUserPreferences` was not actually deleting preferences from the page because the key sent to `RockBlock.DeleteUserPreference` didn't include the prefix

- One issue is when the filter options are first turned on from the dynamic data block options, this error occurs.  It goes away after refreshing the page and never comes back even after a query change

error

### DIFF
--- a/RockWeb/Blocks/Reporting/DynamicData.ascx.cs
+++ b/RockWeb/Blocks/Reporting/DynamicData.ascx.cs
@@ -488,17 +488,15 @@ namespace RockWeb.Blocks.Reporting
 
                             phContent.Controls.Add( div );
 
-                            if ( useDynamicFilterControls )
+                            GridFilter = new GridFilter()
                             {
-                                GridFilter = new GridFilter()
-                                {
-                                    ID = "gfFilter"
-                                };
+                                ID = "gfFilter"
+                            };
 
-                                div.Controls.Add( GridFilter );
-                                GridFilter.ApplyFilterClick += ApplyFilterClick;
-                                GridFilter.DisplayFilterValue += DisplayFilterValue;
-                            }
+                            div.Controls.Add( GridFilter );
+                            GridFilter.ApplyFilterClick += ApplyFilterClick;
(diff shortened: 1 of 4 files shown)


edmistj:
@benwiley Wow, that's really cool!

RockWeb/Blocks/Reporting/DynamicData.ascx.cs
```diff
@@ -634,6 +779,103 @@ private void SortTable( Grid grid, DataTable dataTable )
...
+                if ( control is DateRangePicker )
+                {
+                    var dateRangePicker = control as DateRangePicker;
+                    var minValue = dateRangePicker.LowerValue;
+                    var maxValue = dateRangePicker.UpperValue;
+                    var colName = string.Format( "[{0}]", dateRangePicker.ID.Remove( 0, 3 ) );
+
+                    if ( minValue.HasValue )
+                    {
+                        query.Add( string.Format( "{0} > #{1}#", colName, minValue.Value ) );
```
mikepetersonccv: @benwiley you might want to change the '>' to '>='

benwiley:
@mikepetersonccv good catch!  Here it is now:
datefilter

benwiley:
By hiding the filter panel rather than creating it when needed, I was able to solve this problem:

error

I tried to not create the gridfilter on postback but that made the filters not functional, as if their event handlers were not attached.

PR #1414, opened 2016-03-01
Upgrade google analytics tracker to latest

sheck:
> Universal Analytics is the new operating standard for Google Analytics; however, many sites and applications still use outdated tracking libraries to send data to Google Analytics.

> It is strongly recommended that you upgrade to use a Universal Analytics tracking library such as analytics.js or the latest mobile SDKs for Android and iOS. Non-Universal Analytics libraries, including ga.js or v1.x of the mobile SDKs, will be deprecated sometime in the near future.

— https://developers.google.com/analytics/devguides/collection/upgrade/

Also see: https://support.google.com/analytics/answer/2790010?hl=en


edmistj:
@sheck Thanks Nick!

PR #1423, opened 2016-03-09
API Transactions by Giving Group

benwiley:
- Add endpoint to get a person's giving group's transactions.  This mimics the TransactionList block functionality with a person context (https://github.com/SparkDevNetwork/Rock/blob/develop/RockWeb/Blocks/Finance/TransactionList.ascx.cs#L906)
- Add `DataMember` to AuthorizedPersonAlias to allow `$expand` ODATA operation

### DIFF
--- a/Rock.Rest/Controllers/FinancialTransactionsController.Partial.cs
+++ b/Rock.Rest/Controllers/FinancialTransactionsController.Partial.cs
@@ -317,29 +317,25 @@ namespace Rock.Rest.Controllers
         }
 
         /// <summary>
-        /// Gets the contribution transactions.
+        /// Gets transactions by people with the supplied givingId.
         /// </summary>
-        /// <param name="groupId">The group unique identifier.</param>
+        /// <param name="givingId">The giving ID.</param>
         /// <returns></returns>
         /// <exception cref="System.Web.Http.HttpResponseException"></exception>
         [Authenticate, Secured]
         [HttpGet]
         [EnableQuery]
-        [System.Web.Http.Route( "api/FinancialTransactions/GetTransactionsForGivingGroup/{personId}" )]
-        public IQueryable<FinancialTransaction> GetTransactionsForGivingGroup( int personId )
+        [System.Web.Http.Route( "api/FinancialTransactions/GetByGivingId/{givingId}" )]
+        public IQueryable<FinancialTransaction> GetByGivingId( string givingId )
         {
-            var rockContext = new RockContext();
-            var personService = new PersonService( rockContext );
-            var person = personService.Get( personId );
-
-            if ( person == null )
+            if ( string.IsNullOrWhiteSpace( givingId ) || !( givingId.StartsWith( "P" ) || givingId.StartsWith( "G" ) ) )
             {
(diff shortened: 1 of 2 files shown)


Rock.Rest/Controllers/FinancialTransactionsController.Partial.cs
```diff
@@ -317,7 +317,33 @@ public DataSet GetContributionTransactions( int groupId, int? personId, [FromBod
...
+        /// </summary>
+        /// <param name="groupId">The group unique identifier.</param>
+        /// <returns></returns>
+        /// <exception cref="System.Web.Http.HttpResponseException"></exception>
+        [Authenticate, Secured]
+        [HttpGet]
+        [EnableQuery]
+        [System.Web.Http.Route( "api/FinancialTransactions/GetTransactionsForGivingGroup/{personId}" )]
+        public IQueryable<FinancialTransaction> GetTransactionsForGivingGroup( int personId )
+        {
```
azturner: @benwiley, Does your app already depend on this method/parameter? A better name for the method might be `GetByPersonGivingId` since that is what is being used to query the transactions (there is also a GivingGroupId on person which is why it's a bit confusing).  

Also, if the calling app already has a person object (with a GivingId value already), an even better idea would be to just pass the GivingId instead of PersonId and call the method `GetByGivingId( int givingId )`

benwiley: @azturner Yes our app depends on this to show giving history.  The idea was that it would return all transactions for the giving group that the supplied person is a part of.  I thought the GivingId was the same concept as the GivingGroupId?  I just checked with @jbaxleyiii and the GivingId is not currently mapped in the client app.  

I'll be glad to call it whatever you want so long as we are all clear on what it does (it sounds like I may be confused about the GivingGroupId vs GivingId)

jbaxleyiii: @benwiley @azturner the `GetTransactionsForGivingGroup` feature has not been shipped yet. If `GetByGivingId( int givingId )` is an easier and more performant method, I can start to send the `GivingId` to the client for this call.

benwiley: @jbaxleyiii it would be more performant since I wouldn't have to query for the person first thing.  @azturner I will make the change - just need to know if we mean `int GivingGroupId` or `string GivingId`.  Plus I would like to know if those are the same concept.

azturner: @benwiley , sorry yeah it should use 'GivingId' and that is a string. So method should be `GetByGivingId( string givingId )`

They are similar concepts. If person has selected to give with their family, their GivingGroupId is the id of their family group, and their GivingId is that same id with a 'G' prefix. If they have selected not to give with their family, then their GivingGroupId is null, and their GivingId is their person id with a 'P' prefix. So you could never query by GivingGroupId as many people would have a null value, but different GivingIds.

jbaxleyiii: @azturner thanks for the clarifications :+1:

PR #1425, opened 2016-03-10
Add a test username endpoint by query string

benwiley:
We were running into issues with checking the availability of a username that had a `.pl` in it.  We think this was because it looks like a file extension.  For example `phil.plex` would cause the API call to fail.  This endpoint seems to fix the issue by passing the username as a query string.


benwiley:
This actually didn't fix the problem

PR #1443, opened 2016-03-21
Added 'Warning' property to Rock Controls similar to 'Help' Property.

tcavaletto:


dcs619:
How does this differ from the existing `Warning` notification that fires when a form doesn't pass validation?  Is it intended to replace it?

edmistj:
It places a little exclamation point near the label to provide a custom warning message. It's more of an instructional warning than a validation error. Something like 'If you set this value no email will be sent.'

Issue #1403, opened 2016-02-18
Group Finder Deletes Details Setting


BelAirIT:
When specifying a details or registration page under Linked Pages, the block resets back to blank after a page refresh.http://alpha.rockrms.com/page/5384

BelAirIT:
Logged in users seem to still be directed to the correct pages while non-authenticated users get a 404 error page. Another strange behavior is that deleting the Group Detail block from the page specified in the Group Detail setting of the Group Finder block causes it to send the user on correctly.

matthewmeekins:
I've seen this behavior as well on 4.3. However, I had a page already setup before the update and it still works even though the page settings are blank.

PR #1454, opened 2016-03-28
Fixed settings not saving and added linked pages to lava

trydyingtolive:
Fixes issue #1403 of linked pages not saving.
Fixes issue of Show Grid control showing the value of Show Map.
Added linked pages to Lava Output so both InfoWindow and LavaOverview get linked page references (before only InfoWindow did).
Applied standard code formatting


edmistj:
Looks good, thanks!

PR #1456, opened 2016-03-28
[EN] Update the dynamic data block with better filters

dcs619:
- Updates the dynamic data filter with DBA-friendly syntax
- Allows filtering by any text in the column, not just at the beginning
- Sets the filter visibility on by default 

@edmistj any issues with item 3?  Every time we've turned the filter on for a DD block, users have been ecstatic.


edmistj:
@dcs619 I think #3 is good. Could we move the block setting though to the edit screen. Most of the meat of that block is configured on the custom edit screen.

dcs619:
@edmistj Should I move the `Paneled Grid` attribute as well?  It's the only one not in the custom edit screen.

edmistj:
Sure, do you mean 'Update Page' though?

dcs619:
@edmistj I was referring to [this](https://github.com/SparkDevNetwork/Rock/blob/develop/RockWeb/Blocks/Reporting/DynamicData.ascx.cs#L65).  In any case those changes are pushed.

PR #1459, opened 2016-03-31
[EN] Better Ages on the Person Profile for Newborns and Infants

arranf:
The current ages for any child under the age of 1 just appear as 0 which doesn't provide meaningful information. If a family visitor wanted to know the age of the child they'd have to do the maths themselves. This pull request adds a new method to the Person model for getting a string with a formatted age.

E.G. 12 yrs old, 6 mnths old, 14 days old

Comparison images below.

withoutdateformatchange
withdateformatchange

### DIFF
--- a/Rock/Model/Person.cs
+++ b/Rock/Model/Person.cs
@@ -1010,6 +1010,30 @@ namespace Rock.Model
         }
 
         /// <summary>
+        /// Returns a formatted age in years, months, or days.
+        /// </summary>
+        /// <returns></returns>
+        public string FormatAge()
+        {
+            if (Age > 0)
+            {
+                return Age + (Age == 1 ? " yr old " : " yrs old ");
+            }
+            var today = RockDateTime.Today;
+            if (BirthMonth != null && BirthMonth < today.Month)
+            {
+                int months = today.Month - BirthMonth.Value;
+                return months + (months == 1 ? " mnth old " : " mnths old ");
+            }
+            if (BirthDay != null)
+            {
+                int days = today.Day - BirthDay.Value;
+                return days + (days == 1 ? " day old " : " days old ");
+            }
+            return string.Empty;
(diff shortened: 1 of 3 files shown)


azturner:
@arranf We like the idea of showing months/days for those people that are less than a year old, but can you have the text be different depending on whether it is being displayed on the bio block vs, the groupmembers block. Maybe add a `bool condensed = false` parameter to that FormatAge method. These are example values for each block:

Bio Block
1 day old
2 days old
1 mo old
2 mos old
1 yr old
2 yrs old

GroupMember Block
1
2
1 day
2 days
1 mo
2 mos

Rock/Model/Person.cs
```diff
@@ -1010,6 +1010,30 @@ private set
...
         /// <summary>
+        /// Returns a formatted age in years, months, or days.
+        /// </summary>
+        /// <returns></returns>
+        public string FormatAge()
+        {
+            if (Age > 0)
+            {
+                return Age + (Age == 1 ? " yr old " : " yrs old ");
+            }
```
azturner: Also, the `Age` property on person is not a stored value. It does an age calculation from the birthdate property whenever it is referenced. Because you're referencing this property three times, it would be better to only use it once and save it to a local int? variable before doing the comparisons. Also, it may return a null value, be sure to check for that.

azturner:
@arranf , are you able to make these changes?

arranf:
@azturner Done

mikepetersonccv:
@arranf There seems to be an issue if a person is less than a year old, but the current month is before their birthmonth
birthdate_age

arranf:
@mikepetersonccv Still averaging a 0.5 bug rate on my pull requests. I'll take a look at this today.

mikepetersonccv:
@arranf OK Cool :)

Issue #1355, opened 2016-01-20
[EN] Rename IsSpecialNeeds to HasSpecialNeeds


dcs619:
Our kids ministry would prefer marking children with "Has Special Needs" instead of "Is Special Needs".  

If this change won't break any custom code (@nairdo?), I can submit a PR to change [FilterGroupsBySpecialNeeds](https://github.com/SparkDevNetwork/Rock/blob/1974c937c8219ba311565440faad4bf6dde7176a/Rock/Workflow/Action/CheckIn/FilterGroupsBySpecialNeeds.cs) and a migration to update the Attribute name/key in the database.

nairdo:
I think it would be a great time to also change that filter to use a configurable setting instead of hard coding the attribute key. Then, also, if it's a core attribute it should be prefixed with the new "core." naming convention.

nairdo:
@dcs619 Here is the write-up that would be added to the Rock v5 check-in guide.  I'll modify it after you make the needed changes and add it to the book.
[Checkin - Advanced Feature - Special Needs Handling.docx](https://github.com/SparkDevNetwork/Rock/files/100861/Checkin.-.Advanced.Feature.-.Special.Needs.Handling.docx)

dcs619:
@edmistj Any preference re adding a HasSpecialNeeds person attribute to core?  Based on the documentation @nairdo posted, users need to add that attribute themselves.

edmistj:
I'm open if you think it would be better to have it exist in core.

dcs619:
It sounds like @nairdo and I are the only ones using it, so ¯\_(ツ)_/¯

It does seem weird to have a core check-in filter (see PR) that's dependent on a non-core attribute.

edmistj:
@nairdo Can you add a migration to add this person attribute. No hurry as this will be a v5 feature.

dcs619:
@nairdo Here are the settings we're using:

![screenshot 2016-02-03 11 19 35](https://cloud.githubusercontent.com/assets/1210933/12788645/48b4ea44-ca68-11e5-932a-437453ce90e7.png)

We changed from a Boolean data type to Single-Select as `No` means the same thing as nothing selected.

dcs619:
@nairdo I have an updated version of this workflow action that allows you to select the Person/Group Special Needs attribute on the action, instead of depending on matching key string.

Did you ever add a migration for the person attribute?  If so, I can set the default GUID for the action before I push it.

PR #1460, opened 2016-03-31
[EN] Remove hardcoded attribute keys from the Special Needs filter

dcs619:
See discussion at #1355 (@nairdo).  This removes the dependency that the Special Needs key has to match its name, minus whitespace.

image

There isn't a Person Special Needs attribute currently in core, so the GUID provided for the default attribute will show up blank on other systems until one is selected (or we add one to core).


Rock/Workflow/Action/CheckIn/FilterGroupsBySpecialNeeds.cs
```diff
@@ -34,10 +35,11 @@ namespace Rock.Workflow.Action.CheckIn
...
-    [BooleanField( "Remove (or exclude) Special Needs Groups", "If set to true, special-needs groups will be removed if the person is NOT special needs. This basically prevents non-special-needs kids from getting put into special needs classes.  Default true.", true, key: "RemoveSpecialNeedsGroups" )]
-    [BooleanField( "Remove (or exclude) Non-Special Needs Groups", "If set to true, non-special-needs groups will be removed if the person is special needs.  This basically prevents special needs kids from getting put into regular classes.  Default false.", false, key: "RemoveNonSpecialNeedsGroups" )]
-    [BooleanField( "Remove", "Select 'Yes' if groups should be be removed.  Select 'No' if they should just be marked as excluded.", true )]
+    [AttributeField( Rock.SystemGuid.EntityType.PERSON, "Person Special Needs Attribute", "Select the attribute used to filter special needs people.", true, false, "8B562561-2F59-4F5F-B7DC-92B2BB7BB7CF", order: 1 )]
+    [AttributeField( Rock.SystemGuid.EntityType.GROUP, "Group Special Needs Attribute", "Select the attribute used to filter special needs groups.", true, false, "9210EC95-7B85-4D11-A82E-0B677B32704E", order: 2 )]
```
azturner: These guids ( `8B562561-2F59-4F5F-B7DC-92B2BB7BB7CF`, `9210EC95-7B85-4D11-A82E-0B677B32704E` ) don't look like they exist in the default install, so probably shouldn't default to them.

dcs619: @azturner Correct, core doesn't have any `Special Needs` attributes by default.  The only mention of it was in this filter, which initially relied on the hardcoded `IsSpecialNeeds` AttributeKey.

I thought @nairdo was planning to add a migration for Person.HasSpecialNeeds and Group.HasSpecialNeeds, but haven't seen anything yet.  I can add it to my list of things to do.

azturner: @dcs619 Ok, that makes sense. We'll pull this in, and when @nairdo creates the migration, he can use these guid values.

nairdo:
@dcs619 I'm getting in on this now.  I'll be creating the migration for you guys for these two attributes using those two guids.

Rock/Workflow/Action/CheckIn/FilterGroupsBySpecialNeeds.cs
nairdo: @dcs619 Exactly which group type are you expecting/wanting the group special needs attribute defined on?  My thoughts are you want it on the "Check in by Age" type groups since all check-in groups inherit from this one.
rockcheckingrouptypegroupstructure_20130815b

nairdo:
Change of plans. After reviewing this today, we've decided to deprecate this filter and add a new, more generic reusable 'Filter Groups By Attribute' filter in 6.0.  This new filter will put the decision on the value that is set in the group/grouptype instead of the 'policy' approach the current filter requires to adopt when you set it in the workflow configuration. 

So the only additional changes we will make with this particular filter are:

- to remove the hard-coded default guids
- to change the name to "Filter Groups By Special Needs (discontinue use)"

@dcs619 if you want to continue using this past 6.0, feel free to recreate as a custom filter.  Otherwise you can move to the generic one.

PR #1461, opened 2016-03-31
[EN] Remove hardcoded attribute keys from Filter Groups By Age and fix edge case

dcs619:
This removes the dependency that "AgeRange" will always be the key for a group age range attribute.

Also fixes an edge case related to #1385 by adding a Decimal GetPrecision extension. The issue is for an age of 5.97, which MidpointRounding causes to round up to the nearest even number (6).

### DIFF
--- a/Rock/Workflow/Action/CheckIn/FilterGroupsByAge.cs
+++ b/Rock/Workflow/Action/CheckIn/FilterGroupsByAge.cs
@@ -29,12 +29,13 @@ namespace Rock.Workflow.Action.CheckIn
     /// <summary>
     /// Removes (or excludes) the groups for each selected family member that are not specific to their age.
     /// </summary>
+    [ActionCategory( "Check-In" )]
     [Description( "Removes (or excludes) the groups for each selected family member that are not specific to their age." )]
     [Export( typeof( ActionComponent ) )]
     [ExportMetadata( "ComponentName", "Filter Groups By Age" )]
     [BooleanField( "Remove", "Select 'Yes' if groups should be be removed.  Select 'No' if they should just be marked as excluded.", true, "", 0 )]
     [BooleanField( "Age Required", "Select 'Yes' if groups with an age filter should be removed/excluded when person does not have an age.", true, "", 1 )]
-    [AttributeField( "9BBFDA11-0D22-40D5-902F-60ADFBC88987", "Group Age Range Attribute", "Select the attribute used to define the age range of the group", true, false, "43511B8F-71D9-423A-85BF-D1CD08C1998E", order: 2 )]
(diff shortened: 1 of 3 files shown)


Rock/Workflow/Action/CheckIn/FilterGroupsByAge.cs
```diff
@@ -18,24 +18,23 @@
...
 using Rock.Data;
+using Rock.Web.Cache;
 
 namespace Rock.Workflow.Action.CheckIn
 {
     /// <summary>
     /// Removes (or excludes) the groups for each selected family member that are not specific to their age.
     /// </summary>
-    [ActionCategory( "Check-In" )]
     [Description( "Removes (or excludes) the groups for each selected family member that are not specific to their age." )]
```
azturner: @dcs619 Curious why the [ActionCategory] attribute was removed?

Rock/Workflow/Action/CheckIn/FilterGroupsByAge.cs
```diff
@@ -18,24 +18,23 @@
...
     [Export( typeof( ActionComponent ) )]
     [ExportMetadata( "ComponentName", "Filter Groups By Age" )]
-
     [BooleanField( "Remove", "Select 'Yes' if groups should be be removed.  Select 'No' if they should just be marked as excluded.", true, "", 0 )]
-    [BooleanField( "Age Required", "Select 'Yes' if groups with an age filter should be removed/excluded when person does not have an age.", true, "", 1)]
+    [BooleanField( "Age Required", "Select 'Yes' if groups with an age filter should be removed/excluded when person does not have an age.", true, "", 1 )]
+    [AttributeField( "9BBFDA11-0D22-40D5-902F-60ADFBC88987", "Group Age Range Attribute", "Select the attribute used to define the age range of the group", true, false, "43511B8F-71D9-423A-85BF-D1CD08C1998E", order: 2 )]
```
azturner: Might consider using the guid constants when we have them instead of the guid string. i.e.`Rock.SystemGuid.EntityType.GROUP` instead of `9BBFDA11-0D22-40D5-902F-60ADFBC88987`. Makes it a bit more readable.

Rock/Workflow/Action/CheckIn/FilterGroupsByAge.cs
dcs619: Strange!  I can't find it being removed in any of our recent commits.  I added it back.

Rock/Workflow/Action/CheckIn/FilterGroupsByAge.cs
dcs619: Done!

PR #1462, opened 2016-03-31
Update ScheduledTransactionSummary to include all schedules with same giving leader id.

Tschrock:
This fixes the summary to match what the list blocks are doing. (Like 5309be58abd38b885a0a3800b3fb83e7faa26b44)

### DIFF
--- a/RockWeb/Blocks/Finance/ScheduledTransactionSummary.ascx.cs
+++ b/RockWeb/Blocks/Finance/ScheduledTransactionSummary.ascx.cs
@@ -118,7 +118,7 @@ namespace RockWeb.Blocks.Finance
                 FinancialScheduledTransactionService transactionService = new FinancialScheduledTransactionService( rockContext );
 
                 var schedules = transactionService.Queryable( "ScheduledTransactionDetails.Account" )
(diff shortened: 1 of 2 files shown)


RockWeb/Blocks/Finance/ScheduledTransactionSummary.ascx.cs
```diff
@@ -118,7 +118,7 @@ private void ShowContent()
                 FinancialScheduledTransactionService transactionService = new FinancialScheduledTransactionService( rockContext );
 
                 var schedules = transactionService.Queryable( "ScheduledTransactionDetails.Account" )
-                                .Where( s => s.AuthorizedPersonAlias.PersonId == CurrentPerson.Id && s.IsActive == true );
+                                .Where( s => s.AuthorizedPersonAlias.Person.GivingLeaderId == CurrentPerson.GivingLeaderId && s.IsActive == true );
```
azturner: @Tschrock , Nice Catch! Can you switch this to use GivingId though instead of GivingLeaderId. Both are computed columns based on the GivingGroupId value, but GivingLeaderId is using a user-defined function in its computed column and does not have an index. GivingId does not use a function, and has an index (it will be faster).

We have GivingLeaderId for those times you actually need to know who the "leader" is, but in this case you just need to group the people, so GivingId would be better.

PR #1468, opened 2016-04-04
Attempting to save invalid html in the HtmlContentDetail block now displays a warning message detailing the errors.

tcavaletto:
The first time a user clicks the 'Save' button a warning box will appear listing any HTML errors should there be any. The warning box will prompt the user to click the 'Save' button again should the errors be intentional, after which the new content will be saved.

image


dcs619:
That's awesome!

Issue #1467, opened 2016-04-04
Adding a new site causes a database timeout


trydyingtolive:
Creating a new site in develop causes the following error:

```
Error
The wait operation timed out


   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   at System.Data.SqlClient.SqlDataReader.get_MetaData()
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString)
   at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, SqlDataReader ds, Boolean describeParameterEncryptionRequest)
   at System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method, TaskCompletionSource`1 completion, Int32 timeout, Task& task, Boolean asyncWrite)
   at System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method)
   at System.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior, String method)
   at System.Data.SqlClient.SqlCommand.ExecuteDbDataReader(CommandBehavior behavior)
   at System.Data.Common.DbCommand.ExecuteReader(CommandBehavior behavior)
[...]

PR #1469, opened 2016-04-04
#1467 Passed RockContext through to fix database timeout

trydyingtolive:
This is a possible fix to issue #1467. I solved the problem by passing an optional RockContext into each method. This may or may not be the preferred solution.

### DIFF
--- a/Rock/Security/Authorization.cs
+++ b/Rock/Security/Authorization.cs
@@ -1,5 +1,5 @@
 ﻿// <copyright>
-// Copyright 2013 by the Spark Development Network
+// Copyright by the Spark Development Network
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


Rock/Security/Authorization.cs
```diff
@@ -1,5 +1,5 @@
 ﻿// <copyright>
-// Copyright by the Spark Development Network
+// Copyright 2013 by the Spark Development Network
```
azturner: We recently removed the year from all of the copyright statements, looks like it somehow came back here.

Rock/Security/Authorization.cs
```diff
@@ -166,8 +166,13 @@ public static bool Load()
...
         /// <param name="entityTypeId">The entity type identifier.</param>
         /// <param name="entityId">The entity identifier.</param>
-        public static void ReloadEntity( int entityTypeId, int entityId )
+        public static void ReloadEntity( int entityTypeId, int entityId, RockContext rockContext = null )
         {
+            if ( rockContext == null )
+            {
+                rockContext = new RockContext();
+            }
+
```
azturner: This check for null and creation of context here isn't really needed since it's just being passed to `RefreshEntity()` and that will create it if it's null, and with the `using` which is even better.

azturner:
@trydyingtolive Yep, we can't create a new RockContext inside of a `.WrapTransaction`. Nice catch!. Just made a few comments. If you could update those, we'll pull this in.

PR #1470, opened 2016-04-04
- Added security to the SearchField control to limit the SearchComponents people can view.

tcavaletto:
Currently, even though an administrator can limit security on a SearchComponent, anyone is able to view and search by it. This code change uses the SearchComponents' security settings to dictate who can actually view them, while adding .12 milliseconds to the page's load time. Below are two examples of users with the SearchComponents they can search by:

Admin Admin, a user with view access to the BusinessName SearchComponent:
image

Ted Decker, a user without view access to the BusinessName SearchComponent:
image

### DIFF
--- a/Rock/Web/UI/Controls/SearchField.cs
+++ b/Rock/Web/UI/Controls/SearchField.cs
@@ -24,6 +24,8 @@ using System.Web.UI.HtmlControls;
 using System.Web.UI.WebControls;
 
 using Rock;
+using Rock.Model;
+using Rock.Security;
 
 namespace Rock.Web.UI.Controls
 {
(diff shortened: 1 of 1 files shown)


Rock/Web/UI/Controls/SearchField.cs
```diff
@@ -58,13 +60,25 @@ public override void RenderControl( HtmlTextWriter writer )
...
+            if ( rockPage != null )
             {
-                if ( !service.Value.Value.AttributeValues.ContainsKey( "Active" ) || bool.Parse( service.Value.Value.AttributeValues["Active"].Value ) )
+                var currentPerson = rockPage.CurrentPerson;
+                if ( currentPerson != null )
                 {
-                    searchExtensions.Add( service.Key.ToString(), Tuple.Create<string, string>( service.Value.Value.SearchLabel, service.Value.Value.ResultUrl ) );
-                    if ( string.IsNullOrWhiteSpace( hfFilter.Value ) )
-                        hfFilter.Value = service.Key.ToString();
+                    foreach ( KeyValuePair<int, Lazy<Rock.Search.SearchComponent, Rock.Extension.IComponentData>> service in Rock.Search.SearchContainer.Instance.Components )
```
azturner: Security may be configured to allow access even if current person is null (i.e. All Users). Instead of checking for a valid currentPerson, just pass the rockPage.CurrentPerson value to the IsAuthorized method. It will handle the null value.

Rock/Web/UI/Controls/SearchField.cs
```diff
@@ -58,13 +60,25 @@ public override void RenderControl( HtmlTextWriter writer )
...
+                    {
+                        if ( service.Value.Value.IsAuthorized( Authorization.VIEW, currentPerson ) )
+                        {
```
azturner: Might want to consider creating\setting a local variable equal to `service.Value.Value` and use it instead since it is being used so often..
`var searchComponent = service.Value.Value;`

PR #1477, opened 2016-04-07
Added ability to add comment to a new Connection Request

bwitting:
via the Create Connection Request workflow action.

### DIFF
--- a/Rock/Workflow/Action/CreateConnectionRequest.cs
+++ b/Rock/Workflow/Action/CreateConnectionRequest.cs
@@ -143,6 +143,7 @@ namespace Rock.Workflow.Action
             connectionRequest.ConnectionStatusId = status.Id;
             connectionRequest.CampusId = campusId;
             connectionRequest.ConnectorPersonAliasId = opportunity.GetDefaultConnectorPersonAliasId( campusId.Value );
+            connectionRequest.Comments = comment;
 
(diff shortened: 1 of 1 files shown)


Rock/Workflow/Action/CreateConnectionRequest.cs
```diff
@@ -129,6 +131,9 @@ public override bool Execute( RockContext rockContext, WorkflowAction action, Ob
                 }
             }
 
+            // Get the Comment
+            String comment = action.GetWorklowAttributeValue(GetAttributeValue(action, "ConnectionCommentAttribute").AsGuid());
```
nairdo: @bwitting Is the plan to use this comment as the connectionRequest.Comments property?  If so I think you forgot that part, right?

bwitting: D'oh. Yes.  Added now, thanks!

JayGreentree:
@bwitting 
action.GetWorklowAttributeValue is missing the f .in workflow

nairdo:
Yeah, but that's not Bronson's fault. We actually, accidentally implemented that method misspelled a while back ([in May 2014](https://github.com/SparkDevNetwork/Rock/commit/39b9969e3200a84c1c02f2d96e905c5e7817264c)).

PR #1496, opened 2016-04-14
Improvements to RegistrationEntry block.

Tschrock:


azturner:
@Tschrock , We actually recently changed the confirmation so that it did not default to registrant since a very common registration was a parent registering their child. The other two changes are great, and in fact I went ahead and already made them in core so that we have that fix now. If you'd like to add a new block setting that if set to true will auto-fill the `Completed by` field, that would be great (default being false).

PR #1507, opened 2016-04-26
Fixed Side Navbar Transition Toggle

garrettjohnson:
Transition doesn’t stutter thanks to CSS relative fix on
content-wrapper, and chained jQuery.


arranf:
:+1:

PR #1508, opened 2016-04-26
+ Added ability to copy EventItemOccurrences.

tcavaletto:


nairdo:
...and this also normalizes most blocks with a "copy" function to change those buttons into a copy icon that is right-aligned (as per discussion on 4/22/2016).

PR #1516, opened 2016-05-02
+ Added ability to filter transactions by campus.

tcavaletto:


azturner:
@tcavaletto , @nairdo Curious what the use case is for this? Just afraid there might be some confusion as to what the the campus filter is doing. I see by looking at the code that it filters based on the campus of the batch that transaction belongs too, but some might think that it filters based on the campus associated to the account(s) that transaction contains.

nairdo:
@azturner The use case is: when the finance team is searching for a particular transaction issue they must examine each transaction within a date range but they only know which campus has the problem transaction.  To clarify the concern you raise, we could change the filter label to say "Batch Campus" or "Campus (of Batch)" if you think this would work better.

azturner:
@nairdo , I merged this, updated the label, and also modified it so that it's only available on the main transaction list block (i.e. not when used on batch, person, or scheduled txn detail blocks)

PR #1519, opened 2016-05-08
[WIP] Optimize attendance analytics

dcs619:
One of our larger campuses has been unable to pull attendance data for Easter weekend, so this will be the first round of optimizations. Feel free to merge this and I'll open new PR's for the others.

This PR optimizes the block code to defer filtering on the attendee visit data set (in our case, was 1k+ rows) instead of forcing the data set to recalculate after every filter.  

I also removed duplicate code that added demographic and visit information but was only calculated when `Xth Visit` was set to `No` (0).  Adding a [check for Xth Visit](https://github.com/NewSpring/Rock/blob/optimize-attendance-analytics/RockWeb/Blocks/CheckIn/AttendanceAnalytics.ascx.cs#L1081) prevents the `No` condition data set from getting filtered out.

### DIFF
--- a/RockWeb/Blocks/CheckIn/AttendanceAnalytics.ascx.cs
+++ b/RockWeb/Blocks/CheckIn/AttendanceAnalytics.ascx.cs
@@ -1,11 +1,11 @@
 ﻿// <copyright>
 // Copyright by the Spark Development Network
 //
-// Licensed under the Apache License, Version 2.0 (the "License");
+// Licensed under the Rock Community License (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-// http://www.apache.org/licenses/LICENSE-2.0
+// http://www.rockrms.com/license
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


dcs619:
Just noticed the header attributes were marked modified -- I must've pulled an old version of develop?  I'll revert that change.

azturner:
@dcs619 , Not sure I'm following the 'removed duplicate code' change. When Nth visit is equal to zero (showNonAttenders), the allAttendeeVisits dictionary is not loaded, the allResults dictionary is. So it doesn't look like that option would work with this change.

dcs619:
@azturner I'll take a look at that.  I think there's only one dictionary being used now, but I can't remember

azturner:
@dcs619 Any more info on this?

dcs619:
@azturner I reverted the change but left the filtering optimization. I was trying to avoid the duplicate logic in [this section](https://github.com/sparkdevnetwork/Rock/blob/develop/RockWeb/Blocks/CheckIn/AttendanceAnalytics.ascx.cs#L1079-1248):

    if ( attendee )
        - filter by data view
        - filter by visit
        - add first visit
        - add last attended
        - add demographics
        - allResults = filtered data 
    else 
        - filter by data view
        - add first visit
        - add last attended
        - allResults = filtered data

PR #1520, opened 2016-05-09
Fix issues with ProcessPayments downloading transactions 

dcs619:
Fixes two immediate issues:

- [x] Downloaded payments that didn't match a ScheduledTransactionDetail were being discarded
- [x] Updates the transaction status if it changed since that transaction was last downloaded

Item 2 seems to only affect the NMI gateway ( pending -> completed ) as the other gateways only pull completed transactions.  However, I'd assume other gateways never change a transaction status once it's completed.


dcs619:
@azturner already fixed, closing.

PR #1532, opened 2016-05-20
+ Added Address Field Type

tcavaletto:


arranf:
In the field type there's a 'todo' where I think there should be some exception handling

tcavaletto:
Thanks for catching that. That's actually vestigial code from an iteration where I thought error handling would be necessary, but it turns out it's unneeded. If the method doesn't get a location guid it's intended to return the empty guid string. The same thing happens in the PersonFieldType: https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Field/Types/PersonFieldType.cs#L202. But thanks for the catch, I'll clean up that bit of code.

arranf:
Vestigal, eh? ;) Good to know!

PR #1549, opened 2016-05-31
Bcrypt for Database Authentication

benwiley:
- Establish pattern for user chosen encryption method for database authentication
- Implement bcrypt as a second choice to the default HMACSHA1
- Allow "hot swapping" of encryption method (userlogin passwords are rehashed as users login if settings change)

We chose to implement bcrypt as a recommendation from a security expert, [Tim Tomes](http://www.lanmaster53.com/about/) (who [writes regularly](http://www.lanmaster53.com/categories/) about security risks and injection attacks), that recently provided security training to the NewSpring Web Team:

> Basically, HMAC is a signed hash, focused on integrity of information, not necessarily confidentiality. From a security standpoint, it's similar to fast hashing with a secret salt. In any case, HMAC was not designed for this kind of implementation. While it still provides a level of security vs. storing in plain text, it's not a good option. They should refactor this to use an Adaptive Hashing Algorithm like bcrypt. Regardless of the new system, if they change the way creds are stored, they will need to engineer a solution that implements a multi-stage conditional authentication system that supports the temporary use of both the old and new algorithms.

cc @jbaxleyiii @dcs619


cbump:
We reviewed this pull request here at SECC (@trydyingtolive and @cbump) and we are happy to see the algorithm being updated to use bcrypt.  We had reviewed the way the HMACSHA1 was implemented previously and the double-hashing with 2 different salts seemed nice/interesting but most documentation online is pretty straightforward that interesting != better.  We aren't convinced that the implementation of this change is ideal.  If we collectively agree that the HMACSHA1 implementation is more insecure than the bcrypt option, then we think we need to completely remove the HMACSHA1 option in future versions of Rock.

jbaxleyiii:
@cbump one of the things we were trying to include in this PR was the addition of choice in your encryption methods. Given the open nature of Rock, and putting choice to the users, we thought this was a better approach. Would you feel better if bcrypt became the new default?

jbaxleyiii:
Also, thanks for taking a look at in! 🎉

arranf:
@jbaxleyiii I think that opening up encryption methods opens up a can of worms. Encryption can and be done poorly and I imagine it would be if this was opened up.  Encryption can be objectively evaluated and consequently in my opinion it makes much more sense to have a secure peer-reviewed single option.

benwiley:
I saw the user chosen option as more of a migration pathway for churches that already have Rock in production.  We need a way to transform all of the existing hashes into bcrypt while providing a smooth user experience.

cbump:
@jbaxleyiii We think we DO want to provide some choices but strongly encourage against ones that are less secure than others.  Where these choices matter is when it impacts user experience.  For example, an authentication provider which enables 2-factor authentication vs single-factor database authentication is a choice that we think every church should make but we think we agree with @arranf that we all collectively should settle on a single option.  (Sorry for all the "we's" above.  This is coming from both @trydyingtolive and @cbump)

arranf:
@benwiley I'm not sure if that is necessarily the answer to your problem. A user chosen option allows me as a malicious individual to write a poor cipher, have it installed by churches, and then exploit my poorly designed encryption. Given how easy it is to obfuscate code and the depth of knowledge required to actually evaluate the effectiveness of a encryption algorithm this could potentially lead to a number of headaches down the line.

dcs619:
@arranf Hypothetically, what's to say my new MySpace OAuth plugin isn't poorly written and would be exploited the same way?  I don't think @azturner/@edmistj want a `No Authentication Plugins` rule for the Rock store, but if you follow that line of reasoning, you'd have to evaluate it.

I'd vote for BCrypt encryption by default in core.  However, this isn't a democracy, so ¯\_(ツ)_/¯

arranf:
@dcs619 You're completely right about poorly written plugins. OAuth is relatively easy to get right due to OAuth's nature. Encryption requires a much greater level of expertise and there's a reason most encryption is handled by a number of well tested and review libraries for each language, such as bcrypt. For the record I'm in favour of using bcrypt.

edmistj:
We've added this to our Core Developer topics that we review every Friday.

edmistj:
While we believe the current password encryption is good in Rock we do want to make it even better by incorporating this change. We'll probably take much of the code from the PR and incorporate it manually (vs merging the request) as we'd like to make the usage of bcrypt required (not option or setting). We'll work to make the conversion over to bcrypt transparent (like the PR does).

There's probably not enough time to get this into v5 and make the alpha deadline of 6/7. So we'll target it as a fix in v5.1 or v5.2.

Changes like these show the power of community!

azturner:
We updated core to use BCrypt. Took a lot of what you implemented with this pull request, but took away option for selecting encryption method. Going forward all database passwords will be hashed using BCrypt, and existing SHA1 hashed passwords will by converted to BCrypt first time anyone logs in again.  https://github.com/SparkDevNetwork/Rock/commit/20127a8ee4f36dcf7e0ca6a1879b13aa66f38553

dcs619:
@benwiley++

dcs619:
@azturner With this transaction, could we optimize the [type checking in Authenticate()](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Security/Authentication/Database.cs#L102)?  According to the [spec](https://en.wikipedia.org/wiki/Bcrypt), the Bcrypt hash will always start with a `"$2a$"`, `"$2b$"`, or `"$2y$"` prefix.  Given an increasing cost factor over time, using a try/catch adds its own overhead to the hash/authentication time.

The original PR included a [regex match](https://github.com/NewSpring/Rock/blob/bcrypt/Rock/Security/Authentication/Database.cs#L424) but only accommodated `$2a$`.  The regex `^\$2[aby]\$\d{2}\$[\/\.a-zA-Z0-9]{53}$` will match any of the valid options.  @benwiley or I can submit a PR if you want.

azturner:
@dcs619 I was actually thinking that over time the cost factor would be reduced by not doing the regex match and just trying bcrypt first. Each existing person should only get the handled exception once on the first time they log in. After that, their password is converted to BCrypt, and they wouldn't get the exception and in those cases, doing the regex match just adds overhead. (I was also a little worried we might not get the regex right for all passwords).

PR #1572, opened 2016-06-13
Centering the profile image on the login status block.

cbump:
We have rectangular photos that were imported into Rock during migration.  This helps keep them centered in the login status bubble.  It shouldn't have any impact on any existing Rock implementations since the photo uploading tools use a square crop.


edmistj:
Hey Chuck! We have quite a few other churches that have imported non-square images. We typically find that the current setting of:

background-size: cover;
background-repeat: no-repeat;

Every picture is different (as the face is not always in the same place) but this has worked well to date. Do you have some samples of when this works better?

austinspooner:
Jon, we also see this with our staff images on Rock that came over. Here is a screen shot.

image

The person's ID in our system is 12215 if you want to take a look.

-Austin

cbump:
@edmistj Technically you are correct but in practice our photographs do feature people (mostly) centered which is why we are seeing an improvement with the background being centered horizontally.  The example that @austinspooner posted is the same kind of stuff we are seeing without the fix.

edmistj:
@austinspooner can you provide the source image.

austinspooner:
Attached. 
terrih

edmistj:
Sorry for the delay in getting back to this. I did do quite a bit of testing when we came up with the current solution, but if you can run some tests and post images of the results that would help. Be sure to test photos that are very long and very tall.

cbump:
The background:cover works fine on any portrait style photos but anytime we go to a landscape photo where the subject is centered on the x-axis the subject will be pulled over to the right of the circle.  I've attached the original photo and a screenshot of what we are seeing.  My original pull request fixes this in the login status block but we need to fix the css in the theme.css as well.

photo-round
bumpcharles

edmistj:
It would be best to get a sampling of all the tests. Something like before and after of all these:
samples

arranf:
@cbump To get this PR moving I think you'll either need to go and get photo examples of why this PR is better (like Jon outlined above). If you're happy with the current solution, I think we should close this. It's been open a while!

dcs619:
I believe the photo centering under discussion was an issue for us too, but @edolyne or @richarddubay would have to provide more granular details.

cbump:
I've FINALLY revisited this and I've update 3 less files because I still think it slightly improves things.  Here are the photos which show the difference:
Before:
old

After:
new

This targets a subject centered in a landscape photo.  It more-or-less leaves every other situation in the same shape as it was previously.

edmistj:
I manually made these changes as it was in conflict with another PR with more changes.

PR #1592, opened 2016-06-27
- Adding slice to lava string filters.

cbump:
This implements the slice filter from standard liquid: https://help.shopify.com/themes/liquid/filters/string-filters#slice


nairdo:
note: I changed the lowercase s to uppercase to match the standard Lava tag pattern.

PR #1615, opened 2016-07-08
Add MergeTemplatePageRoute override to Batch List Block

treyhendon:
Added block setting (and logic to see if populated) to provide an override to the Merge Template Page Route so that alternate Merge Template Entry block can be used if desired.

### DIFF
--- a/RockWeb/Blocks/Finance/BatchList.ascx.cs
+++ b/RockWeb/Blocks/Finance/BatchList.ascx.cs
@@ -37,6 +37,7 @@ namespace RockWeb.Blocks.Finance
     [Description( "Lists all financial batches and provides filtering by campus, status, etc." )]
     [LinkedPage( "Detail Page", order: 0 )]
     [BooleanField( "Show Accounting Code", "Should the accounting code column be displayed.", false, "", 1 )]
+    [TextField( "Merge Template Page Route", "Route to merge template block. Sample: ~/FinanceMergeTemplate/{0}", false, "", "", 2 )]
     public partial class BatchList : Rock.Web.UI.RockBlock, IPostBackEventHandler
     {
         #region Fields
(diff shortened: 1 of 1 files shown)


azturner:
Closing this as it breaks pattern of merge button consistently taking user to same place, and is a block setting that most people would not need (adds complexity). Am curious though of what the use-case is for an alternate merge template block.

treyhendon:
I created a custom merge template block that exposes all Accounts and another Finance specific area (populated via SQL based on the merge Id) to Lava end points for a Batch Summary Report that several churches have grown accustomed to using in other ChMS.

Unfortunate that I'll have to keep a custom version of the Batch List just to have that one setting. Correct that use case is rare, but also seems that the use of an alternative custom merge page should be accessable somehow. Remembering that the grid had a property for Merge Route, what about maybe a page setting that blocks could access (like when shifting context for a page via url)?

azturner:
@treyhendon Thanks for the additional context. How about this... what if you just added your block to the existing merge template page (above existing block) and it interprets the entitytype of the entityset id being passed in (the EntityTypeId field on the EntitySet table) and only displays itself when a FinancialBatch entityset was used. It could also hide the core block by doing something like this in your block...
```
foreach ( RockWeb.Blocks.Core.MergeTemplateEntry block in RockPage.RockBlocks.Where( a => a is RockWeb.Blocks.Core.MergeTemplateEntry ) )
{
    block.Visible = false;
}
```
That also requires that you add a reference in your markup...
```
<%@ Reference Control="~/Blocks/Core/MergeTemplateEntry.ascx" %>
```
-David

treyhendon:
That's a great idea!  Now that we've got this first wave of the Batch report out, I'll work on trying to fold the merge block back like that.  I love not having one more thing to try to config!  Would make putting this block in the shop much easier. Thanks!

PR #1622, opened 2016-07-13
Closed HTML elements that were preventing DOM navigation

saconger:
Closed HTML elements that were preventing DOM navigation via XmlDocument().


edmistj:
Thanks, great catch! Do you think this should go into v5.2 or is v6.0 ok?

saconger:
Hi Jon, timing of this does not matter to me. It's just something I found
while working on the translator.  I'll probably find more like it, as I go
screen to screen. So whatever version you want to put it in, is fine. Scott

PR #1624, opened 2016-07-13
Fixed nav from showing text selection cursor

garrettjohnson:
Fixed so side nav no longer shows the text selection cursor on on non-clickable headers (and the font awesome icons in IE/Edge)


edmistj:
Thank you!

PR #1647, opened 2016-07-27
Adding the whole transaction entity as a merge field option for receipts.

cbump:
Added a new merge field for the whole Transaction entity in payment receipts.  This allows us to include information such as the transaction date/time, the payment method etc.


dcs619:
👍 Could we fix the spelling of `reciepts`?  I've noticed that misspelled in several places across Rock.

Issue #1653, opened 2016-07-28
Confirm Account for change password does NOT validate password complexity


cbump:
Steps to reproduce:
- Go to the login page on the external site and click "Forgot Password"
- Enter your email address and submit
- Once you receive the email, click the link to change your password
- Enter a password with any complexity (e.g. "1") in both the password and confirmation fields and it will allow you to change your password.

PR #1654, opened 2016-07-28
Issue #1653 - Confirm Account password complexity

cbump:
- Adding the standard Rock password complexity checking to the "Confirm Account" block.


arranf:
Great catch!

PR #1674, opened 2016-08-03
Fix V6 syntax in V5

arranf:


edmistj:
Good catch! This needs to have the whole 'is this attribute and attribute' logic put back with the fix applied. Can you add that?

edmistj:
Nevermind... this is right.

PR #1679, opened 2016-08-05
Changed grid filter to advanced mode.

jreciomoreno:
Switching to advanced filter mode fixes an issue where some registration attributes do not show a valid filter attached to it.

### DIFF
--- a/RockWeb/Blocks/Event/RegistrationInstanceDetail.ascx.cs
+++ b/RockWeb/Blocks/Event/RegistrationInstanceDetail.ascx.cs
@@ -867,8 +867,8 @@ namespace RockWeb.Blocks.Event
                         {
                             try
                             {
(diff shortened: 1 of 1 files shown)


azturner:
@jreciomoreno , Can you share screenshots of the field types in question and what they look like before and after your change. Thanks.

jreciomoreno:
For the following single form inside a registration template:
image

Initially the grid filters ofthe Registration Attribute fields were not showing:
image

Changing the filter to advanced mode does make them work properly on such fields:
image

azturner:
Thanks for the screenshots. Couple thoughts... looks like the email field type is not consistent with other field types when using "simple" mode so we'll fix that. As far as using advanced instead of simple, we really think that most users would prefer the simple view. So rather than switching it for everyone, if you would like to add a block setting for this block (defaulted to simple mode) and resubmit a new pull request with that setting, we can pull it in.

PR #1681, opened 2016-08-05
Changed Device Detail block label.

jreciomoreno:
If a printer can be setup using it's IP Address and DNS hostname, changing the wording on the Label and Help textbox attributes will make these options clear to the user.


azturner:
@jreciomoreno, closing this since we don't really officially support using hostname in this field (and we haven't tested it on all check-in devices). We may also want to add functionality later that depends on value being an ip address.

PR #1682, opened 2016-08-05
Print check-in tags using printer's DNS hostname

jreciomoreno:


azturner:
@jreciomoreno , Sorrry, didn't see this one when reviewing/commenting on your other pull request (https://github.com/SparkDevNetwork/Rock/pull/1681), but since we do printing differently within the block vs. IOS app vs. Windows app we can't really update one without updating all of them. And we may want to actually require/depend on that value being an IP address.

PR #1686, opened 2016-08-08
+ Adds a block to allow editing of user's account and family from the home page

tcavaletto:
This adds a block that can be placed on the front-end website, allowing a user to edit not only their information, but that of their family(s), including adding new family members. It also includes a button to send the user to a page where they could request additional changes. This block did require some changes to the AttributeFieldType and AttributeFieldAttribute classes to add in qualifier column/value functionality. This was because the 'Family Attributes' block setting was displaying attributes from check-in groups, and we wanted it to only display Family GroupType attributes.
image
image

image

### DIFF
--- a/Rock/Attribute/AttributeFieldAttribute.cs
+++ b/Rock/Attribute/AttributeFieldAttribute.cs
@@ -17,7 +17,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


edmistj:
Thanks @tcavaletto for completing that. Looks just like what we discussed!

Issue #1601, opened 2016-06-28
[Obsolete] Remove Populate Three Families script


dcs619:
The [Dev Tools\Populate Three Families](https://github.com/SparkDevNetwork/Rock/blob/develop/Dev%20Tools/Sql/Populate_Three_Families.sql) script is quite a bit out of date and should be removed to avoid confusion.  The PopulatePlethoraRandom script or SampleData block can be used instead.

arranf:
So is the script that creates attendance groups and quite a few others :)

mikepetersonccv:
Would you say we have a plethora of obsolete scripts?

arranf:
Fixed in pull request #1696

PR #1696, opened 2016-08-11
- An opinionated cleanup of the SQL Dev Tools

arranf:
# Context
An issue reported broken dev tools SQL scripts #1601 

# Goal
Remove them from the repository removing false hope, reducing repo clutter, an reducing wasted time on broken scripts.

# Strategy
Anything that was broken that took me longer than 30 seconds to fix I just deleted on the basis it likely hadn't been used and wasn't being maintained.

# Possible Implications
Aside from someone wanting to recreate a script and missing a starting point there aren't any implications. Anyone wanting to do that can check historical branches.

# Screenshots
N/A


azturner:
Can you move these to an "Archive" folder instead. That way if we need similar functionality in the future we can revive them easier.

PR #1706, opened 2016-08-15
- Add TransactionLinkCache migration

### LINKED ISSUE #1693: OutputCache
The 'Transaction Links' block on Contribution tab needs to have its Cache Duration set to 0 instead of 3600, otherwise you'll start pulling up the wrong person info.  

This block is the one that has buttons for 'Add One-time gift' and 'New Scheduled Transaction'.

arranf:
# Context
Fixes #1693

# Goal
Change the cache on the TransactionLink HTML Content block on the Profile Page Contributions tab to 0 if it was previously set to 3600

# Strategy
A migration that targets the first HTMLContent block type with the name of Transaction Link as added by default

# Possible Implications
This will only fix this single instance


azturner:
@arranf Added this as a migration rollup: https://app.asana.com/0/22067109204476/167849025887929

PR #1709, opened 2016-08-16
+ Add Activate Activity in Other Workflow workflow action

arranf:
# Context
There's a new action in v6 that lets you activate a workflow using a workflow action. If based on logic you want to complete ~~kill~~ that workflow processing there's no way to do so. 

# Goal
To provide the ability for a workflow to activate an activity in another workflow.

# Strategy
* Edit the workflow fieldtype to provide a drop down list of possible workflow types to select an activity type
* Add a workflow action that will take an activity type and a workflow id and then activate the activity in the given workflow.

# Possible Implications
The WorkflowActivityTypeFieldType was not designed to be used outside of the context of a workflow and so uses a HTTP.context method, there could be issues there (although these haven't been observed).

# Screenshots
image

image

### DIFF
--- a/Rock/Field/Types/WorkflowActivityTypeFieldType.cs
+++ b/Rock/Field/Types/WorkflowActivityTypeFieldType.cs
@@ -37,6 +37,79 @@ namespace Rock.Field.Types
 
         private const string WORKFLOW_TYPE_KEY = "WorkflowType";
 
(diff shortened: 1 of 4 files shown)


arranf:
@azturner I've updated this commit to add the second action which matches attributes

I've also spent a while changing the fieldtype so it now can defaults to a 'Current Workflow' option.

PR #1740, opened 2016-09-02
Use the new Google Maps url structure & HTTPS

garrettjohnson:
# Context
Google maps doesn't use HTTPS or the new map link standard (`https://www.google.com/maps/place/`)

# Goal
Add HTTPS link style and remove unused label parameter. 

# Strategy
Remove commas and encoded linebreaks which google doesn't like

# Possible Implications
The new URL structure seems to be quite a bit more sensitive to poorly encoded URLs so there might be some outliers.

### DIFF
--- a/Rock/Model/Location.cs
+++ b/Rock/Model/Location.cs
@@ -613,7 +613,7 @@ namespace Rock.Model
         /// </summary>
         /// <param name="title">A unused <see cref="System.String"/> containing the location name parameters.</param>
         /// <returns>A <see cref="System.String"/> containing the link to Google Maps for this location.</returns>
-        public virtual string GoogleMapLink( string title )
+        public virtual string GoogleMapLink( string title = "" )
         {
             string qParm = this.GetFullStreetAddress();
             // Remove linebreaks and commas
--- a/Rock/Model/Location.cs
+++ b/Rock/Model/Location.cs
@@ -611,17 +611,15 @@ namespace Rock.Model
         /// <summary>
         /// Returns a Google Maps link to use for this Location
         /// </summary>
-        /// <param name="title">A <see cref="System.String"/> containing the parameters needed by Google Maps to display this location.</param>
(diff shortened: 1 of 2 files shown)


garrettjohnson:
Definitely missed something in the merge ... 😦

edmistj:
Is this ready to be merged?

garrettjohnson:
@edmistj I think so; I'm not quite at home with C# yet

edmistj:
Do you have more information (link) to the map link standard (https://www.google.com/maps/place/)? Wanted to read up on that but can't seem to find it.

arranf:
What I can find is similar to that form but its versioned.
https://developers.google.com/maps/documentation/embed/guide#forming_the_url

garrettjohnson:
The link structure was pieced together from StackOverflow questions, and the links google generates. It doesn't appear that any Google Maps link API is officially documented.

edmistj:
We reviewed this in the Friday PR review and have a few thoughts:
+ Adding HTTPS is a great addition and we should implement that
+ We're a little nervous on the URL to use. Like you we want to use the best one, but the documentation is unclear on which that might be. I haven't found any that references https://www.google.com/maps/place/. Can you double-check that that is correct?
+ Finally on the GroupMember block. Is there a reason to remove the Person.FullName?

Thanks for all the effort to polish this for Rock!!!

garrettjohnson:
The more I've thought on this one, the more it seems like this version of
the PR is a bad idea. Google hasn't documented this link structure ever.
That and most open source software uses the the maps.google.com link
structure, and is less likely to break.

In testing I was able to get https to work on the current link structure (
maps.google.com). This is a much better solution, I'll resubmit the pull. I
removed the link label because Google has removed support for using the
location label on maps.

--
Garrett Johnson
559.272.6863

edmistj:
Thanks for all your effort on this!!!

PR #1743, opened 2016-09-06
Add workflow action with large code editor for running lava

Tschrock:
# Context
With the use of lava becoming more and more prevalent, the existing method of running lava in a workflow - the Attribute Set Value action - is getting harder to use (It only has a small text box).

# Goal
To have a large editor for running lava in a workflow.

# Strategy
This PR adds a new workflow action dedicated to running lava and provides a large code editor for editing.

# Possible Implications
None that I know of.

# Screenshots
capture31


edmistj:
This PR had a merge conflict (probably due to previous pulled PR) so I'll incorporate it in manually. Nice addition though!

edmistj:
Added it to the the v6 branch

PR #1744, opened 2016-09-06
Add workflow action to run powershell

Tschrock:
# Context
Our IT guy showed some interest in being able to run powershell from a workflow. This would allow him to use, for example, remote O365 to update or add people. This would also allow for a lot of custom scripting and add a nice tool to workflows.

# Goal
To provide a way to run powershell from a workflow and get some usable output back.

# Strategy
This PR adds a new workflow action that allows powershell to be run and the results to be formatted with lava and stored in a workflow attribute.

# Possible Implications
Powershell has a lot of power - it may be wise to restrict editing of this action to admins. Also, some hosts may have restrictions in place on powershell access.

# Screenshots
capture32


edmistj:
This is pretty cool, but it might make a better plugin. Do you mind adding it to the store? I'm sure @arranf  would help you.

PR #1745, opened 2016-09-06
Add workflow action to start a Job

Tschrock:
# Context
There are some jobs (Such as calculating group requirements or group sync) that someone may want to be able to trigger automatically via a workflow. This way they can run a job on demand without needing to manually run it themselves or waiting for the scheduled time.

# Goal
Allow triggering jobs to run from inside a workflow.

# Strategy
This PR adds a new workflow action that will trigger a job to run.

# Possible Implications
I didn't explicitly check if a job is already running, but I'm pretty sure the scheduler handles that?

# Screenshots
capture33


edmistj:
This is a good idea. I'll take a look at the code later this week.

PR #1746, opened 2016-09-06
Add option to toggle tracking on mailgun emails

Tschrock:
# Context
Someone on slack brought up wanting to disable tracking links in Mailgun emails. Right now the email headers are hard-coded to enable tracking. 

# Goal
Add an option to disable tracking on Mailgun emails.

# Strategy
This PR adds a "Track Clicks" option to the MailgunSmtp config which toggles the respective email headers.

# Possible Implications
None: the option is true by default, maintaining the previous behavior.

# Screenshots
capture34


edmistj:
Will look at this one too. With @arranf  new http://www.rockrms.com/CommunityProjects page it would be good to submit the concepts first in the future (now that we have this tool).

PR #1750, opened 2016-09-07
Fix ActivateWorkflow action not copying attributes properly

Tschrock:
# Context
The ActivateWorkflow activity doesn't copy attributes to the new workflow properly - it's only copying the formatted value.

# Goal
Fix how the attributes are copied.

# Strategy
This fixes the issue by changing it so that the raw value is copied instead of the formatted value.


edmistj:
Good catch!

PR #1751, opened 2016-09-07
Fix Remove Person from Group Using Attribute Workflow Action removing all groups

Tschrock:
# Context
The `Group Member Remove Person from Group Using Attribute` Workflow action removes the person from all of their groups instead of just the one specified.

# Goal
Fix so that the person is only removed from the specified group.

# Strategy
Add a filter to the list of group members to delete so it only removes ones in the specified group


edmistj:
Face palm... my bad.

PR #1752, opened 2016-09-07
Feature cb pluginstartup

cbump:
# Context
I created an OAuth Server plugin that uses this to initialize itself.  We use an OAuth server both for 3rd party integrations with our implementation of Rock as well as for mobile app authentication.  This allows for refresh token authentication which is nice for re-authenticating a user without having to have them enter their credentials every time.

# Goal
The goal is to allow any plugin to have code executed on application startup.  By using OWIN, we can also register callbacks with OWIN's cancellation token which allows you to run code at shutdown as well.

# Strategy
This pull request adds the ability for any plugin to extend Rock.Plugin.IStartup and then implement the following method:
`        public void Configuration(IAppBuilder app)
        {
        }
`
# Possible Implications
This adds a lot of power for plugins since they can actually execute things on startup.  A plugin could potentially completely break someone's instance of Rock.  I don't really think that's a whole lot different than a lot of the architecture as things stand today.  Rock's plugin architecture is very powerful and full-featured and this just adds to that.


cbump:
Let me know if the Appveyor failure is my fault.  I just double-checked a clean checkout of this branch and it seems to build just fine.

edmistj:
We've added the ability to process startup items here:
https://github.com/SparkDevNetwork/Rock/commit/a99845195e995c532fa534aacb29705fef958c33

We'll add shutdown processing in a similar way.

PR #1529, opened 2016-05-18
#1528 - Plugin Migration now extends DbMigration.

cbump:
- Adding a RunMigration method which takes the operations from DbMigration and applies them using the standard Sql method.
- Calling RunMigration from Global.asax.cs when a plugin migration is run.

Issue #1528, opened 2016-05-18
Plugin Migration does not extend DbMigration


cbump:
The Plugin Migration class should extend DbMigration so a lot of the standard functions are available such as CreateTable, AddForeignKey etc.

dcs619:
@cbump have you tested this in multiple environments?  I'm fairly certain Entity Framework requires DbMigrations to be in order, which means you couldn't install a plugin at a later date (once other DbMigrations have been applied).

cbump:
@dcs619 It doesn't change anything about the way the order of the plugins are installed.  It just allows the SQL that is applied by the migration to be generated from the base DbMigration functionality.  The plugin migration is actually still applied and tracked the same way.

PR #1753, opened 2016-09-08
#1528 - Implement standard DbMigration methods in Plugin Migration Take 2

cbump:
# Context
This adds the ability to use standard DbMigration methods in a plugin migration.

# Goal
The goal is to make plugin migration development "feel" more like standard entity framework.

# Strategy
I wrapped a handful of the standard migration methods into plugin migration. This allows for Rock to include the DbMigration methods for things like CreateTable, DropIndex etc. to expose them to Plugin Migrations.  More information about a previous PR which would have added an EF dependency to plugins is available here: https://github.com/SparkDevNetwork/Rock/pull/1529

# Possible Implications
This should not break any backwards compatibility.  It only adds new functionality.

# Screenshots
There are no UI changes.

### DIFF
--- a/Rock/Plugin/Migration.cs
+++ b/Rock/Plugin/Migration.cs
@@ -15,8 +15,14 @@
 // </copyright>
 //
 using System;
+using System.Collections.Generic;
 using System.Data;
+using System.Data.Entity.Migrations.Builders;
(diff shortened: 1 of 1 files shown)


arranf:
:+1:

PR #1756, opened 2016-09-09
Eliminate extra spaces in SQL Query block

ccofnow:
# Context
Extra spaces at the end of the SQL command default contents which show up as indent marks to the end user and make the more picky among us feel the need to delete them each time.

# Goal
This pull request uses comments to keep the coding clean and in-line vertically, while not putting the side-effects on the end user

# Strategy
Strategic use of comments

# Possible Implications
The only impact should be the cleanliness of the code, which is important to the project and perhaps there is a different preferred way to deal with this.

# Screenshots
sql query indents

The line break and indents make sense in the ascx file but left unpleasant spaces in the default block contents. I recommend this strategic use of comments to maintain the cleanliness of the code while not leaving any side-effects for the end user.

### DIFF
--- a/RockWeb/Blocks/Reporting/SqlCommand.ascx
+++ b/RockWeb/Blocks/Reporting/SqlCommand.ascx
@@ -17,8 +17,7 @@
 SELECT
     TOP 10 *
 FROM
-    [Person]
-                        </Rock:CodeEditor>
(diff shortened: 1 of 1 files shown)


nairdo:
@ccofnow as I just chatted with you... we ended up moving it to the code-behind and cleaned up the extra line breaks in the ascx.

PR #1763, opened 2016-09-14
Update Google Map Link to HTTPS

garrettjohnson:
# Context
A simpler version of PR #1740, that maintains the current, and ubiquitous `maps.google.com/maps?q=` link structure. Updates to include HTTPS, confirmed working with US and international addresses.

# Goal
Use HTTPS without using Google's new, and undocumented link structure.

# Strategy
Added an "S" to the current links. This PR also removes the location label from the URL, but makes the title string optional to maintain backwards compatibility.

# Possible Implications
Minimal comparative to PR #1740, Google could depreciate this link format, but it's unlikely considering the number of sites and projects using this structure.

### DIFF
--- a/Rock/Model/Location.cs
+++ b/Rock/Model/Location.cs
@@ -611,17 +611,13 @@ namespace Rock.Model
         /// <summary>
         /// Returns a Google Maps link to use for this Location
         /// </summary>
-        /// <param name="title">A <see cref="System.String"/> containing the parameters needed by Google Maps to display this location.</param>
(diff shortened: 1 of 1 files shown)


edmistj:
Just curious on the thinking of removing the title from the link?

garrettjohnson:
@edmistj Browser history; if we have staff on a less private workstation removing the name increases security just a little. Plus Google doesn't use the label value anymore 😄

garrettjohnson:
👍 Updated

PR #1783, opened 2016-09-27
- Added GetFormattedDataValue to CallbackField to make it easier to g…

mikepetersonccv:
# Context
We had a Report that included a person attribute called "Working Campus" which was a CampusField.  When exporting to excel, it would export the Guid instead of the Campus.Name

# Goal
The ReportingHelper creates a CallbackField for Attributes, so any exports from a Report or DynamicReport will have the raw value for Attributes instead of the value shown on the Grid.  This solution fixes the problem by detecting a CallbackField in Grid.cs and getting the FormattedValue of the field instead of the raw value.

# Strategy
a change the Grid.cs and CallbackField.cs

# Possible Implications
If people were counting on the Raw Value of an attribute when exporting from a Reporting grid, this will change the behavior. So that is a concern.

# Screenshots

…et the resulting value of a CallbackField

- Updated Grid Excel Export so that CallbackField exports are the formatted value instead raw value.  This fixes an issue where Attributes that uses field types such as "CampusField" will return export the name of the campus instead of the Guid in Reports and DynamicReports.

### DIFF
--- a/Rock/Web/UI/Controls/Grid/CallbackField.cs
+++ b/Rock/Web/UI/Controls/Grid/CallbackField.cs
@@ -51,6 +51,16 @@ namespace Rock.Web.UI.Controls
         }
 
         /// <summary>
+        /// Gets the formatted data value.
+        /// </summary>
+        /// <param name="dataValue">The data value.</param>
+        /// <returns></returns>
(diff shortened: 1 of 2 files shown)


arranf:
I love this. I have had this exact same problem!

Issue #1478, opened 2016-04-08
Mailgun bounced event not working due to doubling of communication_recipient_guid


nairdo:
When the bounced event is processed by the Mailgun.ashx, the `request.Form["communication_recipient_guid"]` contains the following string:

```
f5badd45-7bbb-4ff9-b1b0-f8c86e66bc77,f5badd45-7bbb-4ff9-b1b0-f8c86e66bc77
```
(normally it does not contain the guid twice -- I'm only seeing this in the bounced event)

Currently the handler is expecting to be able to do this, however it will not work with the above situation and the bounced event is basically not processed (and the person's email is not marked inactive):

```csharp
Guid? communicationRecipientGuid = request.Form["communication_recipient_guid"].AsGuidOrNull();
```

Here is a sample of some extra logging I did with our live environment:
```
4/7/2016 3:12:50 PM,Webhhook Called
4/7/2016 3:12:50 PM,dropped event for communication_recipient_guid 'f5badd45-7bbb-4ff9-b1b0-f8c86e66bc77'
4/7/2016 3:12:50 PM,Webhhook Called
4/7/2016 3:12:50 PM,bounced event for communication_recipient_guid 'f5badd45-7bbb-4ff9-b1b0-f8c86e66bc77,f5badd45-7bbb-4ff9-b1b0-f8c86e66bc77'
```

nairdo:
I've opened a [ticket with Mailgun (#269764)](https://mailgun.com/app/support/269764) to see how this is happening but I'm thinking about coding a workaround by splitting on the comma until this is resolved/understood.

PR #1785, opened 2016-09-28
Fixes Mailgun webhook bounced email event.

jreciomoreno:
Fixes an issue that prevented a bounced email from triggering the custom action on a SendEmailWithEvents activity by splitting the workflow action guid on comma if it exists.
Bounced emails should are now also marked as inactive.

### DIFF
--- a/RockWeb/Webhooks/Mailgun.ashx
+++ b/RockWeb/Webhooks/Mailgun.ashx
@@ -78,19 +78,26 @@ public class Mailgun : IHttpHandler
             
             if ( !string.IsNullOrWhiteSpace( request.Form["workflow_action_guid"]))
             {
-                Guid? actionGuid = request.Form["workflow_action_guid"].AsGuidOrNull();
+                Guid? actionGuid = request.Form["workflow_action_guid"].Split( ',' )[0].AsGuidOrNull();
                 string status = string.Empty;
                 switch ( eventType )
                 {
-                    case "complained":
-                    case "unsubscribed":
+                    case "complained": break;
+                    case "unsubscribed": break;
                     case "delivered": status = SendEmailWithEvents.SENT_STATUS; break;
                     case "clicked": status = SendEmailWithEvents.CLICKED_STATUS; break;
                     case "opened": status = SendEmailWithEvents.OPENED_STATUS; break;
-                    case "dropped": 
-                    case "bounced": 
-                        status = SendEmailWithEvents.FAILED_STATUS; break;
-                }                
+                    case "dropped": break;
+                    case "bounced": status = SendEmailWithEvents.FAILED_STATUS;
+                        int secs = request.Form["timestamp"].AsInteger();
+                        DateTime ts = new DateTime( 1970, 1, 1, 0, 0, 0, DateTimeKind.Utc ).AddSeconds( secs ).ToLocalTime();
+                             Rock.Communication.Email.ProcessBounce(
+                                    request.Form["recipient"],
+                                    Rock.Communication.BounceType.HardBounce,
+                                    request.Form["notification"],
+                                    ts );
+                        break;
+                }
                 if ( actionGuid != null && !string.IsNullOrWhiteSpace( status ) )
                 {
                     SendEmailWithEvents.UpdateEmailStatus( actionGuid.Value, status, eventType, rockContext, true );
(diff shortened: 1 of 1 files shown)


RockWeb/Webhooks/Mailgun.ashx
```diff
@@ -78,19 +78,26 @@ public class Mailgun : IHttpHandler
...
                     case "clicked": status = SendEmailWithEvents.CLICKED_STATUS; break;
                     case "opened": status = SendEmailWithEvents.OPENED_STATUS; break;
-                    case "dropped": 
-                    case "bounced": 
-                        status = SendEmailWithEvents.FAILED_STATUS; break;
-                }                
+                    case "dropped": break;
+                    case "bounced": status = SendEmailWithEvents.FAILED_STATUS;
+                        int secs = request.Form["timestamp"].AsInteger();
+                        DateTime ts = new DateTime( 1970, 1, 1, 0, 0, 0, DateTimeKind.Utc ).AddSeconds( secs ).ToLocalTime();
```
arranf: Rather than ToLocalTime() (which will report the time the event occurred as UTC regardless of the Rock installation's time zone) you will want to use RockDateTime.OrgTimeZoneInfo to identify the offset that should be used.

jreciomoreno: Thank you, will do that.

RockWeb/Webhooks/Mailgun.ashx
```diff
@@ -78,19 +78,26 @@ public class Mailgun : IHttpHandler
             
             if ( !string.IsNullOrWhiteSpace( request.Form["workflow_action_guid"]))
             {
-                Guid? actionGuid = request.Form["workflow_action_guid"].AsGuidOrNull();
+                Guid? actionGuid = request.Form["workflow_action_guid"].Split( ',' )[0].AsGuidOrNull();
                 string status = string.Empty;
```
azturner: @jreciomoreno , Not sure why there would be more than one guid here, this is set by the SendEmailWithEvents workflow action, and it looks like it would only ever have one guid.

RockWeb/Webhooks/Mailgun.ashx
```diff
@@ -78,19 +78,26 @@ public class Mailgun : IHttpHandler
...
                 switch ( eventType )
                 {
-                    case "complained":
-                    case "unsubscribed":
+                    case "complained": break;
+                    case "unsubscribed": break;
                     case "delivered": status = SendEmailWithEvents.SENT_STATUS; break;
```
azturner: Adding the breaks, would cause the status not to get set to 'SENT' when a complained or unsubscribed event was relieved. Without the 'break' these fall through to the "delivered" case.

RockWeb/Webhooks/Mailgun.ashx
```diff
@@ -78,19 +78,26 @@ public class Mailgun : IHttpHandler
...
-                    case "dropped": 
-                    case "bounced": 
-                        status = SendEmailWithEvents.FAILED_STATUS; break;
-                }                
+                    case "dropped": break;
+                    case "bounced": status = SendEmailWithEvents.FAILED_STATUS;
+                        int secs = request.Form["timestamp"].AsInteger();
+                        DateTime ts = RockDateTime.ConvertLocalDateTimeToRockDateTime( new DateTime(1970,1,1,0,0,0,DateTimeKind.Utc).AddSeconds(secs).ToLocalTime());
+                             Rock.Communication.Email.ProcessBounce(
+                                    request.Form["recipient"],
```
azturner: Not sure why the ProcessBounce was added here since it's done further on. This case statement is just to handle cases when email was from workflow action.

RockWeb/Webhooks/Mailgun.ashx
jreciomoreno: @azturner , On testing I noticed that Mailgun was returning two identical guids for some reason, I added this so the code would not fail. This seems to be an issue related to @nairdo  fix on 4/21/2016 for FIx #1478

RockWeb/Webhooks/Mailgun.ashx
jreciomoreno: My mistake, those breaks for "complained", "unsubscribed" and "dropped" should not be there as the delivered case should catch the first two and bounced should catch dropped.

RockWeb/Webhooks/Mailgun.ashx
jreciomoreno: I added ProcessBounce in there since the other one only process the bounces for communications; I noticed the response coming from Mailgun had either a workflow_action_guid or communication_recipient_guid request form. We ideally want to process bounces for both scenarios.

Issue #1791, opened 2016-10-05
Excel export and Communcation links from DynamicData block are not filtered


cbump:
### Prerequisites
- [Y ] Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
- [Y ] Did you include your Rock version number and client culture setting?
- [N ] Have you attached your [environment and diagnostics information](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) to this issue?
- [Y ] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?
### Description

When using a dynamic data block, the excel export and the communication link (if it's a person report) do not honor the filter.
### Steps to Reproduce
1. Setup a dynamic data block with filtering.
2. Filter the block
3. Export to Excel

**Expected behavior:**
The exported file should match the filter.

**Actual behavior:**
Everything is exported!
### Versions
- **Rock Version:** Rock McKinley 5.3
- **Client Culture Setting** en-US

PR #1792, opened 2016-10-05
#1791 - Updating the Excel export and PersonData methods to honor the RowFilter.

cbump:
# Context
When exporting data to Excel or using the Communication tool from a filtered DynamicData block, it always used all the data instead of using the data with the filter applied.

# Goal
This updates the PersonData and ExcelExport to used the applied RowFilter.

# Strategy
Use the DefaultView of the DataTable instead of just iterating through the Rows.

# Possible Implications
It's possible that exporting all the data is someone's desired behavior.  I think this is unlikely but it's hard to know these kinds of things.

### DIFF
--- a/Rock/Web/UI/Controls/Grid/Grid.cs
+++ b/Rock/Web/UI/Controls/Grid/Grid.cs
@@ -1563,11 +1563,11 @@ namespace Rock.Web.UI.Controls
                     }
 
                     // print data
-                    foreach ( DataRow row in data.Rows )
+                    foreach ( DataRowView rowView in data.DefaultView )
                     {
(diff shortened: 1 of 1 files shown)


mikepetersonccv:
@cbump I had already started on this but forgot to make a pull request.  Oops :)  It looks like you ended up doing the same thing except there is one more similar issue in GetEntitySetFromGridSourceDataTable @ https://github.com/secc/Rock/blob/7f3bed029538e1d4f2c4ea58f0d09b9bd51cd298/Rock/Web/UI/Controls/Grid/Grid.cs#L2474

Here is what I had done
image

PR #1794, opened 2016-10-05
Added missing lava tip.

jreciomoreno:
# Context
The single select field control is missing a lava tip although it supports lava.

# Possible Implications
None, is just a minor UI change.

image


edmistj:
Nice catch!

PR #1797, opened 2016-10-06
Group finder: show info window if no detail page

sheck:
# Context
When using a group finder block that does not have a group detail page set, clicking on a group in the grid view just reloads the page with no visible change.

# Goal
Give users better visual feedback after selecting a group, given the context

# Strategy
Show the info window when group is clicked.
In the postback, when there is no group detail page set, a javascript function is called that opens the corresponding infowindow.

# Possible Implications
None I can see

# Note
This is not a perfect solution as it would probably be better to not fire the post back at all, but I was having trouble trying to dynamically set the OnRowSelected attribute for the grid. If there is a better way to implement this, feel free to change it.


azturner:
@sheck Like this idea, have you checked by chance that it works with each of the major browsers?

sheck:
I'm going to be out of town for the next few days, but I'll take a look when I get back.

sheck:
Ah, totally forgot about this. Sorry! Yes, I checked it with a few versions of IE, Chrome, & FF — it worked as expected. Thanks!

PR #1798, opened 2016-10-06
Person Search Performance

cbump:
# Context
Person search was incredibly slow.  Even with a search that matched a single record, it took around 5 seconds for the search to execute before being redirected to the person's profile.  This PR updates the person search to first load a list of id's and then re-query for all the rest of their data as this significantly speeds up the person search block.

# Goal
This is drastically faster (and a relatively small change)

# Strategy
This does a search first just to get the person id's that matches, then does a secondary search to load all their data.

# Possible Implications
None that I know of.  It seems to work fine and improve speed no matter the size of the result set.

### DIFF
--- a/RockWeb/Blocks/Crm/PersonSearch.ascx.cs
+++ b/RockWeb/Blocks/Crm/PersonSearch.ascx.cs
@@ -262,6 +262,10 @@ namespace RockWeb.Blocks.Crm
                         }
                 }
 
+                var personIdList = people.Select( p => p.Id ).ToList();
+
+                people = personService.Queryable(true).Where( p => personIdList.Contains( p.Id ) );
+				
                 SortProperty sortProperty = gPeople.SortProperty;
                 if ( sortProperty != null )
                 {


arranf:
:+1:

RockWeb/Blocks/Crm/PersonSearch.ascx.cs
```diff
@@ -262,6 +262,10 @@ private void BindGrid()
                         }
                 }
 
+                var personIdList = people.Select( p => p.Id ).ToList();
+
+                people = personService.Queryable(true).Where( p => personIdList.Contains( p.Id ) );
```
mikepetersonccv: this might not work when there are 1000s of personIds since it is going to turn into a SELECT * FROM PERSON WHERE ID IN (1,2,3,....) type of query.  You might want to test it in that scenario.

cbump: I actually considered this and tested it by searching for the letter "e" which returned over 14,000 records with no issues.  I've seen scenarios where Linq created a whole bunch of parameters on a query which then led to situations where it surpassed the parameter limit in SQL but this seems to work fine.

PR #1800, opened 2016-10-06
Find Families Performance Enhancements

cbump:
# Context
Phone number searches during check-in were slow and put extensive load on the SQL server.  We were seeing SQL waits  and DB contention that we were pretty uncomfortable with during checkin times.

# Goal
The goal is to speed up the FindFamilies workflow action as much as possible.

# Strategy
This is a 2-fold approach to improving the FindFamilies searching. The majority of the brain-power behind this solution is from @tom-powers who helped optimize the query and is solely responsible for the phone number reversal:

- First, the query itself has been optimized to search the phonenumber table with an inner join on the person and group member to get the family id.  This as as optimized as we could find query-wise. 
- Second, we updated the PhoneNumber model to add a phone number reversed computed column to the table.  This literally stores the phone number backwards in the database, which, when searched using a reversed phonenumber "Starts With" style query, allows for the usage of an index which dramatically improves the results of this search (see the Statistics below).

# Possible Implications
This just needs to be tested well.  I've tested it pretty well on this end but more testing is better!

# Statistics
@trydyingtolive Created a phone number load test block which executed this 10 times every iteration in a loop.  These are examples of the time improvements we saw:


edmistj:
This approach makes a lot of sense. Have you run this code in production?

cbump:
@edmistj Yes, we are using it in production right now.  We have a custom FindFamilies workflow action that we use (once this PR gets merged we can use the core one again) and there were a couple changes made to FindFamilies for excluding Inactive records a few weeks ago that I merged into this specific PR.

azturner:
@cbump I pulled this in and am working on redoing the migration since we had already added more migrations since your branch. However I don't see where you were creating the computed column in a migration.  Can you post the properties of that column so I can see what the formula is for it?

cbump:
@azturner It's a persisted computed column.  If you look at how the column is created the reverse() function is actually in the definition: https://github.com/SparkDevNetwork/Rock/blob/5166f8e3d0c988b4d6b7903a1d188ca050e61be5/Rock.Migrations/Migrations/201609140843400_PhoneNumberReversed.cs#L29

Issue #1710, opened 2016-08-17
Cursor defaults to text on datepicker


garrettjohnson:
# Context
When hovering over labels, or the previous/advance arrows the cursor turns to a text selection cursor

# Expected Behaviour
Cursor to default to pointer

# Actual Behaviour
Cursor changes to text selection

* **Rock Version**: 5.2

# Screenshots
Can't show the cursor on OS X, but the month name and arrows default to text selection...
<img width="329" alt="screen shot 2016-08-16 at 9 49 02 pm" src="https://cloud.githubusercontent.com/assets/374209/17725002/e2652dfc-63fb-11e6-9fa7-7628a8df8181.png">

arranf:
Fixed in #1739

PR #1802, opened 2016-10-06
Feature mp datepicker update

mikepetersonccv:
# Context
People that use the keyboard instead of the mouse to enter in dates were getting frustrated that the date range picker wasn't cooperating with them.  For example, it was changing the enddate when the startdate was getting keyed in.  I had also noticed #1710 as an issue and thought that a datepicker update might fix that too.

# Goal
make keyboard entry of dates easier without impacting mousers and change the daterange picker to not try to set the enddate to a later date when the startdate changes

# Strategy
The pullrequest updates the bootstrap-datepicker javascript and LESS from v1.3.1 to v1.7

# Possible Implications
There was a LESS change made to pickerdate-less manually by the Rock community that this update un-does.  I'm pretty sure the stock LESS from date-picker also fixes this same issue.  Also please review how I updated the LESS file.  

# Screenshots


cbump:
Thank you for this fix!  We've heard quite a few complaints about the date picker since we launched Rock.

mikepetersonccv:
I might have accidentally closed this

PR #1807, opened 2016-10-14
Adding "New Tab" support for SignNow

cbump:
# Context

Iframe support for SignNow was pretty sketchy on mobile and IE
# Goal

This adds a block setting which will open the document in a new tab instead of the iframe
# Strategy

Added a block attribute to configure whether to use an Iframe or a new tab.  If new tab is selected, the button actually opens the document in a new tab and then the DocumentReturn.html has a little javascript to tell the window.opener what the DocumentId is.
# Possible Implications

The default is Iframe so it shouldn't cause anything to break for people who are already using this (shouldn't be too many people).
# Screenshots

image

### DIFF
--- a/RockWeb/Blocks/Event/DocumentReturn.html
+++ b/RockWeb/Blocks/Event/DocumentReturn.html
@@ -2,6 +2,13 @@
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
     <title></title>
+    <script>
+        // This handles the callback when SignNow redirects back to Rock.
+        if (window.opener && window.opener.processIframeReturn) {
+            window.opener.processIframeReturn(window.location.search);
+            window.close();
+        }
+    </script>
 </head>
 <body>
 </body>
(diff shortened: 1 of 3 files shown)


jreciomoreno:
I can confirm that this new block setting being enabled fixes the issue of the signature form not loading on mobile browsers.
Although the continuous-integration check appears as failed in here, the code compiles fine.

edmistj:
Can you post a quick video showing the interaction between tabs. I see the screenshot above for the link to open in a new tab, but I'm curious how the full UX flow looks / feels.

cbump:
You can test it in action here.  Go ahead and submit the form and everything: http://rockstaging.secc.org/Registration/T920Believe17

edmistj:
Attaching a video for others. https://www.opentest.co/share/7eb4e900ac0811e6a05477093ceb15c3

edmistj:
We made a fix in v6.2 that should negate the need for this.

cbump:
We will re-test the signature stuff in 1.6.2 and verify the iframe version is working.  Thanks!!

Issue #1819, opened 2016-10-20
[EN] Inconsistent constructor visibility for AttributeCache and AttributeValueCache


dcs619:
The constructors for [AttributeValueCache](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Web/Cache/AttributeValueCache.cs#L59) are all marked public.  The constructors for [AttributeCache](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Web/Cache/AttributeCache.cs#L56) are all marked private.

I would expect the visibility for those types to match. That 
1) simplifies creating a Model that has those properties added, and
2) allows for unit testing the AttributeCache and AttributeValueCache functionality.  

If there's a reason visibility is limited on AttributeCache I'd love to learn that too.

azturner:
The AttributeCache class is a designed to only ever have one instance for each attribute. It (just like most of the other ...Cache classes) should never be instantiated by external code. The Read() method should always be used to get an instance of the class. AttributeValueCache, on the other hand, is more of a helper class with public constructors. Attribute Values are not actually cached.

dcs619:
@azturner 👍  This was to ensure models that depend on AttributeCache are getting an instance, but maybe there's a way to mock it for testing.  Since the tests run in a headless space, there's never an actual ...Cache to call Read on.

Issue #1811, opened 2016-10-17
[Discussion] Unit Testing


johnthepink:
We want to start the process of adding unit tests to Rock! This will definitely need to be a community effort, as there are lots of things we need to get covered. Our goal for this week is to:
1. Document the types of things that need to be tested
2. Research and select a testing framework
3. Provide some examples
4. Get feedback from everyone

At the end of the week, @dcs619 and I hope to have some documentation to easily onboard people willing to help out.
## Things To Test

Here are things we initially want to cover with unit testing:
1. Entities
2. API controllers
3. Stored procedures
4. UI
5. Workflow actions

Surely this will be an evolving list. What are we missing?
## Testing Frameworks

I've done some reading on NUnit, xUnit, and MSTest. I'm starting to develop my own opinion, but I wanted to hear feedback from people with experience unit testing in .NET.
1. What do you like?
2. What is appropriate for this project?
3. Why?

arranf:
I think it'd also be good as part of this effort to identify repeated logic in places such as Workflow Actions, Blocks, and other places that could be pulled in reusable static methods, that could then be tested

johnthepink:
@arranf that sounds like a good ongoing plan. Our hope was that we could get some initial work merged, and then slowly chip away at all the things. That way we won't have a long-running branch.

dcs619:
@mikepetersonccv Any thoughts?  We're looking at your NUnit models as a potential starting point.  Model testing seems easy enough to implement, but I'm curious about workflow actions.

arranf:
I think @mrupsidedown has done some testing of check-in, and consequently
workflow actions. I don't know if he's got any strategies he can share.

mikepetersonccv:
@dcs619 I'm not sure NUnit models you are referring to.  Anyways, I don't have an opinion quite yet

dcs619:
@mikepetersonccv I was looking at the tests started [here](https://github.com/SparkDevNetwork/Rock/tree/develop/Rock.Tests), but just realized that was @nairdo's work...you had last modified it to update the license 😸

johnthepink:
## Mocking

One interesting problem we spent some time researching is mocking. Mocking will be necessary to maintain true unit tests by not making any requests to the database. Tests that do hit the database may become necessary later, but those will be integration tests.

[...]

PR #1821, opened 2016-10-20
Added initial unit testing framework

dcs619:
# Context

This PR represents the initial work being discussed in #1811.
# Goal

The goal for this PR is to provide a unit test boilerplate for the Rock community.  So far there are 44+ boilerplate tests for Entity Models and Workflow Actions. 
# Strategy

This uses the Xunit testing framework. After testing several options, we felt Xunit was a good direction given its simplified style, and the fact that every test is run in isolation. This helps enforce good testing practices.

We converted the existing tests from the Rock.Tests project and included it back in the Rock solution.  As part of the solution, Visual Studio will run the tests after a successful build of the other projects.  The tests will not run in MSBuild unless you specify it. The Rock.Tests project structure directly mimics the file structure of the Rock project (there is basically a 1:1 mapping of class file to test file).  

**Note: no database is being spun up to test against, as true unit testing does not involve any calls to the database.**

These tests currently run on [our Appveyor build](https://ci.appveyor.com/project/NewSpring/rock/build/1.3.964). If this PR is merged, you'll need to include [the test section](https://github.com/NewSpring/Rock/blob/unit-testing/appveyor.yml#L50) on the official Rock AppVeyor.
# Future Strategy

We would love to set up code coverage for the tests so we can easily know where we need to focus.  Also, some methods and/or database calls need to be mocked to be completely tested.
# Possible Implications

We've included two bug fixes in this PR that came up during testing:
1) [BlockType.Blocks ](https://github.com/NewSpring/Rock/commit/3087af0f870b9912b6b8bb55276f39b7301c1361) weren't being included in deep clones.  According to the comment on [IEntity Clone()](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Data/Entity.cs#L272), all properties should be included during a deep clone.  We added a Json decorator to the Blocks object to ensure it was included during serialization/deserialization.

2) [AttributeCacheJsonConverter](https://github.com/NewSpring/Rock/commit/ac9d9f8b3d485c74f656abdcdfa5887bad6e9b44#diff-1ac4c68a604d0eefafcdad73bf2c0cd6R75) was crashing on null values.  Prior to the fix, we had to [massage the JSON object](https://github.com/NewSpring/Rock/commit/ac9d9f8b3d485c74f656abdcdfa5887bad6e9b44#diff-562059f247d6c1e80f6642fb69c476e0R272) to test successfully.  This is related to #1819 so we're open to other implementations.


johnthepink:
Looks like the tests ran on Spark's Appveyor 🎉

dcs619:
@azturner Any feedback on this?  Do you want me to remove the bugfixes for a cleaner commit?

mikepetersonccv:
@dcs619 In my opinion, the bug fixes should be a separate pull request

dcs619:
@mikepetersonccv I've removed the bugfixes and just added comments where entities need additional work.

nairdo:
@dcs619 We're looking at this today...  Can you explain the hard-coded Skips? `[Fact( Skip = "Need the SqlServerTypes library to resolve" )]`  Are they hard-coded because there is no SqlServerTypes available to load in AppVeyor and so therefore if we want to run these tests locally in our environment we would need to remove them by hand temporarily before running?

dcs619:
@nairdo Tests with the `Skip` decorator had dependencies on low-level functions: RockWeb Cache or an activated DbContext.  In your specific case, I added the Nuget dependency for SqlServerTypes but Rock.Tests wouldn't recognize that the library was available.  

You won't have to remove any tests in your local environment for this to run.  The goal of this PR is to establish the testing framework and provide some examples for other contributors to build off of/submit PR's to.

dcs619:
Just to be clear, the `Skip` decorator lets you show what _should be tested_, but keeps the test runner from running the test.

nairdo:
@dcs619 I edited my question for greater clarity but I do understand what you're saying.

nairdo:
@dcs619 after merging this our (at Central Christian Church AZ) AppVeyor project, it no longer builds the deploy package and it basically stops right after building/running this new Rock.Tests assembly.  Was there something you had to do in order to get past this in your AppVeyor environment?

dcs619:
@nairdo did you compare your appveyor.yml before and after?  you can remove it, but the only thing it adds is this section:

```
# test
test:
  assemblies:
    - Rock.Tests\**\*.dll
```

mikepetersonccv:
I had to update our AppVeyor to ignore the new appveyor.yml that was introduced with this

dcs619:
@narido It sounds like it's creating problems for multiple orgs, just remove it.

nairdo:
@dcs619 We don't use a appveyor.yml with AppVeyor, however once we deleted this file it works/builds correctly again.  I may still remove this appveyor file from the Rock project since Rock really has no dependency (or condoned usage) with AppVeyor.

dcs619:
@nairdo +1 for deleting appveyor.yml; I think it will cause conflicts with multiple orgs.  

I added that file because Spark has an [AppVeyor service](https://ci.appveyor.com/project/sparkdevnetwork-service/rock) set up to ensure successful builds for every PR.  And I wanted to make sure the [tests](https://ci.appveyor.com/project/sparkdevnetwork-service/rock/build/1.3.9067-yoptbihe#L244) were included in the build.

PR #1837, opened 2016-10-27
Optimizing Attendance Analytics

tom-powers:
# Context

Attendance Analytics (Attendees toggle button) was not performant.  It was taking over 2 minutes to run just 2-3 weeks worth of attendance data, and it would timeout at 3 minutes and throw an error.
# Goal

To be able to run months worth of attendance data in a matter of seconds.  Exporting to Excel should also be just as quick.  
# Strategy

Rewrite and optimize the stored procedure spCheckin_AttendanceAnalyticsQuery_AttendeeLastAttendance

Eliminate redundant code and use window ranking function to find last attendance data.
# Possible Implications

In the migration I didn't put a Down method to rollout back to the previous version of the stored procedure, because I don't see why you would ever want to go back to the way it was performing.

ROW_NUMBER() function has existed since SQL 2005, so no concern about it working with Rock.
# Screenshots

No UI changes

When executed within SSMS, the previous version of the stored procedure took nearly 10 minutes to run 3 months worth of data.  With the new version, it executed in less than 1 second.
# Conflicts:
# Rock.Migrations/Migrations/RockMigrationSQL.Designer.cs
# Rock.Migrations/Migrations/RockMigrationSQL.resx
# Rock.Migrations/Rock.Migrations.csproj


dcs619:
🙏 I'd imagine something similar could be done with Giving Analytics?

mikepetersonccv:
@tom-powers We are trying this at CCV now.  We were having the same problem, around 10 minutes+, and now just a few seconds! THANK YOU!!  This is awesome!

azturner:
@tom-powers Thanks for taking a look at this and making such a drastic improvement. I wasn't able to merge this pull request as it includes a migration farther back than our current migration path. I did pull the change though into a recent migration ( https://github.com/SparkDevNetwork/Rock/commit/3398e1bdd7b3aed91accf7f2958f8663605b3048)

JimMichael:
Tom, any reason this was never re-submitted? We sure could use that perf improvement in Attendance Analytics!

PR #1838, opened 2016-10-28
Email Functionality

cbump:
# Context

We have the need to cc emails to addresses dynamically (Person attributes, Transaction attributes etc).  We also have times we would like to set the cc/bcc field via the communication entry form.
# Goal

Set Cc and Bcc fields with Lava.
# Strategy

This moves the CC/Bcc fields down into the recipient section so we can merge lava fields for each recipient iteration.  It also add's the ability to use lava in all the fields.  Lastly it adds CC and Bcc fields to the Email Communication control and allows for setting it to show or not show based on a CommunicationEntry block attribute.
# Possible Implications

# Screenshots

This is the communication entry page when the CC/Bcc functionality is enabled:
image

### DIFF
--- a/Rock/Communication/SMTPComponent.cs
+++ b/Rock/Communication/SMTPComponent.cs
@@ -170,32 +170,13 @@ namespace Rock.Communication.Transport
                         string replyTo = communication.GetMediumDataValue( "ReplyTo" );
                         if ( !string.IsNullOrWhiteSpace( replyTo ) )
                         {
-                            message.ReplyToList.Add( new MailAddress( replyTo ) );
+                            // Resolve any possible merge fields in the replyTo address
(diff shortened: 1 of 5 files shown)


nairdo:
@cbump Do you guys tend to need the Cc and Bcc field on the new communication block because you're entering them in manually each time -- or do you want them on Email Templates (similar to System Emails) so that they can be set up in advance and reused multiple times?

image

cbump:
@nairdo Right now we only use them for system emails.  The "Allow CC/BCC" block attribute isn't even set on the standard new communication page.  We have a few instances where we use lava in system emails in those fields to allow an email to be sent to both the recipient and cc someone internally at the same time.  We don't often need lava for the fields but it sure is handy when the need arises!

PR #1847, opened 2016-11-02
Added badge type which displays if person is in dataview

trydyingtolive:
# Context
We wanted to add a badge to a person's profile if they are safe to work with children. We have an existing data view which we use extensively. Rather than recreating and managing parallel Lava any changes to the data view can appear on the badge as well.

# Goal
This will allow churches to use data views to show badges. 

# Strategy
I created a badge type that displays lava enabled output if the person is in the data view.

# Possible Implications
There should be none, except for many complex badges at the same time.

### DIFF
--- a/Rock/PersonProfile/Badge/InDataView.cs
+++ b/Rock/PersonProfile/Badge/InDataView.cs
@@ -1,0 +1,75 @@
+﻿// <copyright>
+// Copyright by the Spark Development Network
+//
+// Licensed under the Rock Community License (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.rockrms.com/license
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+//
+using System.ComponentModel;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Linq;
+using System.Data.Entity;
+
+using Rock.Attribute;
+using Rock.Model;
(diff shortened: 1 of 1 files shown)


edmistj:
Sounds like a great idea!

nairdo:
@trydyingtolive Looks pretty good except I'd say it's missing a`entityTypeName: "Rock.Model.Person"`  in the DataViewField which currently prevents the list of available DataViews from being populated.  I'd also set the attribute field to be required-true:

```[DataViewField( "Data View", "The dataview to use as the source for the query. Only those people in the DataView will be given the badge.", true, entityTypeName: "Rock.Model.Person", order: 0 )]```

Once you correct that I'll merge it in right away.

File: Rock/PersonProfile/Badge/InDataView.cs [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+using System.Diagnostics;
+using Rock.Web.Cache;
+
+namespace Rock.PersonProfile.Badge
+{
+
+    /// <summary>
+    /// 
+    /// </summary>
+    [Description( "Displays if the person has an alert note." )]
```
nairdo: correct description

Rock/PersonProfile/Badge/InDataView.cs
```diff
@@ -0,0 +1,75 @@
...
+    [Export( typeof( BadgeComponent ) )]
+    [ExportMetadata( "ComponentName", "In Data View" )]
+
+    [DataViewField( "Data View", "The dataview to look at.", order: 0 )]
```
nairdo: Missing arg entityTypeName: "Rock.Model.Person" which currently prevents the list of available DataViews from being populated. Also set the required field to be true:

```csharp
[DataViewField( "Data View", "The dataview to use as the source for the query. Only those people in the DataView will be given the badge.", true, entityTypeName: "Rock.Model.Person", order: 0 )]
```

PR #1848, opened 2016-11-03
- Increased the database timeout when generating merge templates.

tcavaletto:
# Context
We were continuously getting RequestTimedOut errors while trying to generate letters for 8k+ people, and needed the DatabaseTimeout to be increased so that Rock could handle the load.

# Goal
This pull request increases the DatabaseTimeout so that Rock has the time to generate a merge template for each row.

# Strategy
We've implemented this by changing the DatabaseTimeout on the RockContexts in the CreateDocument() methods to be 10 minutes instead of 3, and also updated the MergeTemplateEntry block's timeout properties so that the block doesn't timeout while Rock is still processing data.

# Possible Implications
People may have to be careful to not exit out of their tab while the merge template is being generated.

### DIFF
--- a/Rock/MergeTemplates/HtmlMergeTemplateType.cs
+++ b/Rock/MergeTemplates/HtmlMergeTemplateType.cs
@@ -61,8 +61,6 @@ namespace Rock.MergeTemplates
             BinaryFile outputBinaryFile = null;
 
             var rockContext = new RockContext();
-            rockContext.Database.CommandTimeout = 600;
-
             var binaryFileService = new BinaryFileService( rockContext );
 
             var templateBinaryFile = binaryFileService.Get( mergeTemplate.TemplateBinaryFileId );
--- a/Rock/MergeTemplates/HtmlMergeTemplateType.cs
+++ b/Rock/MergeTemplates/HtmlMergeTemplateType.cs
@@ -61,6 +61,8 @@ namespace Rock.MergeTemplates
             BinaryFile outputBinaryFile = null;
 
             var rockContext = new RockContext();
+            rockContext.Database.CommandTimeout = 600;
+
             var binaryFileService = new BinaryFileService( rockContext );
 
             var templateBinaryFile = binaryFileService.Get( mergeTemplate.TemplateBinaryFileId );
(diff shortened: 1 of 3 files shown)


Rock/MergeTemplates/HtmlMergeTemplateType.cs
```diff
@@ -61,6 +61,8 @@ public override BinaryFile CreateDocument( MergeTemplate mergeTemplate, List<obj
             BinaryFile outputBinaryFile = null;
 
             var rockContext = new RockContext();
+            rockContext.Database.CommandTimeout = 600;
```
mikepetersonccv: this shouldn't be necessary, getting a BinaryFile by Id should take less than a second

Rock/MergeTemplates/WordDocumentMergeTemplateType.cs
```diff
@@ -73,6 +73,8 @@ public override BinaryFile CreateDocument( MergeTemplate mergeTemplate, List<obj
             BinaryFile outputBinaryFile = null;
 
             var rockContext = new RockContext();
+            rockContext.Database.CommandTimeout = 600;
+
```
mikepetersonccv: same thing here, this shouldn't be necessary, getting a BinaryFile by Id should take less than a second

RockWeb/Blocks/Core/MergeTemplateEntry.ascx.cs
```diff
@@ -140,6 +147,7 @@ protected void btnMerge_Click( object sender, EventArgs e )
             // NOTE: This is a full postback (not a partial like most other blocks)
 
             var rockContext = new RockContext();
+            rockContext.Database.CommandTimeout = 600;
```
mikepetersonccv: this could be a block setting,  for example https://github.com/SparkDevNetwork/Rock/blob/d94bfb29f4b376d6596bf1d06a285ff29c208863/RockWeb/Blocks/Reporting/ReportDetail.ascx.cs#L42

nairdo: @tcavaletto can you step through an example of the problem you're seeing and try to isolate where/when the timeout occurs?  If it's only in the Merge_Click of the MergeTemplateEntry block then, a block setting with a default of 600 is probably a good way to go.

PR #1849, opened 2016-11-07
Bugfix in ContributionStatementLava Pledges

bwitting:
# Context
On the ContributionStatementLava block, the pledge amount given number is incorrect - it showed ALL gifts to that fund, not just for the individual/giving group.

# Goal
Fix ContributionStatementLava.ascx.cs so only pledges for current giving group are shown.
# Strategy
Alter the LINQ statement to only get transaction for the current giving group.

# Possible Implications
none


azturner:
@nairdo I agree.

PR #1851, opened 2016-11-07
Multiple improvements to grid export files

chrismcgrath:
# Contributor Agreement

Yes

# Context

Initially, I found that exporting a Dynamic Data Block resulted in a spreadsheet that did not have any Date or Number formatting. I also found that a single formatting for Dates and Numbers did not always make sense, such as showing the hours and minutes on a BirthDate column, or showing some currency columns with decimal and others as text with a dollar sign. Finally, I found some other miscellaneous bugs such as extra rows/columns being added, and the column headers not being included in the frozen header section.

# Goal

All grids will intelligently format their exported spreadsheets in a common way, no matter what kind of data source is used to populate the grid.

# Strategy

* Improved formatting of dates and numbers to choose the most logical formatting based on the data in the column
	- Column will be formatted to show the date only (no time) if none of the fields in the column have a time component
	- Column will show decimal places only if there is a decimal value in any of the cells of the column, or it's a currency field
	- Template fields that are populated using the FormatAsCurrency() function will format the value as a decimal rather than text using the ReverseCurrencyFormatting() function
* Added proper formatting of dates and numbers for dynamic data block exports
* Worksheet title, header text, and file name will now use title of the page the grid is on if there is no grid caption set.
* No longer adds an extra row and column in some cases
* Now correctly includes the column headers in the frozen header section
* Created an ExcelHelper class to house common Excel exporting functionality (see https://github.com/SparkDevNetwork/Rock/pull/1852)

# Possible Implications

I tried to think of the most common uses of number columns and format them in a meaningful way. I'm sure that not every possibility has been covered here, but the result should be an improvement over totally unformatted columns. The only place the data itself is being exported differently is for grids like the Event Registration grid where dollar amounts are displayed as text in a template field, but are now exported as decimals.

# Screenshots
## Example 1
Example showing Data View of pledge data
**Before**
image
> Note the position of the freeze pane (light gray line) above the header on the third row.

**After**
image
> Note the corrected position of the freeze pane (light gray line) _below_ the header on the third row.

## Example 2
Showing Report based on previous example's Data View.
**Before**
image

**After**
image

# Documentation

### DIFF
--- a/Rock/Rock.csproj
+++ b/Rock/Rock.csproj
@@ -1135,6 +1135,7 @@
     <Compile Include="UniversalSearch\SearchFieldCriteria.cs" />
     <Compile Include="UniversalSearch\SearchResultModel.cs" />
     <Compile Include="Utility\DebugHelper.cs" />
+    <Compile Include="Utility\ExcelHelper.cs" />
     <Compile Include="Utility\FileUtilities.cs" />
     <Compile Include="Web\UI\RockTheme.cs" />
     <Compile Include="Web\Cache\AttributeValueCache.cs" />
--- a/Rock/Utility/ExcelHelper.cs
+++ b/Rock/Utility/ExcelHelper.cs
@@ -10,6 +10,9 @@ using System.Linq;
 
 namespace Rock.Utility
 {
+    /// <summary>
+    /// Methods that facilitate generating Excel files
+    /// </summary>
     public static class ExcelHelper
     {
         public const string GeneralFormat = "General";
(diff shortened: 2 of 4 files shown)


azturner:
@chrismcgrath This appears to have some issues if the grid does not have any records. Realize that a user probably wouldn't bother exporting an empty table, but if they do, it should probably still show excel file with headers rather than causing an error. Is that something you can take a look at?

chrismcgrath:
I should have tested that... I'll take a look this week.

Get Outlook for Android<https://aka.ms/ghei36>



On Mon, Dec 5, 2016 at 6:22 PM -0600, "David Turner" <notifications@github.com<mailto:notifications@github.com>> wrote:


@chrismcgrath<https://github.com/chrismcgrath> This appears to have some issues if the grid does not have any records. Realize that a user probably wouldn't bother exporting an empty table, but if they do, it should probably still show excel file with headers rather than causing an error. Is that something you can take a look at?

-

dcs619:
@chrismcgrath I can't tell you how many tickets I just closed after pulling in this fix!  Thanks so much!

chrismcgrath:
@azturner I was just looking at this and not able to reproduce the problem you describe. Could you let me know if this is still an issue and exactly how to reproduce it, including which grid you're trying to export? Thanks!

PR #1852, opened 2016-11-07
Created a workflow action that will generate an Excel file from an SQL statement

chrismcgrath:
# Contributor Agreement

Yes

# Context

Our accounting department needed a four-tab report that could not be easily generated from Rock; the use of multiple reports or grids to export the data would be too tedious for the users.

# Goal

This new workflow action will take an SQL statement that returns one or more data sets and generate a single Excel file with a tab for each result. 

# Strategy

Using the ExcelHelper class I created for this pull request https://github.com/SparkDevNetwork/Rock/pull/1851 I have the action generate an Excel file and save it to the database as a binary file. The URL of the file is stored in an attribute so the user can be redirected to it, or so a link to the file can be displayed on the screen.

# Possible Implications

The usual security concerns with allowing someone to craft and run SQL.

# Screenshots

# Documentation

### DIFF
--- a/Rock/Rock.csproj
+++ b/Rock/Rock.csproj
@@ -1135,6 +1135,7 @@
     <Compile Include="UniversalSearch\SearchFieldCriteria.cs" />
     <Compile Include="UniversalSearch\SearchResultModel.cs" />
     <Compile Include="Utility\DebugHelper.cs" />
+    <Compile Include="Utility\ExcelHelper.cs" />
     <Compile Include="Utility\FileUtilities.cs" />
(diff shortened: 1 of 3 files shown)


nairdo:
Thanks @chrismcgrath I think you may have talked with Jon already about this type of plugin.  We'd really love to see this as a separate [plugin/package](http://www.rockrms.com/Rock/Developer/Book/26/26) made available in the Rock Shop from Willow Creek.  We know creating an maintaining plugins does "cost" a little more time and energy (package, document, maintain, etc), but we feel this is the best approach for Rock's long term health for this sort of plugin.

PR #1861, opened 2016-11-11
Extended DynamicChart block to allow for arbitrary x-axis values

cabal95:
# Contributor Agreement

Yes

# Context

DynamicChart in Line-mode does not allow for arbitrary X-axis values, they must be in DateTime format. This creates problems when trying to do weekly year-over-year charts. For example, weekly attendance over the past 3 years. Three lines: Current Year, Last Year, 2 Years Ago. X-axis is week numbers, 1-52 (actually 53 because of odd start weeks).

# Goal

Provide support in the DynamicChart block to allow SQL queries to provide non-DateTime x-axis values.

# Strategy

Support new XValue column in addition to the existing DateTime column. If DateTime column is provided it is used, otherwise the XValue column is used if it exists.

Additionally, adds support for MetricTitle to be used in the tooltip instead of the DateTime column.

Finally, unrelated to original need but made sense for our use case, adds support for a YValueFormatted column that is used, if present, in the tooltip instead of the YValue. This allows the SQL to format the value in a user-friendly manner, for example as a $12,345.67 dollar amount instead of "12345.67"

# Possible Implications

No security considerations. Only possible break to backwards compatibility would be to a SQL query that provides a MetricTitle column for use in Pie Chart that is switched to Line Chart; but I think the query would not display right as a line to begin with.

# Screenshots

sample

# Documentation

In-block documentation (sample SQL) might need to be updated. I have not done this as I wasn't sure how to present this new information as optional but not clutter up the example SQL.

### DIFF
--- a/RockWeb/Blocks/Reporting/DynamicChart.ascx.cs
+++ b/RockWeb/Blocks/Reporting/DynamicChart.ascx.cs
@@ -191,6 +191,14 @@ function labelFormatter(label, series) {
             public decimal? YValue { get; set; }
 
             /// <summary>
+            /// Gets the y value as a formatted string (for Line and Bar Charts)
+            /// </summary>
+            /// <value>
+            /// The formatted y value.
+            /// </value>
+            public string YValueFormatted { get; set; }
+
+            /// <summary>
             /// Gets or sets the metric title (for pie charts)
             /// </summary>
             /// <value>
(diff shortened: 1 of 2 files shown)


mikepetersonccv:
cool!

edmistj:
This is fine, but would warn that as the documentation notes the dashboard components are still in a beta and there will be some breaking changes coming as we move from Flot to Chart.js.

From the docs: "Feel free to play with these features but don't roll them out in production just yet."

PR #1863, opened 2016-11-14
- Added the ability to pass parameters into the SQL used to generate the DynamicChart charts.

tcavaletto:
# Contributor Agreement
Yes

# Context
We're working on generating campus-specific dashboard pages that contain DynamicChart blocks, and wanted the ability to pass in campus Ids so that we could generate charts specific to that campus.

# Goal
This change allows users to pass in custom parameters into the SQL used to generate DynamicCharts. Not only does this allow users to pass in page parameters, but it also allows them to pass in lava-based parameters such as DateTime values, etc.

# Strategy
We implemented this by copying the same functionality that the DynamicData block uses over to the DynamicChart block.

# Possible Implications
Depending on how you set up your parameters in the block setting, you could potentially cause your sql to error out (e.g.) passing an nvarchar parameter where your sql is expecting a datetime value). These are the same implications that already exist in the DynamicData block.

# Screenshots
image

# Documentation
This does not affect any of the UI screenshots or descriptions in the documentation / user guides.

### DIFF
--- a/RockWeb/Blocks/Reporting/DynamicChart.ascx.cs
+++ b/RockWeb/Blocks/Reporting/DynamicChart.ascx.cs
@@ -94,6 +94,7 @@ order by YValue desc
 
             var pageReference = new Rock.Web.PageReference( this.PageCache.Id );
             pageReference.QueryString = new System.Collections.Specialized.NameValueCollection();
+            pageReference.QueryString.Add( this.Request.QueryString );
             pageReference.QueryString.Add( "GetChartData", "true" );
             pageReference.QueryString.Add( "GetChartDataBlockId", this.BlockId.ToString() );
             pageReference.QueryString.Add( "TimeStamp", RockDateTime.Now.ToJavascriptMilliseconds().ToString() );
--- a/RockWeb/Blocks/Reporting/DynamicChart.ascx.cs
+++ b/RockWeb/Blocks/Reporting/DynamicChart.ascx.cs
@@ -69,7 +69,7 @@ order by YValue desc
 </pre>
 </code>",
               CodeEditorMode.Sql )]
-
+    [TextField( "SQL Params", "Parameters to pass to query", false, "" )]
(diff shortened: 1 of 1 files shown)


RockWeb/Blocks/Reporting/DynamicChart.ascx.cs
```diff
@@ -69,7 +69,7 @@ PageView pv
 </pre>
 </code>",
               CodeEditorMode.Sql )]
-
+    [TextField( "SQL Params", "Parameters to pass to query", false, "" )]
```
nairdo: For consistency with the DynamicData block change:
* SQL Params to Query Params
* Change description text to match the corresponding help text found in the tbParams (RockTextBox) of the DynamicData.ascx:  "The parameters that the stored procedure expects in the format of 'param1=value;param2=value'. Any parameter with the same name as a page parameter (i.e. querystring, form, or page route) will have it's value replaced with the page's current value. A parameter with the name of 'CurrentPersonId' will have it's value replaced with the currently logged in person's id."

RockWeb/Blocks/Reporting/DynamicChart.ascx.cs
```diff
@@ -328,6 +330,52 @@ protected void Block_BlockUpdated( object sender, EventArgs e )
         }
 
         /// <summary>
+        /// Gets the parameters.
+        /// </summary>
+        /// <returns></returns>
+        private Dictionary<string, object> GetParameters()
+        {
+            string[] queryParams = GetAttributeValue( "SQLParams" ).SplitDelimitedValues();
```
nairdo: For consistency with the DynamicData block, change to "QueryParams"

nairdo: Sorry.  I unintentionally cleared the review.

tcavaletto:
Requested changes made.

PR #1867, opened 2016-11-15
- small fix to EntityHelper where it could return incorrect cached re…

mikepetersonccv:
# Contributor Agreement
_Have you filled out and sent your [Spark Contributor Agreement](http://www.rockrms.com/Content/RockExternal/Misc/Contributor%20Agreement.pdf) to secretary [at] sparkdevnetwork.org?_ YOU BET!

# Context
Noticed that EntityHelper could return wrong results if it was multiple times in the same request with different values for includeReportingFields or limitToFilterableFields.  I'm not sure how often that happens, if at all, but it could happen.

# Goal
Fix it

# Strategy
included the values of includeReportingFields or limitToFilterableFields in the CacheKey

# Possible Implications
This is a pretty safe change

# Screenshots

# Documentation

…sults if the includeReportingFields or limitToFilterableFields where specified


Rock/Reporting/EntityHelper.cs
```diff
@@ -50,7 +50,7 @@ public static List<EntityField> GetEntityFields( Type entityType, bool includeOn
 
             if ( HttpContext.Current != null )
             {
-                entityFields = HttpContext.Current.Items[string.Format( "EntityHelper:GetEntityFields:{0}", entityType.FullName )] as List<EntityField>;
+                entityFields = HttpContext.Current.Items[string.Format( "EntityHelper:GetEntityFields:{0}_{1}_{2}", entityType.FullName, includeOnlyReportingFields, limitToFilterableFields )] as List<EntityField>;
```
mikepetersonccv: @nairdo Yes, thanks for catching this.  I've committed another change to address this.

PR #1868, opened 2016-11-15
- small optimization the Grid.SetLinqDataSource so that it doesn't ha…

mikepetersonccv:
# Contributor Agreement
You betcha

# Context
After doing the SetLinqDataSource originally, we knew that the downside is that it would have to 2nd database call to query the COUNT.  A few months ago, I realized that it wouldn't have to the 2nd call if the Count could be safely determined based on the PageSize, Page, and how many records were fetched.  It didn't make a real big difference, so I just stashed the idea away.

# Goal
Try to avoid the 2nd COUNT call if it can be done safely.

# Strategy
If the Grid is on the "LAST" page, and the number of fetched records is less than the page size, then we can calculate the item count without the COUNT call.  In those cases, it would be calculated as follows:

this.VirtualItemCount = ( this.PageIndex * this.PageSize ) + currentPageData.Count;

# Possible Implications
I think the code is pretty safe, but I might have missed something, and in most cases the performance benefit is small. But I wanted throw this out there and see what you guys thought

# Screenshots

# Documentation

…ve to query the COUNT if it can be determined based on the fetched count (and it is on the "Last" page of data)

### DIFF
--- a/Rock/Web/UI/Controls/Grid/Grid.cs
+++ b/Rock/Web/UI/Controls/Grid/Grid.cs
@@ -947,7 +947,15 @@ namespace Rock.Web.UI.Controls
                 this.AllowCustomPaging = true;
                 var currentPageData = qry.Skip( this.PageIndex * this.PageSize ).Take( this.PageSize ).ToList();
                 this.DataSource = currentPageData;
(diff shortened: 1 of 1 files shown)


edmistj:
+1

PR #1872, opened 2016-11-16
Added transaction code to giving history

tom-powers:
# Contributor Agreement
yes - just sent

# Context
Our accounting dept wanted the transaction code information to display in giving history grid

# Goal
Make accounting happy, and benefit other churches as well who need the same information

# Strategy
Added a block setting that allows for transaction code column to not be shown (shown by default)

# Possible Implications
none

# Screenshots
transaction code

# Documentation
na


edmistj:
+1

PR #1876, opened 2016-11-17
Added View Profile button to Person Picker results and a profile link on the name on the Transaction Matching block.

chrismcgrath:
# Contributor Agreement

Yes

# Context

When entering or matching transactions, sometimes it's not clear whether a person with a matching name is the correct person, even after looking at the extended info that displays when selecting an individual. It can often require looking at the family, the address history, or the contribution history to determine whether this is the person the user wants.

# Goal

Adds links that open a profile in a new window so the user can investigate a person's information without having to search again or lose their place.

# Strategy

Added a View Profile button to the results in the person picker. Right now it goes at the bottom because I could not figure out the styling to have it display nicely under the photo or under the member status tag.

Also added a profile link around the person name that displays on the Transaction Matching screen after a name has been selected. 

# Possible Implications

None

# Screenshots

Person Picker:
image

Transaction Matching:
image

# Documentation

…


nairdo:
Thanks @chrismcgrath we had a chance to discuss this today... can you make this an optional setting (EnableProfileLink; default false) on the PersonPicker? ( and -- it's OK if you want to hard-code the PersonPicker control declaration in the TransactionMatching block to set the flag to true.)

nairdo:
@chrismcgrath let me know if you've had a chance to review this and if you're still interested in doing this.  To be clear, this is what we are asking for:

1. The PersonPicker would need a new EnableProfileLink property (default false) 
2. The personPicker.js would need to pass this bool along to the PeopleController's Search method.
3. The Search method would pass the bool to its private methods where your new View Profile button is added.

chrismcgrath:
I plan on doing this, but I'll be off for the holidays and I've been and will be swamped with year-end accounting work for at least the first half of January. Sorry for the lagging!

nairdo:
@chrismcgrath We're going to pull this in (even though it's not an optional setting on the PersonPicker user control) but Jon's going to restyle the "view profile" button to something less intrusive.

PR #1892, opened 2016-11-29
- Prevent CalculateMetrics from creating duplicates

dcs619:
# Contributor Agreement

yes

# Context

When calculating Metrics via Source SQL and the MetricValueDateTime column contained more than one date, the job could create duplicate metric values.

# Goal

We have a metric that runs every night to calculate volunteer attendance.  Most of the attendance is on Sunday, but there are some volunteer opportunities during that week that need to be tracked.  That attendance may be checked in live, but can also be posted later by a staff member.  With this change, we can calculate attendance and replace existing values for multiple dates.

# Possible Implications

I've removed the limitation that the MetricValueDateTime column only contain a single date. @mikepetersonccv may be able to provide clarification why that limitation was there.


nairdo:
Yes, I'd like to get @mikepetersonccv 's eyes on this change before I merge it.

mikepetersonccv:
@dcs619 @nairdo Yep, this is a good fix.  Thanks!

PR #1899, opened 2016-12-02
Person Badges on Connection Request Detail.

cbump:
# Contributor Agreement
Yes

# Context
We had a request from ministries to show person detail badges on the connection request detail block.  This PR adds the ability to show person badges on the connection request detail.

# Strategy
I had to change the ConnectionRequestDetail block to be a PersonBlock in order to place badges on it.  Then loaded the person from the connection request.

# Possible Implications
One of the changes I had to make was to remove the using around RockContext because the badge needed a person entity that was still attached to an active context just in case attributes and defined values and such needed to be loaded to display the Lava.

# Screenshots
image

# Documentation
The connection request block can be configured to show person badges by editing the block, checking which Person Badges to show, and then saving.  By default no badges are displayed so the Person Info section won't even show up.  We still might need to update the document here:
http://www.rockrms.com/Rock/BookContent/30/79#working-with-requests

### DIFF
--- a/RockWeb/Blocks/Connection/ConnectionRequestDetail.ascx
+++ b/RockWeb/Blocks/Connection/ConnectionRequestDetail.ascx
@@ -27,7 +27,6 @@
                 <Rock:PanelDrawer ID="pdAuditDetails" runat="server"></Rock:PanelDrawer>
 
                 <div class="panel-body">
-
                     <div class="row">
                         <div class="col-md-6">
                             <div class="row">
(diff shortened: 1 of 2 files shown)


nairdo:
+1

edmistj:
FYI... I moved the location a bit via https://github.com/SparkDevNetwork/Rock/commit/d46b9141bd03abe47e47f43ee3d3aff4a4589821 to make the location more consistent with other blocks.

PR #1900, opened 2016-12-02
Fix occassional Access Denied JS error in HtmlEditor Rock Pickers

ccvct:
# Contributor Agreement
Yep!

# Context
In some situations, the Rock File Picker would fail to paste the link when a File was selected.  We narrowed it down to cases when there was a Vimeo already in the HTML

# Goal
This pull request will fix it so that a File and an Image can be picked even if there is a Vimeo already in it (or whatever)

# Strategy
Narrowed it down to a javascript 'Access Denied' that sometimes occurred when trying to JQuery something inside an IFRAME.  There are some situations where this isn't allowed.  So, the function was rewritten to avoid JQuerying into an IFRAME, so this type of problem won't occur

# Possible Implications
I tested this in IE, FF, and Chrome for both the Image and File Picker and it still works.

# Screenshots

# Documentation


ccvct:
oops, wrong account, I'll redo this pull request as mpeterson

PR #1905, opened 2016-12-03
Added an optimization to GetMappedLocationsByGeofences

trydyingtolive:
# Contributor Agreement
Yes

# Context
We have just shy of 200,000 group locations in our database. We are looking into connecting people in their neighborhoods and we noticed that searching for all people in a small geofence takes about 30 seconds.  

# Goal
We want to find all of the groups within an area quickly.

# Strategy
 We suspected that SQL Server 2016 was doing a full table search on all of the GroupLocation.
Locations. There is a spacial index on the geopoint column, but for some reason SQL Server isn't using that index. Upon investigation we discovered that the pagination of the grid (line 951 of Grid.cs) causes SQL Server to not use the index. That is when Entity Framework adds in "TOP 50" to the SQL it causes a full table scan. Normally we would add an index hint to SQL, but EF does not support hints. 

The strategy to make SQL Server use the indexes is to get the Ids of all intersecting locations, and then get all of the entities afterward. This changes searching for groups within a neighborhood from about 30 seconds on our database to trivial time. There is a check to see if the first query returns more than 10,000 Ids. If so it falls back on the original method, as too many Ids will cause problems and isn't much faster than a full table scan anyway. 

# Possible Implications
Searching in an area that contains over 10,000 group locations will be slower because of the first search, but it wont be fast anyway because of the large number.

### DIFF
--- a/Rock/Model/GroupLocationService.Partial.cs
+++ b/Rock/Model/GroupLocationService.Partial.cs
@@ -57,24 +57,48 @@ namespace Rock.Model
         public IQueryable<GroupLocation> GetActiveByLocations( List<int> locationIds )
         {
             return Queryable().Where( g =>
-                    locationIds.Contains(g.LocationId) &&
+                    locationIds.Contains( g.LocationId ) &&
                     g.Group.IsActive );
         }
 
         /// <summary>
-        /// Gets the 'IsMappedLocation' locations that are within and of the selected geofences
+        /// Gets the 'IsMappedLocation' locations that are within any of the selected geofences
         /// </summary>
         /// <param name="geofences">The geofences.</param>
         /// <returns></returns>
         public IQueryable<GroupLocation> GetMappedLocationsByGeofences( List<DbGeography> geofences )
         {
-            return Queryable()
-                .Where( l =>
-                    l.IsMappedLocation &&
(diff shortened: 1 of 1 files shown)


mikepetersonccv:
@trydyingtolive We added a way to apply Query Hints a while ago that you might be able to use.  Take a look at https://github.com/SparkDevNetwork/Rock/blob/cdde7a2c8033744a35021175e240af32e9145b20/Rock/Data/Interception/QueryHintScope.cs#L32

trydyingtolive:
We looked into using this. The problem is that the context is created in the DataView class but then gets executed in the Grid class. There would need to be a way to wrap the whole thing in the using and pass in the context.

mikepetersonccv:
@trydyingtolive oh yep, you're right.  Back when I added the QueryHintScope, I ran into the same problem.  Seems like there should be a way to do this, but in the meantime the prefetch of ids works :)

mikepetersonccv:
@trydyingtolive Just thinking out loud, but if there was an Advanced "Query Hint" option on DataView and/or Report, would that work? In other words, the DataViewDetail UI (and/or ReportDetail UI) would have a textbox where you could specify the query hint, then that would get used when the Grid is bound.

PR #1906, opened 2016-12-03
Added a person report filter to find people within a geofence

trydyingtolive:
# Contributor Agreement
Yes

# Context
We need a way to find everyone within a neighborhood so we can connect them to their neighbors.

# Goal
This will allow us to use or create a geofence in a dataview or dynamic report to locate people.

# Strategy
I created a new filter similar to the ingroupgeofence filter

# Possible Implications
It should have no additional implications.

# Important:
This request requires #1904 in order to work properly.

### DIFF
--- a/Rock/Reporting/DataFilter/Person/InLocationGeofenceFilter.cs
+++ b/Rock/Reporting/DataFilter/Person/InLocationGeofenceFilter.cs
@@ -1,0 +1,227 @@
+﻿// <copyright>
+// Copyright by the Spark Development Network
+//
+// Licensed under the Rock Community License (the "License");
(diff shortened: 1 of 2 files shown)


nairdo:
@trydyingtolive Looks good Mark, except when you set the filter to use GeoFence (instead of a Named Location).  In this case, it does not retain the fence when you reopen the picker (after saving the DataView).  It's just a small problem in the SetSelection method -- the GetBestPickerModeForLocation method returns a mode so I'll fix it to this and merge it in:
`locationPicker.CurrentPickerMode = locationPicker.GetBestPickerModeForLocation( location );`

PR #1918, opened 2016-12-09
Added null check to valueControl

trydyingtolive:
# Contributor Agreement
Yes

# Context
When adding a group member attribute that is a single select field type, if you do not give it any values and select show in grid, rock will throw an error. This is because the field type value control returns null if there are no values, and filter type does not null check the control before attempting to add it to the parent

# Goal
To keep Rock from exploding in this situation.

# Strategy
I added a null check. I had considered changing the field type to return a control instead of null, but the null check is probably needed anyway.

# Possible Implications
None.

### DIFF
--- a/Rock/Field/FieldType.cs
+++ b/Rock/Field/FieldType.cs
@@ -286,7 +286,10 @@ namespace Rock.Field
             col2.ID = string.Format( "{0}_col2", id );
             row.Controls.Add( col2 );
             col2.AddCssClass( col2Class );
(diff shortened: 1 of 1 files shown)


nairdo:
Thanks Mark @trydyingtolive  Since this is actually a bug fix I'm going to close this pull request and just implement the fix into the hotfix-1.6.1 branch (which is about to be 'frozen' -- after which we'll create a hotfix-1.6.2 for any new fixes you submit a pull request for).

PR #1929, opened 2016-12-14
Added a block setting and overload to allow user to hide inactive cam…

treyhendon:
# Context
Needed to hide some "office only" campuses from the Connection Search Filter.

# Goal
Only display Active Campuses to public.

# Strategy
Added block setting and used overload for `CampusCache.All( _blockSetting_ )`.

# Possible Implications
None.  The existing behavior is left the same - defaults to `true`.


nairdo:
@treyhendon For this one, I'm going to add a migration to create the new block setting.  What this means is if you are running this now in the wild, once this migration runs (in v7) it will remove any existing block setting/value so you'll have to re-set your block's setting value at that time.

treyhendon:
Sounds good, thanks!  Is there any doc or guideline for when a migration is preferred?
In this case I honestly never even thought of a migration since the block setting was optional and has a default.  I'd love to learn when to take each course of action.  Thanks!

nairdo:
In general we'll usually create a migration for any new block setting. Otherwise it's difficult to reference it (if we need to) in a future change. Unfortunately it's essentially impossible for anyone but the coordinated core team to add an EF migration due to their sensitive ordering needs.

PR #1932, opened 2016-12-16
Fix inconsistent check-in behavior that's not returning selected GroupTypes in CheckInPerson.cs

dcs619:
# Contributor Agreement

Yes.

# Context

When using the new `PossibleSchedules` and `SelectedForSchedule` fields to implement check-in room balancing, I wasn't getting back the GroupTypes that had been selected.  

This was because CheckInPerson.GetGroupTypes is inconsistent with the pattern used in [CheckInGroup.GetLocations](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/CheckIn/CheckInGroup.cs#L143), [CheckInLocation.GetSchedules](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/CheckIn/CheckInLocation.cs#L201), etc.

# Goal

This solves the issue of CheckInPerson not returning selected GroupTypes and makes the behavior consistent with the other CheckIn methods.

# Strategy

Removed the check for `if ( PossibleSchedules.Any() )` and changed to the pattern consistent with the other CheckInModels:
``` 
if ( selectedOnly ) 
{
    return GroupTypes.Where( t => t.Selected || t.SelectedForSchedule.Any( s => selectedScheduleIds.Contains( s ) ) ).ToList();
}
```

# Possible Implications

Unknown why this method was written differently.  This behaves as expected once fixed.

### DIFF
--- a/Rock/CheckIn/CheckInPerson.cs
+++ b/Rock/CheckIn/CheckInPerson.cs
@@ -209,13 +209,8 @@ namespace Rock.CheckIn
         {
             if ( selectedOnly )
             {
-                if ( PossibleSchedules.Any() )
-                {
-                    var selectedScheduleIds = SelectedSchedules.Select( s => s.Schedule.Id ).ToList();
-                    return GroupTypes.Where( t => t.SelectedForSchedule.Any( s => selectedScheduleIds.Contains( s ) ) ).ToList();
-                }
-
-                return GroupTypes.Where( t => t.Selected ).ToList();
+                var selectedScheduleIds = SelectedSchedules.Select( s => s.Schedule.Id ).ToList();
+                return GroupTypes.Where( t => t.Selected || t.SelectedForSchedule.Any( s => selectedScheduleIds.Contains( s ) ) ).ToList();
             }
 
             return GroupTypes;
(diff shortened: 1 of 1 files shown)


nairdo:
As far as I can tell, the original change was such that the PossibleSchedules contains only those schedules that apply to the particular person checking in: https://github.com/SparkDevNetwork/Rock/commit/c23a4021d2ce7be96a30bae8c431c113f942f26f#diff-0ef9c24bec1873aea7f3e1806731f48aR101  Before I merge this pull request, I'm going to request David do a review on this proposed removal of the _PossibleSchedule-check_ in the CheckInPerson's GetGroupTypes method.

dcs619:
@nairdo Correct, if PossibleSchedules was empty then it would return selected GroupTypes correctly.  

To provide context, we use PossibleSchedules to see what's available at a certain service.  If a person is allowed to check-in to multiple groups, and another Grouptype.Group has a lower attendance count at the current service, we room balance to the other group.

Without this change, I'd need to add code for every section that sets GroupTypes.Selected to also add/remove SelectedSchedules.  That seems redundant and problematic for the future.

PR #1933, opened 2016-12-16
New Group -> GroupType DataView related filter

cbump:
# Contributor Agreement
Yes

# Context
There was now GroupType related DataView filter when you built a DataView for Groups.

# Goal
This creates a new "Related Data Views" filter when building Groups Data Views that lets you select a Group Type dataview.

# Strategy
I more-or-less copied an existing related dataview filter and implemented it to work for this situation.

# Possible Implications
None.  Just new functionality.

# Screenshots
image

# Documentation
We may need to update the reporting documentation . . .  Also, this should probably get documented in the release notes for the next version of Rock.

### DIFF
--- a/Rock/Reporting/DataFilter/Group/GroupTypeDataViewFilter.cs
+++ b/Rock/Reporting/DataFilter/Group/GroupTypeDataViewFilter.cs
@@ -224,8 +224,8 @@ function() {
             // Define Control: History Data View Picker
             var ddlDataView = new DataViewPicker();
             ddlDataView.ID = filterControl.GetChildControlInstanceName( _CtlDataView );
-            ddlDataView.Label = "Connected to Group Types";
-            ddlDataView.Help = "A Data View that provides the list of GroupTypes to which the Group may be connected.";
+            ddlDataView.Label = "Has a Group Type in this Data View";
+            ddlDataView.Help = "A Data View that provides the set of Group Types to match.";
 
             filterControl.Controls.Add( ddlDataView );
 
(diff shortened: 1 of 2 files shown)


Rock/Reporting/DataFilter/Group/GroupTypeDataViewFilter.cs
```diff
@@ -0,0 +1,322 @@
...
+        /// <param name="entityType">Type of the entity.</param>
+        /// <param name="parentControl">The parent control.</param>
+        /// <returns></returns>
+        public override Control[] CreateChildControls( Type entityType, FilterField filterControl )
+        {
+            // Define Control: History Data View Picker
+            var ddlDataView = new DataViewPicker();
+            ddlDataView.ID = filterControl.GetChildControlInstanceName( _CtlDataView );
+            ddlDataView.Label = "Connected to Group Types";
+            ddlDataView.Help = "A Data View that provides the list of GroupTypes to which the Group may be connected.";
```
nairdo: Since this new filter is pretty straight-forward and is just filtering on matching GroupTypes ([line 312](https://github.com/SparkDevNetwork/Rock/pull/1933/files#diff-79adee19a7703fcf4547dee95db0fe82R312)), would you be open to changing (or allowing us to change):

1.  the `Label` to "Has a Group Type in this Data View" (similar to the [Group Member Data View filter's label](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Reporting/DataFilter/Person/GroupMemberDataViewFilter.cs#L242) ).
2. the `Help` to say "A Data View that provides the set of Group Types to match."

cbump:
I've made the changes requested.  The specific thing we needed this for is we use the Purpose flag on the Group Types to associate all the Serving Area group types together.  Then we want to fetch all active groups of those types and this was the way that seemed to make the most sense.

PR #1934, opened 2016-12-16
Add Financial Transaction Details Attribute UI to Modal

treyhendon:
# Context
Need to be able to set Financial Transaction Detail Attributes.

# Goal
Add UI to the Modal for Financial Transaction Details.

# Strategy
Used existing code in the FinancialTransactionDetails block to add UI in Details Modal.

# Possible Implications
No foreseeable impact to other areas.

# Screenshots
capture2

### DIFF
--- a/RockWeb/Blocks/Finance/TransactionDetail.ascx.cs
+++ b/RockWeb/Blocks/Finance/TransactionDetail.ascx.cs
@@ -230,7 +230,7 @@ namespace RockWeb.Blocks.Finance
             var txnDetail = new FinancialTransactionDetail();
             txnDetail.LoadAttributes();
             phAccountAttributeEdits.Controls.Clear();
-            Helper.AddEditControls( txnDetail, phAccountAttributeEdits, true );
+            Helper.AddEditControls( txnDetail, phAccountAttributeEdits, true, mdAccount.ValidationGroup );
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


RockWeb/Blocks/Finance/TransactionDetail.ascx.cs
```diff
@@ -1646,14 +1666,25 @@ private void ShowAccountDialog( Guid guid )
...
                 apAccount.SetValue( null );
                 tbAccountAmount.Text = string.Empty;
                 tbAccountSummary.Text = string.Empty;
+
+                txnDetail = new FinancialTransactionDetail();
+                txnDetail.LoadAttributes();
             }
 
+            phAccountAttributeEdits.Controls.Clear();
+            Helper.AddEditControls( txnDetail, phAccountAttributeEdits, true );
```
nairdo: FYI. Validation is not getting triggered for any "required" attributes because the validationGroup parameter is missing. It should probably get the validation group from the modal and should look like this:
`Helper.AddEditControls( txnDetail, phAccountAttributeEdits, true, mdAccount.ValidationGroup );`

I'll add that after I merge this request if you're unable to update it.

treyhendon:
Great catch!

RockWeb/Blocks/Finance/TransactionDetail.ascx.cs
treyhendon: Great catch!  Updated and pushed.

PR #1941, opened 2016-12-20
Added new person data filter for Postal Codes

trydyingtolive:
# Contributor Agreement
Yes

# Context
We often need to filter our dataviews by zipcode. We can create a location filter, and then filter by that dataview, but this doesn't work so well in a dynamic report.

# Goal
This will allow us to filter by zipcode.

# Strategy
I created a new postal code filter. It works with 5 digit, 9 digit, or alphanumeric postal codes.

# Possible Implications
None.

### DIFF
--- a/Rock/Reporting/DataFilter/Person/PostalCodeFilter.cs
+++ b/Rock/Reporting/DataFilter/Person/PostalCodeFilter.cs
@@ -1,0 +1,294 @@
+﻿// <copyright>
+// Copyright by the Spark Development Network
+//
+// Licensed under the Rock Community License (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.rockrms.com/license
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


edmistj:
+1

Rock/Reporting/DataFilter/Person/PostalCodeFilter.cs
```diff
@@ -277,12 +276,27 @@ public override Expression GetExpression( Type entityType, IService serviceInsta
...
+            IQueryable<Rock.Model.Person> qry;
+
             var groupMemberQry = groupLocationQry.Select( gl => gl.Group )
                 .Where( g => g.GroupType.Guid == familyGroupTypeGuid )
                 .SelectMany( g => g.Members );
 
-            var qry = new PersonService( ( RockContext ) serviceInstance.Context ).Queryable()
-                .Where( p => groupMemberQry.Any( xx => xx.PersonId == p.Id ) );
+            //Families which do not have locations need to be added separately
+            if ( comparisonType == ComparisonType.IsBlank || comparisonType == ComparisonType.DoesNotContain )
```
nairdo: It still not working for the _Not Equal To_ case, but I'll just fix this after I merge it in.

Issue #1441, opened 2016-03-18
[EN] Reducing page/block size by compressing ViewState


dcs619:
This may or may not turn into a PR...bear with me.  I've been watching our data performance since launch and noticed a 200-300ms increase in page load times as we've added new blocks.  We also have the opportunity to reach mobile users who may not live in a 4G area.

So would there any benefit to implementing ViewState compression?  Examples [here](http://www.codeproject.com/Articles/14733/ViewState-Compression), [here](http://stackoverflow.com/questions/1547532/how-to-minimize-viewstate-size-of-a-page-in-asp-net), and a full write-up of how ViewState works [here](http://weblogs.asp.net/infinitiesloop/Truly-Understanding-Viewstate).  The only drawback to compression seems to be server CPU cycles, which we (currently) barely touch.

It's entirely possible this approach is obsolete, as these articles are almost 10 years old.  @mikepetersonccv any thoughts?

edmistj:
We actually been thinking about using ViewState caching. For this the ViewState is stored on the server. Still much to be research on this but it could have a good impact on page loads.

mikepetersonccv:
When I looked into it last week, it sounds like the default https://msdn.microsoft.com/en-us/library/system.web.ui.hiddenfieldpagestatepersister(v=vs.110).aspx, which Rock currently uses, is already compressed, so that seemed like a dead end.  The alternative, https://msdn.microsoft.com/en-us/library/system.web.ui.sessionpagestatepersister(v=vs.110).aspx, reduced the hiddenfield ViewState size by 90% or more when I tried it on Rock.  

A couple of concerns with SessionPageStatePersister
1) What about load balancing?
2) What is the impact to Server Memory Usage.  For example, if 2000 sessions each had 250KB viewstates, that could 500MB of RAM.  
3) What about multiple tabs on a browser for the same session and page?

A few options we have
1) The option of HiddenField or Session ViewState page persister, is a "Per Page" option.  So, we could have it be an option on each RockPage and/or Site.
2) We could keep using the HiddenField method, but be more deliberate about how/if we use ViewState.  I'm sure there is some low-hanging fruit where we are using ViewState, but don't really need to be  (for example, Content Pages with no form entry)

dcs619:
Fixed in 1e8b10b84d2715a21f15e3758479f09c800395fc

PR #1947, opened 2016-12-29
Add support for ViewState compression with ratio of about 15:1.

cabal95:
# Contributor Agreement

Yes

# Context

Extremely slow postback of pages due to large ViewState size.

# Goal

Apply GZIP compression to the ViewState.

# Strategy

A new subclass of PageStatePersister has been added to the Rock.dll and RockPage class has been updated to override the property getter that retrieves it.

Compression of ViewState using GZIP algorithm is performed if the ViewState is greater than 100KB. The achieved compression ratio is between 15:1 and 20:1. A ViewState that was previously 1.5MB becomes 73KB.

# Possible Implications

Should not impact backwards compatibility beyond a user trying to re-post a page generated from before this patch is applied to a page generated after this patch is applied (unlikely, and would probably throw an exception either way). Security should not be affected as we still encrypt the ViewState with the Machine Key.

# Screenshots

screen shot 2016-12-28 at 12 40 36 pm

# Documentation

n/a

# Comments

This patch works as is, but a possible improvement would be to have a Global Attribute that the Admin could set for the threshold at which to compress ViewState (or disable it completely). That would, however, require migration scripts and I'm not really up to speed enough to tackle that just yet.

### DIFF
--- a/Rock/Rock.csproj
+++ b/Rock/Rock.csproj
@@ -1178,6 +1178,7 @@
     <Compile Include="Utility\DebugHelper.cs" />
     <Compile Include="Utility\ExcelHelper.cs" />
     <Compile Include="Utility\FileUtilities.cs" />
+    <Compile Include="Web\UI\CompressedHiddenFieldPageStatePersister.cs" />
     <Compile Include="Web\UI\RockTheme.cs" />
     <Compile Include="Web\Cache\AttributeValueCache.cs" />
     <Compile Include="Web\UI\Controls\WarningBlock.cs" />
--- a/Rock/Web/UI/CompressedHiddenFieldPageStatePersister.cs
+++ b/Rock/Web/UI/CompressedHiddenFieldPageStatePersister.cs
@@ -1,0 +1,131 @@
+﻿using System;
+using System.IO;
+using System.IO.Compression;
+using System.Web.Security;
+using System.Web.UI;
+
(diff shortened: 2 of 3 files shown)


cabal95:
I looked into providing this as a Plugin, but it would have requires the user to setup a Browser/ file with a PageAdapter setting.  And only one PageAdapter per matched browser can be specified (so documentation says) so that did not seem like a good option.

dcs619:
@cabal95 Praise be.  This would fix #1441.

mikepetersonccv:
@cabal95 this sounds like an awesome enhancement!! BTW, regarding Out Of Memory with IISExpress, you can configure Visual Studio to run the 64bit Version of IIS Express 
image

cbump:
Just as an FYI we merged this and have been running this in our production environment for about a month with no issues.

edmistj:
Great, I have a feature branch that also adds the ability to view the control structure of the viewstate (to help eliminate unneeded viewstate bulk) that I want to merge in with @cabal95 's compression. Was hoping to get to this today but ran out of time. I'll try to get it merged in the next week.

cabal95:
@cbump Thanks for taking the bullet to try this out in production!

edmistj:
@cabal95 I merged this with the changes I ha. You might code review it if you get a chance.

PR #1949, opened 2016-12-29
Add Campus filter and search box to Group Tree View settings panel

nateh777:
# Contributor Agreement
No

# Context
A church wanted the ability to filter the group tree down based on campus. As well as the ability to search from the group tree.

# Goal
A campus picker and search box will be added to the settings panel to give the user the ability to filter based on campus or run a search from the tree view.

# Strategy
I added the new campus picker (with a toggle to include "no campus set" entries) to the group tree settings panel, with the corresponding api and GroupService updates to support this new data to filter based off of. A group search box input-group was also added to the settings panel to be able to search for a specific group.

# Possible Implications

None that I could think of. New parameters were added as optional to the GetChildren function so it shouldn’t affect backwards compatibility at all.

# Screenshots
pasted image at 2016_12_29 02_54 pm

# Documentation
**Filter by Campus** shows only groups assigned to the selected campus.
**Include groups with no campus** is only used when filter by Campus is set and will do as it says.
**Search** allows you to run a search from the group viewer

### DIFF
--- a/Rock.Rest/Controllers/GroupsController.Partial.cs
+++ b/Rock.Rest/Controllers/GroupsController.Partial.cs
@@ -48,6 +48,8 @@ namespace Rock.Rest.Controllers
         /// <param name="excludedGroupTypeIds">The excluded group type ids.</param>
         /// <param name="includeInactiveGroups">if set to <c>true</c> [include inactive groups].</param>
         /// <param name="includeCounts">if set to <c>true</c> [include counts].</param>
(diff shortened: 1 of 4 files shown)


edmistj:
@mikepetersonccv be sure to add a documentation task for this

mikepetersonccv:
@nateh777 Looks good! However, we've started creating overloads of methods instead of adding default parameters.  There are some situations where plugin developers might get a compilation error if a method has additional default parameters added to it.  I'll take care of that :)

PR #1956, opened 2017-01-05
Performance improvement to GetByDevice and GetAllDescendents (sp)

dcs619:
# Contributor Agreement

Yes.

# Context

While profiling check-in queries during our Christmas services, I noticed the device and location lookups were some of the most expensive queries being run.  We have ~4k active locations assigned across 16 campuses (and ~180k total locations) so this is particularly problematic.

Under the original structure/execution plan, there was an index scan on Location: 

screen shot 2017-01-04 at 7 38 13 pm

Average execution time was ~5500ms using a cold cache or OPTION RECOMPILE.

# Goal

This PR seeks to optimize those device and location lookups.

# Strategy

Instead of requesting all location fields while traversing the Device or Descendant tree, I've limited the traversal to only include the Location.Id.  Once the traversal is finished and a much smaller result set is returned, a lookup fires to get the necessary Location information. 

With the new structure/execution plan, SQL Server chooses an index seek instead:  

screen shot 2017-01-04 at 7 54 45 pm

Average execution time is now ~150ms using a cold cache or OPTION RECOMPILE.

# Possible Implications

I've confirmed this change returns the same result set for Rock v6.0 running on SQL 2012 and SQL 2016,  as well as testing with Rock v6.1 from the Rock demo site.

### DIFF
--- a/Rock/Model/LocationService.Partial.cs
+++ b/Rock/Model/LocationService.Partial.cs
@@ -357,12 +357,13 @@ namespace Rock.Model
             return ExecuteQuery( string.Format(
                 @"
                 WITH CTE AS (
-                    SELECT * FROM [Location] WHERE [ParentLocationId]={0}
+                    SELECT Id FROM [Location] WHERE [ParentLocationId]={0}
                     UNION ALL
-                    SELECT [a].* FROM [Location] [a]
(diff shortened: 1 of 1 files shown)


azturner:
+1

PR #1964, opened 2017-01-09
Group Registration Status - Update the status of existing group members.

cbump:
# Contributor Agreement
_Have you filled out and sent your [Spark Contributor Agreement] (http://www.rockrms.com/Content/RockExternal/Misc/Contributor%20Agreement.pdf) to secretary [at] sparkdevnetwork.org?_ Yes

# Context
When a person registered for a group where they already existed in the group but were inactive, it left the person as inactive rather than updating their status per the settings in the Group Registration block.

# Goal
Group member's status will now be updated if the already exist in the group in the same role.

# Strategy
The block was already checking to see if the person existed in that role but it wasn't checking the status.  Now if the person exists but the status is different the person gets updated.

# Possible Implications
None.  I wouldn't think anyone would actually want this to not update the status if someone registers for the group.

### DIFF
--- a/RockWeb/Blocks/Groups/GroupRegistration.ascx.cs
+++ b/RockWeb/Blocks/Groups/GroupRegistration.ascx.cs
@@ -40,21 +40,20 @@ namespace RockWeb.Blocks.Groups
     [Category( "Groups" )]
     [Description( "Allows a person to register for a group." )]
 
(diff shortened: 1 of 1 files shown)


cbump:
I replaced this with [PR#1975](https://github.com/SparkDevNetwork/Rock/pull/1975)

PR #1968, opened 2017-01-11
Add the option to restrict Group Context by Campus Context

dcs619:
# Contributor Agreement

Yes.

# Context

When setting up dashboards to display data by user-selected contexts, the Group Context was pulling every group for a particular grouptype.  Since we have 16 campuses, the list of groups would get fairly large.

# Goal

This PR seeks to limit the list of Groups by Campus when a Campus Context is set.

# Strategy

I've added a Block Attribute for Admins to enable the Campus filter, assuming a Campus Context is set.  If a Campus Context is not set, the Group Context setter behaves as normal.

# Possible Implications

N/A.

### DIFF
--- a/RockWeb/Blocks/Core/GroupContextSetter.ascx.cs
+++ b/RockWeb/Blocks/Core/GroupContextSetter.ascx.cs
@@ -41,6 +41,7 @@ namespace RockWeb.Blocks.Core
     [TextField( "Clear Selection Text", "The text displayed when a group can be unselected. This will not display when the text is empty.", true, "", order: 3 )]
     [BooleanField( "Display Query Strings", "Select to always display query strings. Default behavior will only display the query string when it's passed to the page.", false, "", order: 4 )]
     [BooleanField( "Include GroupType Children", "Include all children of the grouptype selected", false, "", order: 5 )]
+    [BooleanField( "Respect Campus Context", "Filter groups by the Campus Context block if it exists", true, "", order: 6 )]
     public partial class GroupContextSetter : RockBlock
     {
         #region Base Control Methods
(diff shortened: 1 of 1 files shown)


RockWeb/Blocks/Core/GroupContextSetter.ascx.cs
```diff
@@ -41,6 +41,7 @@ namespace RockWeb.Blocks.Core
     [TextField( "Clear Selection Text", "The text displayed when a group can be unselected. This will not display when the text is empty.", true, "", order: 3 )]
     [BooleanField( "Display Query Strings", "Select to always display query strings. Default behavior will only display the query string when it's passed to the page.", false, "", order: 4 )]
     [BooleanField( "Include GroupType Children", "Include all children of the grouptype selected", false, "", order: 5 )]
+    [BooleanField( "Respect Campus Context", "Filter groups by the Campus Context block if it exists", true, "", order: 6 )]
```
nairdo: @dcs619 I'm going to change this to default to 'false' to match the current behavior.  

Also, since we will create a migration pretty much whenever a new attribute is added to a block, do you want to give me the GUID of this new block attribute from your production system?  Then I'll use it in the migration which should help prevent your currently set value from being toasted when the migration runs.

``` 
RockMigrationHelper.AddBlockTypeAttribute( "62F749F7-67DF-4A84-B7DD-84CA8E10E205", 
   "1EDAFDED-DFE6-4334-B019-6EECBA89E05A", "Respect Campus Context", "RespectCampusContext", "", 
   "Filter groups by the Campus Context block if it exists", 6, @"False", 
   "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX" );
```

dcs619: @nairdo Sounds great, thanks!  The GUID we have is `525A2D95-7ACC-4129-9805-DCA4E45A7C63`.

PR #1969, opened 2017-01-11
Add an API endpoint to check if a username is available

dcs619:
# Contributor Agreement

Yes.

# Context

When implementing our native app, we needed to be able to check if a username already existed.  This functionality exists inside Rock, but isn't exposed to the API.

# Goal

This PR adds an API endpoint to check for username availability.

# Strategy

See code.

# Possible Implications

N/A, provides ease of access.  Securing the endpoint could be a discussion.

### DIFF
--- a/Rock.Rest/Controllers/UserLoginsController.Partial.cs
+++ b/Rock.Rest/Controllers/UserLoginsController.Partial.cs
@@ -36,21 +36,10 @@ namespace Rock.Rest.Controllers
         /// <returns></returns>
         [HttpGet]
         [System.Web.Http.Route( "api/userlogins/available/{username}" )]
+        [System.Web.Http.Route( "api/userlogins/available" )]
         public bool Available( string username )
         {
             return ( (UserLoginService)Service ).GetByUserName( username ) == null;
-        }
-
-        /// <summary>
-        /// Tests if a username is available
-        /// </summary>
-        /// <param name="username"></param>
-        /// <returns></returns>
-        [HttpGet]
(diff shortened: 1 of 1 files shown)


nairdo:
@dcs619 If the intent is to be able to GET via `/api/userlogins/available?username=foo` I think it might be better to just add a second route to the existing Available(...) method:
```
        [System.Web.Http.Route( "api/userlogins/available/{username}" )]
        [System.Web.Http.Route( "api/userlogins/available" )]
```

What do you and @mikepetersonccv think about that?  I can't seem to find anywhere that says a controller method is limited to only one route. I tried it on my local environment and it works in all three cases:

- `/api/userlogins/available/foo`
- `/api/userlogins/available/?username=foo`
- `/api/userlogins/available?username=foo`

dcs619:
@nairdo 🤔 the intent was to GET `/api/userlogins/available/{username}` but it appears that's already enabled ([demo](http://rock.rocksolidchurchdemo.com/api/userlogins/available/github)).  Was that a recent addition?  

@jbaxleyiii I wonder if we made this change pre-v5 and it's in core now.

nairdo:
@dcs619 It looks like the original Available method is from a [Dec 18, 2012 commit](https://github.com/SparkDevNetwork/Rock/commit/8c419ac8835753e600d67ab32b8fb536421972f8) but since your method was called blahFromQueryString, I thought perhaps you were trying to use an alternate argument passing approach.  Try removing your new code and let me know if your API call is working as you expect.  If so you can just close this pull request.

PR #1978, opened 2017-01-15
SYS-3184 Fixing checkboxes not updating from a disabled state

edolyne:
# Context
When in the bulk update screen, selecting an input with checkboxes did not remove the disabled state.

# Goal
This pull request removes the disabled states from checkboxes in the formGroup, if they exist.

# Strategy
I've added to the JS onClick event to check if the formGroup contains checkboxes.  If it does, it removes the disabled state.

# Video Example of Issue
https://drive.google.com/file/d/0Bw8uldfEkvqrN25oZ1owMzZqdVE/view


nairdo:
@edolyne in your video example, what Field Type are you using to get checkboxes for those 'schedules'?  I've run into an issue with a seemingly related case that I'm in the process of addressing by slightly altering this new code -- and I don't want to break your use-case.

PR #2006, opened 2017-02-03
Fixed Grid Command Column header text to be vertically aligned

cabal95:
# Contributor Agreement

**Yes**

# Context

Vertical alignment for the `Register` column header in the Group Finder block was incorrect. Text was aligned too low and did not match other column headers. Found this was an issue with the core CSS Grid control in general for any command column header.

# Goal

Adjust the CSS to use correct padding.

# Strategy

The current CSS overrides the padding (and other styles) for both `thead` and `tbody` elements to make the buttons of a command column vertically aligned with the text. However, this makes the `thead` elements now have incorrect alignment with text in other column headers. CSS now overrides the `thead` padding style back to 12px to match other columns and leaves the `tbody` as it is currently.

# Possible Implications

Any grids that use text in command columns (I could not find any) that expect the text to be aligned that way will no longer look as intended.

# Screenshots

baseline

# Documentation

n/a

### DIFF
--- a/RockWeb/Styles/_grid.less
+++ b/RockWeb/Styles/_grid.less
@@ -361,6 +361,10 @@ td.grid-col-actions {
   }
 }
 
+.grid .table > thead > tr > th.grid-columncommand {
+  padding: 12px;
+}
+
 .table .color-field {
   width: 4px;
   padding: 0;


edmistj:
Thanks for this heads up @cabal95 ! I fixed it in a slightly different manner by removing a line of CSS via https://github.com/SparkDevNetwork/Rock/commit/374d6a3f132e931fccb0c9406c58180e456e8656

cabal95:
Hey @edmistj, Because that line you removed also affects the `tbody td` tags in addition to the `thead th` tags I don't think it will do what you were hoping, see screenshot:

change 374d6a3

edmistj:
Fixed that. I think the problem is the padding for the grid-columncommand was not specific enough (applied to both th and td) instead of being limited to just td's.

PR #2008, opened 2017-02-03
Adding some options for handling emails on RegistrationEntry

cbump:
# Contributor Agreement
Yes

# Context
We use the term household to describe everyone who lives in the same address.  "Family" and "Immediate Family" are too specific to relationship rather than living arrangement.  We also decided to just force everyone to update their record's email address if they change it during registration.

# Goal
Adding block attributes for:
- Specifying the term to use for family ("immediate family" is the default but we like "household" better).  This is used on the "Registrant is in the same household as Registrar" area.
- Force updating email addresses.

# Strategy
Just some block attributes and some changes in the block.

# Possible Implications
None.  The default settings don't change anything.

# Screenshots
image

# Documentation
I don't know that the Block Attributes for the RegistrationEvents block are documented anywhere. . .  But if they are, that's what needs updated.

### DIFF
--- a/RockWeb/Blocks/Event/RegistrationEntry.ascx.cs
+++ b/RockWeb/Blocks/Event/RegistrationEntry.ascx.cs
@@ -58,8 +58,8 @@ namespace RockWeb.Blocks.Event
     [BooleanField( "Enable Debug", "Display the merge fields that are available for lava ( Success Page ).", false, "", 5 )]
     [BooleanField( "Allow InLine Digital Signature Documents", "Should inline digital documents be allowed? This requires that the registration template is configured to display the document inline", true, "", 6, "SignInline" )]
     [SystemEmailField( "Confirm Account Template", "Confirm Account Email Template", false, Rock.SystemGuid.SystemEmail.SECURITY_CONFIRM_ACCOUNT, "", 7 )]
-	[TextField ( "Family Term", "The term to use for specifying which household or family a person is a member of.", true, "immediate family", "", 8)]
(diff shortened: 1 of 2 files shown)


edmistj:
:thumbsup:

RockWeb/Blocks/Event/RegistrationEntry.ascx.cs
```diff
@@ -58,6 +58,9 @@ namespace RockWeb.Blocks.Event
     [BooleanField( "Enable Debug", "Display the merge fields that are available for lava ( Success Page ).", false, "", 5 )]
     [BooleanField( "Allow InLine Digital Signature Documents", "Should inline digital documents be allowed? This requires that the registration template is configured to display the document inline", true, "", 6, "SignInline" )]
     [SystemEmailField( "Confirm Account Template", "Confirm Account Email Template", false, Rock.SystemGuid.SystemEmail.SECURITY_CONFIRM_ACCOUNT, "", 7 )]
+	[TextField ( "Family Term", "The term to use for specifying which household or family a person is a member of.", true, "immediate family", "", 8)]
+	[BooleanField( "Force Email Update", "Force the email to be updated on the person's record.", false, "", 9 )]
```
nairdo: I see this new setting is used to control the warning message on screen, but it it actually used somewhere to prevent/force the email on the person's record from being updated?

cbump: Line 1872 actually handles this automatically when the control isn't visible.  It's essentially the same path that is taken as if the person isn't logged in.

cbump:
@nairdo I've made all the changes you requested.  Thanks!

PR #2011, opened 2017-02-06
Enhance GroupDetailLava block to provide new limiting options

cabal95:
Enhanced the GroupDetailLava block of the Group Leader Toolbox to support:

- Block setting to hide the `Role` while editing or adding a Group Member.
- Block setting to hide the `Public` flag while editing the Group.

# Contributor Agreement

**Yes**

# Context

Our Small Group team wanted to prevent leaders from changing the role of an existing group member (for example, prevent the leader from making another member a leader without them having been vetted by our team).

Additionally they needed the Leader's to be able to mark their group as Public or Not Public themselves to be able to hide the group from the group locator.

# Goal

Add block settings and support code to add the needed functionality.

# Strategy

Two new block settings to turn on the new functionality.

# Possible Implications

Backwards compatibility is maintained by the default values keeping the UI looking like it did before this commit.

# Screenshots

sample1

sample2

# Documentation

n/a

### DIFF
--- a/RockWeb/Blocks/Groups/GroupDetailLava.ascx
+++ b/RockWeb/Blocks/Groups/GroupDetailLava.ascx
@@ -21,6 +21,7 @@
                 </div>
                 <div class="col-md-6">
                     <Rock:RockCheckBox ID="cbIsActive" runat="server" Text="Active" />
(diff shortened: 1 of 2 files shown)


edmistj:
👍

PR #2013, opened 2017-02-07
Updated Group Member Add From Url block to enhance security and provide role defaults

cabal95:
# Contributor Agreement

**Yes**

# Context

Potential security concern when using this block. The Person is passed as a GUID but the Group is passed as a integer ID number. Meaning the user can modify the URL to easily add themselves to other Rock Groups that have common numbers out of the box - such as the Admin security role.

Also (I know, I'm a bad boy for not doing two merges) added an additional default member role if none is specified as the GroupType's defined default member role (if it is defined).

# Goal

Add optional layer of security to limit the group types allowed to be used.

# Strategy

New Block Setting of `Limit Group Type` as a multi-select checkbox list that allows the Admin to optionally select which group types to limit the functionality to. For example if `Small Group` is selected then only groups of type Small Group will be allowed to be used.

# Possible Implications

On the security side, no backwards compatibility issues. Default is to allow all group types.

On the member role, very minimal impact if the block was not configured correctly or the URL parameters were not passed correctly then they might get the group type's default member role instead of an error.

# Screenshots

n/a

# Documentation

n/a

### DIFF
--- a/RockWeb/Blocks/Groups/GroupMemberAddFromUrl.ascx.cs
+++ b/RockWeb/Blocks/Groups/GroupMemberAddFromUrl.ascx.cs
@@ -48,6 +48,7 @@ namespace RockWeb.Blocks.Groups
 </div>" )]
     [BooleanField("Enable Debug", "Shows the Lava variables availabled for this block")]
     [EnumField("Group Member Status", "The status to use when adding a person to the group.", typeof(GroupMemberStatus), true, "Active")]
+    [GroupTypesField( "Limit Group Type", "To ensure that people cannot modify the URL and try adding themselves to standard Rock security groups with known Id numbers you can limit which Group Type that are considered valid during add.", false )]
     public partial class GroupMemberAddFromUrl : Rock.Web.UI.RockBlock
     {
         #region Fields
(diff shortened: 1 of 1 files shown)


nairdo:
@cabal95 Seeing as how this change is similar to [this](https://github.com/SparkDevNetwork/Rock/commit/6e313a969f44fd9b3d0c3344955806d73da3b609) and [this](https://github.com/SparkDevNetwork/Rock/commit/12bb8b8bc2bd65c8a55b0bfcf3b5e2d5804b1695) change with the GroupRegistration block, have you considered the possibility of adding the other feature (Enable Passing Group Id) to this block as well?

cabal95:
@nairdo Do you mean a way to disable providing a Group via the URL so it forces to only use the default group configured?  I can through that in. Shouldn't be too hard and could see that being a useful extension to self service kiosks.

nairdo:
@cabal95 Well, perhaps that _and_ the ability to enforce that the group's Guid be passed instead of by Id like the other block does with the Enable Passing Group Id.  I'm just thinking security feature parity between these two blocks is ideal since they are so closely related in base functionality.

cabal95:
Okay gotcha. I just glanced over the other two commits you mentioned I didn't dig in too deep. I'll see about adding another commit that brings parity closer to the other block.

cabal95:
@nairdo Second commit provided for supporting `GroupGuid` in the Query String as welll as block setting to disallow `GroupId`.

cabal95:
@nairdo Also, I'm still vague on the hotfix migrations, like the one in the commit you referenced and when those need to be included. So one may or may not need to be built for this change.

RockWeb/Blocks/Groups/GroupMemberAddFromUrl.ascx.cs
```diff
@@ -118,6 +129,20 @@ protected override void OnLoad( EventArgs e )
                     return;
                 }
 
+                // Validate the group type.
+                if ( !string.IsNullOrEmpty( GetAttributeValue( "LimitGroupType" ) ) )
+                {
+                    bool groupTypeMatch = GetAttributeValue( "LimitGroupType" )
+                        .Split( ',' )
+                        .Contains( group.GroupType.Guid.ToString(), StringComparer.CurrentCultureIgnoreCase );
+
```
azturner: @cabal95 This request looks good. Thanks! In future you could use something like this here: 
`GetAttributeValue( "LimitGroupType" ).SplitDelimitedValues().AsGuidList().Contains( group.GroupType.Guid );`

cabal95: Can do. I'll try to remember those two methods!

PR #2016, opened 2017-02-09
Improve responsiveness of kiosk buttons when tapping quickly

cabal95:
# Contributor Agreement

**Yes**

# Context

On touch devices (such as iPad) on kiosk themed sites it was difficult to tap buttons quickly as the presses were not registered - such as trying to quickly tap in one's phone number.

# Goal

Extend the pixel variance allowed from original press location to allow for more movement while still registering a click/tap.

# Strategy

Updated the fastclick.js to the latest version, which includes numerous other upstream fixes. The new version provides many new configuration options that can be set during initialization.

One such option is the amount of movement allowed. The default was 10px. In testing on my iPad, even the slightest movement of my finger resulted in a 15-25px registered movement. The initialization code has been updated to set this value to 40px.

@nairdo's changes from a few months ago of the double-tap prevention from 200ms down to 40ms have been preserved, as this is now a configurable option as well.

The 4 kiosk themes have been updated to use the new initialization options.

# Possible Implications

Because the movement amount was increased to 40px, an attempt to scroll by a very small amount may result in a click, such as when selecting a person from a long list of matches. If the scroll attempt is a "quick flick" style where the finger barely moves, it may register as a click on one of the people in addition to scrolling. Most people do not try to scroll a list this way so there should be little impact.

There is also a setting (Default is 700ms) that requires that a tap gesture be completed within that time period. Meaning if the user does a normal finger scroll but barely moves it will be unlikely to register as a tap since that gesture would likely take longer than 700ms.

# Screenshots

n/a

# Documentation

n/a

### DIFF
--- a/RockWeb/Scripts/fastclick.js
+++ b/RockWeb/Scripts/fastclick.js
@@ -1,639 +1,841 @@
-/**
- * @preserve FastClick: polyfill to remove click delays on browsers with touch UIs.
- *
- * @version 0.5.6
- * @codingstandard ftlabs-jsv2
- * @copyright The Financial Times Limited [All Rights Reserved]
- * @license MIT License (see LICENSE.txt)
- */
-
-/*jslint browser:true, node:true*/
-/*global define, Event, Node*/
-
-
-/**
- * Instantiate fast-clicking listeners on the specificed layer.
- *
- * @constructor
- * @param {Element} layer The layer to listen on
- */
-function FastClick(layer) {
+;(function () {
 	'use strict';
-	var oldOnClick, self = this;
+
+	/**
+	 * @preserve FastClick: polyfill to remove click delays on browsers with touch UIs.
+	 *
+	 * @codingstandard ftlabs-jsv2
+	 * @copyright The Financial Times Limited [All Rights Reserved]
+	 * @license MIT License (see LICENSE.txt)
+	 */
+
+	/*jslint browser:true, node:true*/
+	/*global define, Event, Node*/
 
 
 	/**
-	 * Whether a click is currently being tracked.
+	 * Instantiate fast-clicking listeners on the specified layer.
+	 *
+	 * @constructor
+	 * @param {Element} layer The layer to listen on
+	 * @param {Object} [options={}] The options to override the defaults
+	 */
+	function FastClick(layer, options) {
+		var oldOnClick;
+
+		options = options || {};
+
+		/**
+		 * Whether a click is currently being tracked.
+		 *
+		 * @type boolean
+		 */
+		this.trackingClick = false;
+
+
+		/**
+		 * Timestamp for when click tracking started.
+		 *
+		 * @type number
+		 */
+		this.trackingClickStart = 0;
+
+
+		/**
+		 * The element being tracked for a click.
+		 *
+		 * @type EventTarget
+		 */
+		this.targetElement = null;
+
+
+		/**
+		 * X-coordinate of touch start event.
+		 *
+		 * @type number
+		 */
+		this.touchStartX = 0;
+
+
+		/**
+		 * Y-coordinate of touch start event.
+		 *
+		 * @type number
+		 */
+		this.touchStartY = 0;
+
+
+		/**
+		 * ID of the last touch, retrieved from Touch.identifier.
+		 *
+		 * @type number
+		 */
+		this.lastTouchIdentifier = 0;
+
+
+		/**
+		 * Touchmove boundary, beyond which a click will be cancelled.
+		 *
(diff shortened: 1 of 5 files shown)


nairdo:
I've been doing multiple tests with this fix (using even an old, 5.1.1 original "iPad") and it looks great! The only caveat I've found is for anyone who may have created a custom check-in theme... they will need to also change their `Site.Master` and supply the recommended "`{ tapDelay: 40, touchBoundary: 40 }`" argument to the FastClick constructor because fastclick's default values are not very good.

nairdo:
I may have spoken too soon.  We attempted to roll this change out to our production environment and we ran into significant trouble using Android devices: in two out of three devices either the check-in button no longer responds or the keypad buttons.

cabal95:
@nairdo I'll check tomorrow when I'm in the office to see if we have any Android devices I can test with, but can you try testing it with the patch but without the change to Site.Master so the fastclick script runs with default values?

Also, do you have a way to tell what version of Chrome the Android devices are using?  If Chrome >= 32.x then fastclick shouldn't even be activating.

cabal95:
One more thing to try. Comment out the call in Site.Master to disable FastClick completely. It looks like it hasn't been needed on Android since 2013 and iOS since 9.3. I just tested with it disabled on my iOS 10 iPad and it seems to work perfectly fine*. So we might need to adjust this PR to only call FastClick on iOS < 9.3.

* The "Check in" button on the Welcome screen didn't work for a while, but it must have been a caching issue somehow as it started working once I bypassed it by going to /checkin/search and then went back to it.

https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away

nairdo:
@cabal95 Since I really want to move forward on this, for now, here at CentralAZ we've decided to create a new version of [our "mobile" Rock check-in theme](https://github.com/CentralAZ/Rock-CentralAZ/tree/central-develop/RockWeb/Themes/Checkin_CentralAZMobile) (used by people using their phones to check-in) that uses the old version of FastClick that we'll include in our theme's Assets folder:

```<script src="<%# ResolveRockUrl("~~/Assets/fastclickV0.5.6.js" ) %>" ></script>```

Since our mobile check-in device world is so vast**, I will also explore possibly removing FastClick completely from our mobile theme, but I don't really want to break check-in tomorrow morning and tomorrow night so for now we'll just include the old version.

> **NOTE: It might also be worth nothing that our mobile theme also has to deal with old iOS devices that open links in new Safari windows: https://github.com/CentralAZ/Rock-CentralAZ/blob/central-develop/RockWeb/Themes/Checkin_CentralAZMobile/Layouts/Site.Master#L15

cabal95:
Actually, scratch my last comment. Removing FastClick completely breaks all the buttons. Apparently I never saved the file when I tested removing it last time. My guess is the problem you experienced is because FastClick (correctly) didn't activate on your Android and, for whatever reason, nothing is working without FastClick. Lemme explore that some more as the current PR I think will break everything non-IOS. :(

cabal95:
Sorry I might have to redo this patch completely. IScroll by default does not pass on click/tap events to the scrolled content. Seriously. This is why it doesn't work without fastclick AND why it doesn't work with Android, because the new version of Fastclick says "hey I'm not needed on Android" and doesn't install it's own touch/click handlers.

You can turn on an option for it to simulate "click" events or "tap" events. With Tap events enabled and Fastclick also enabled it works. With Click events enabled and Fastclick NOT enabled it works. Click events + Fastclick doesn't work. Tap Events - Fastclick doesn't work. So it's a pretty ugly thing.

One fix is to drop support for pre-IOS 9.3, which accounts for about 8% of the iOS market. This would revert pre-IOS 9.3 behavior to the 300ms delay on tap. I don't really care for this method just yet, especially since churches (myself included) are likely to have a much higher percentage because we tend to leave check-in stuff on versions that work rather than upgrade to the latest.

The other fix is to revert to the original fastclick.js and put my changes in directly as hacks. I updated to get the latest bug fixes and the code supported those changes as options passed in, but it might be introducing too much headache. If I end up redoing this patch I'll close this PR and re-submit to keep the commits clean. I should know in another day or two.

cabal95:
Replaced with new PR: https://github.com/SparkDevNetwork/Rock/pull/2037

nairdo:
@cabal95 By the way -- we just rolled out this change here at CentralAZ (_minus our mobile theme as I mentioned above_) and David Anorve scored a perfect 9 out of 9 during kiosks testing.  On all 9 kiosks it never missed a single keypress!  That's like a small miracle. 

Thanks for your persistence in digging into this issue.

PR #2025, opened 2017-02-15
SaveAttributeValues() now loads and saves once

chrismcgrath:
# Contributor Agreement

Yes

# Context

We have about 200 person attributes. Calling SaveAttributeValues() takes a few seconds to run because it loops through all the attributes and loads their values one at a time, then saves each modified attribute one at a time.

# Goal

To optimize the process of saving all attributes.

# Strategy

Instead of calling SaveAttributeValue() for each attribute, instead load all the entity's attributes all at once, update all that need updating, then save once at the end.

# Possible Implications

# Screenshots

# Documentation

### DIFF
--- a/Rock/Attribute/Helper.cs
+++ b/Rock/Attribute/Helper.cs
@@ -800,13 +800,31 @@ namespace Rock.Attribute
         {
             if ( model != null && model.Attributes != null && model.AttributeValues != null )
             {
-                foreach ( var attribute in model.Attributes )
+                rockContext = rockContext ?? new RockContext();
+                var attributeValueService = new Model.AttributeValueService( rockContext );
+                var attributeValues = attributeValueService.GetByEntityId( model.Id ).ToDictionary(x => x.AttributeKey);
+                foreach (var attribute in model.Attributes.Values)
                 {
                     if ( model.AttributeValues.ContainsKey( attribute.Key ) )
                     {
-                        SaveAttributeValue( model, attribute.Value, model.AttributeValues[attribute.Key].Value, rockContext );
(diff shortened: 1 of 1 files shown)


chrismcgrath:
Found a bug, working on fixing it.

chrismcgrath:
OK, Found another issue. Current code is not taking into account the EntityTypeQualifierColumn and EntityTypeQualifierValue when pulling attributes for things like Workflow Actions.  It may be time to drop this if I don't find an easy solution to loading the proper attributes for any IHasAttributes object. Suggestions are welcome but feel free to reject this pull request if this isn't a good path.

nairdo:
@chrismcgrath I think it would be great if you figured out how to make this work with only one SaveChanges() call!  I don't have any suggestions at this point.  It could have a huge impact on various parts of Rock that call the SaveAttributeValues(...) method.  I'll close this for now, but please do resubmit if you figure it out.

chrismcgrath:
@nairdo maybe I'm confused but it looks like there is only one SaveChanges call to me, that was my intent. The save happens outside the for loop. Am I missing something?

Get Outlook for Android<https://aka.ms/ghei36>




On Fri, Mar 3, 2017 at 5:40 PM -0600, "Nick Airdo" <notifications@github.com<mailto:notifications@github.com>> wrote:


Closed #2025<https://github.com/SparkDevNetwork/Rock/pull/2025>.

chrismcgrath:
Nick saw my last comment and thought I hadn't figured this out, but I did a commit afterward that did get it working properly. Worth reviewing. Thanks!

PR #2027, opened 2017-02-15
Added SaveAttributeValue Extension methods

chrismcgrath:
# Contributor Agreement

Yes

# Context

Sometimes there is the need to only save one or more specific attributes rather than all of them, but there was no SaveAttributeValue method to go with the SaveAttributeValues method.

# Goal

To allow the saving of one or more specific attributes.

# Strategy

Created extension methods called SaveAttributeValue, one which takes a key and one which takes a collection of keys. 

# Possible Implications

# Screenshots

# Documentation


cabal95:
I think the first method should be SaveAttributeValues, plural, since it is taking a list of keys.

PR #2037, opened 2017-02-22
Redux: Improve responsiveness of kiosk buttons when tapping quickly

cabal95:
# Contributor Agreement
_Have you filled out and sent your [Spark Contributor Agreement](http://www.rockrms.com/Content/RockExternal/Misc/Contributor Agreement.pdf) to secretary [at] sparkdevnetwork.org?_

**Yes**

# Context

On touch devices (such as iPad) on kiosk themed sites it was difficult to tap buttons quickly as the presses were not registered - such as trying to quickly tap in one's phone number.

# Goal

Extend the pixel variance allowed from original press location to allow for more movement while still registering a click/tap.

# Strategy

Modified the fastclick.js file to change the default value for "allowed movement". The default was 10px. In testing on my iPad, even the slightest movement of my finger resulted in a 15-25px registered movement. The initialization code has been updated to set this value to 35px. (Down from 40px in my original pull request. There were only a few occasionally misses in the 35-40px range)

# Possible Implications

Because the movement amount was increased to 40px, an attempt to scroll by a very small amount may result in a click, such as when selecting a person from a long list of matches. If the scroll attempt is a "quick flick" style where the finger barely moves, it may register as a click on one of the people in addition to scrolling. Most people do not try to scroll a list this way so there should be little impact.

# Screenshots

n/a

# Documentation

n/a

### DIFF
--- a/RockWeb/Scripts/fastclick.js
+++ b/RockWeb/Scripts/fastclick.js
@@ -399,7 +399,7 @@ FastClick.prototype.touchHasMoved = function(event) {
 	'use strict';
 	var touch = event.changedTouches[0];
 
-	if (Math.abs(touch.pageX - this.touchStartX) > 10 || Math.abs(touch.pageY - this.touchStartY) > 10) {
+	if (Math.abs(touch.pageX - this.touchStartX) > 35 || Math.abs(touch.pageY - this.touchStartY) > 35) {
 		return true;
 	}
 


cabal95:
@nairdo Since you have the devices that didn't work with the last pull request, once check-in is no longer happening for you, could you try this change out instead and verify that it works?

If you have the last patch installed (just saw your e-mail that you do) the revert is basically put the Site.Master files back the way they were and replace your fastclick.js with this one and you should be synced up to this branch. All said and done this new change is one line.

Already tested on Physical iPad @ 10.2, Simulator iPad 9.3 and Simulator iPad 8.1 without any errors.

cabal95:
@nairdo We have been using this new modification for 6+ weeks without any issues. We are just rolling out actual check-in now but have been using a copy of the check-in UI phone-entry for volunteer signups and haven't had any responsiveness issues with this change.

azturner:
@nairdo (or @tcavaletto), were you able to confirm this change?

cabal95:
Closing. I think this is so old it will need to be redone and retested if it is ever worked on again.

PR #2038, opened 2017-02-22
Native Notifications

richarddubay:
**Native App Notifications in Rock**

PR Proposal for the addition of Notifications to the Rock Native App Communication Platform.

**Goal**: To bring native app notifications to Rock. 

**How This Will Work (Function)**: A new communication option will be added to the New Communication page called “Notifications”. This communication option will allow the person setting up the notification to choose between sending a notification to a certain set of devices(people), to groups of people, or to send to whole segments based on device type (iOS, Android, or All). This should also allow the creation of jobs that can send notifications to certain people based on an action (when a scheduled gift has been processed, for example).

To accomplish this goal, we will need to create:

A. A new table called PersonalDevice. Table schema includes:

DeviceRegistrationId
PersonAliasId
PersonalDeviceTypeId

**Note** One person id can have many device ids.

B. New defined types and values for device types.
Defined Type: PersonalDeviceType
Defined Values: Mobile Device (for starters - this can pave the way for things like "Laptop" in the future)
This defined type could potentially have an attribute on it called "Supports Notifications" or "Notifications Enabled" that would essentially act as the opt-in/out for sending notifications to that device type. 

** Note** I think that _maybe_ adding or not adding a device to the Personal Device table could act as a way of opting-in or out of notifications that that particular device? Discussion welcome here.

C. A new medium called “Push Notifications”.

D. A new transport called “Firebase”. This transport will use Firebase to send any and all notifications just like Twillio does for SMS messaging. 

_About Firebase_
Firebase is a free service offered by Google (https://firebase.google.com/). It manages all the devices that you have attached to your account. Devices can be subscribed to “topics” that allow you to send notifications to only certain groups of people (like a home group) or to whole segments of users based on device (iOS or Android). You can even get as granular as just sending notifications to one device if so desired. By using Firebase, we circumvent the need to have multiple transports for one medium. This also addresses the desire to keep the transports based on a delivery method as opposed to being based on an operating system.

_About Permissions_
When a user first opens the app after updating to the new version, they will get a pop-up asking if they want to allow notifications. This is their chance to opt-in. If they do, their device would send a Registration Id back to Rock to keep in the database.

_Interactions_
I'm not sure that interactions will be much help here. Unless we can track when a user taps on a notification, there will be nothing we can track. And I'm not sure at this point that we can do that. So potentially, we wouldn't need to do anything here. Thoughts?


richarddubay:
@edmistj @azturner 

This is the start of the PR for native notifications and a chance for us to discuss what we'd like to do. We'll probably need to chat about this soon, but I wanted to get the proposal out here so we can start these discussions.

richarddubay:
@edmistj I created this gist with the migration needed for adding the PersonalDevice table and the PersonalDeviceType DefinedType and Mobile Device DefinedValue. It runs for me, but I wanted to make sure I wasn't missing anything. When you get a moment, can you take a look at this and let me know if I'm forgetting anything? It's been a minute since I've created a migration. Thank you!

https://gist.github.com/richarddubay/2ed342b81641fb61617662d4191b7cde

richarddubay:
I added the migration to this commit, but I can remove it before we do anything with this branch. I just wanted it here for tracking.

edmistj:
@richarddubay Did you see my comments about the Notification attributes and properties?

richarddubay:
@edmistj Just now I did. I'll update the migration with what you suggested. I forgot about those properties yesterday when I was doing this. I'll let you know when it's updated. Thanks!

jbaxleyiii:
@azturner @edmistj @nairdo this should be ready for review. We also wrote docs for the website about it

Rock.Client/CodeGenerated/SystemGuid/RockSystemGuids.cs
```diff
@@ -257,6 +260,10 @@ public class EntityType
         public const string CONTENT_CHANNEL = "44484685-477E-4668-89A6-84F29739EB68";
         public const string CONTENT_CHANNEL_ITEM = "BF12AE64-21FB-433B-A8A4-E40E8C426DDA";
         public const string DATAVIEW = "57F8FA29-DCF1-4F74-8553-87E90F234139";
+        public const string FINANCIAL_ACCOUNT = "798BCE48-6AA7-4983-9214-F9BCEFB4521D";
+        public const string FINANCIAL_BATCH = "BDD09C8E-2C52-4D08-9062-BE7D52D190C2";
+        public const string FINANCIAL_TRANSACTION = "2C1CB26B-AB22-42D0-8164-AEDEE0DAE667";
+        public const string FINANCIAL_TRANSACTION_DETAIL = "AC4AC28B-8E7E-4D7E-85DB-DFFB4F3ADCCE";
```
jbaxleyiii: @edmistj  these came up when I ran the codegen for the other guids. I'm not sure why?

Rock/Communication/Transport/Firebase.cs
```diff
@@ -0,0 +1,244 @@
...
+                    var historyService = new HistoryService( loopContext );
+                    if ( recipient != null )
+                    {
+                        try
+                        {
+                            var service = new PersonalDeviceService( rockContext );
+                            int personAlias = recipient.PersonAliasId;
+
+                            List<string> devices = service.Queryable()
+                                .Where(p => p.PersonAliasId == personAlias && p.NotificationsEnabled)
```
azturner: Not a big deal, but if you are changing other things, this same `.Where` clause is used in several places. You might want to consider creating a service partial class and a method for this. i.e. 
```
        public IQueryable<PersonalDevice> GetByPersonAliasId( int personAliasId )
        {
            return Queryable().Where( p => p.PersonAliasId == personAliasId && p.NotificationsEnabled );
        }
```

Rock/Communication/Transport/Firebase.cs
```diff
@@ -0,0 +1,244 @@
...
+
+                            if ( devices != null )
+                            {
+                                // Create merge field dictionary
+                                var mergeObjects = recipient.CommunicationMergeValues( mergeFields );
+                                
+                                var resolvedMessage = message.ResolveMergeFields( mergeObjects, communication.EnabledLavaCommands );
+                                var notification = new Message
+                                {
+                                    RegistrationIds = devices,
```
azturner: If two people with the same device id are merged, you would have duplicate device id's in this list. Would that cause a problem with Firebase? You might consider removing duplicates when people are merged, or at very least modifying your select above to select distinct device ids.

Rock/Workflow/Action/Communications/SendPushNotification.cs
```diff
@@ -0,0 +1,235 @@
+﻿// <copyright>
```
richarddubay: Nope! I removed it.

Rock/Model/PersonalDevice.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// Configuration class.
+    /// </summary>
+    public partial class PersonalDeviceConfiguration : EntityTypeConfiguration<PersonalDevice>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PersonalDeviceConfiguration"/> class.
+        /// </summary>
+        public PersonalDeviceConfiguration()
+        {
+            //this.HasOptional( r => r.PersonAlias ).WithMany().HasForeignKey( r => r.PersonAliasId ).WillCascadeOnDelete( false );
```
richarddubay: Me neither. Uncommented.

Rock/Communication/Transport/Firebase.cs
richarddubay: Modifying to use` devices.Distinct().ToList()` here to get distinct devices. I don't feel great about messing around with the merge people code just yet. :)

Rock/Communication/Transport/Firebase.cs
richarddubay: I couldn't find where we were using this `.Where` clause in other places. If you know of them, please let me know. I'd be happy to do what you're suggesting here if it's used a lot. I just couldn't find it.

Rock/Communication/Transport/Firebase.cs
```diff
@@ -0,0 +1,244 @@
...
+
+namespace Rock.Communication.Transport
+{
+    /// <summary>
+    /// Communication transport for sending push notifications using Firebase
+    /// </summary>
+    [Description( "Sends a communication through Firebase API" )]
+    [Export( typeof( TransportComponent ) )]
+    [ExportMetadata( "ComponentName", "Firebase" )]
+    [TextField( "SenderId", "Your Firebase account sender Id (find at https://console.firebase.google.com/project/<YOUR_APP>/settings/cloudmessaging)", true, "", "", 0 )]
```
richarddubay: I removed SenderId. No sense in having something there that isn't being used. Great catch! :100:

richarddubay:
> I didn't see any blocks/migrations for managing a person's devices. I imagine new records are created by app, but what about deleting existing devices and/or updating notification flag.

@azturner I talked over with @jbaxleyiii about the blocks/migrations for removing devices. We think this is probably out of scope for right now. This should get native notifications up and running. We can come back and add the ability to delete a device later. 

@jbaxleyiii Please add your thoughts if I misspoke. :)

PR #2039, opened 2017-02-23
Added a minimum age option to new accounts

trydyingtolive:
# Contributor Agreement
Yes

# Context
We cannot allow children under 13 to create an account.

# Goal
This will allow us to set a minimum age for creating a new account.

# Strategy
I used the existing birthday field to see if the person is old enough to create an account. There is a new block attribute to set the minimum age in years. If the person does not meet the required age, a warning is shown.

# Possible Implications
The default minimum age is 0 so no one should be affected.


edmistj:
👍 (Internal Note) I wonder if the default should be 13 as a best practice?

dcs619:
For context, [COPPA](https://en.wikipedia.org/wiki/Children's_Online_Privacy_Protection_Act) is the federal regulation that specifies this (and hands out fines for non-compliance).

tom-powers:
David, I didn't realize the fine was so steep ($16K per infraction).  I think that defaulting to 13 might be wise indeed.

trydyingtolive:
In light of this conversation, I have changed the default to 13 and added a warning to the help text.

PR #2042, opened 2017-02-28
Workflow action to test if person is in a dataview

trydyingtolive:
# Contributor Agreement
Yes

# Context
We have a dataview which is our standard for who can work with children. When someone makes a requests in a connection, we would like to send an email to everyone who not completed the process so they can begin the application to serve. We could use a SQL action, but we would like to have one source for truth. 

# Goal
This action would allow people to use a dataview to set a boolean attribute in their workflow based on if a person is in that dataview. 

# Strategy
I created a workflow action that checks to see if the person is in a particular dataview and sets a workflow attribute "True" or "False" accordingly. That attribute can then be used as a filter to start or end actions and activities. 

# Possible Implications
None

# Documentation
Person: The person who will be tested against the dataview.
DataView: The dataview to test the person against. 
Timeout: This is a safeguard against your dataview taking too long to run. Default is 30 seconds.
Boolean: Boolean attribute to set based on if the person is found in the dataview. Note: when filtering against a boolean attribute, test to see if the value equals "True" or "False"

### DIFF
--- a/Rock/Workflow/Action/People/PersonInDataView.cs
+++ b/Rock/Workflow/Action/People/PersonInDataView.cs
@@ -1,0 +1,146 @@
+﻿// <copyright>
+// Copyright by the Spark Development Network
+//
+// Licensed under the Rock Community License (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.rockrms.com/license
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+//
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


edmistj:
@nairdo please update the workflow actions content channel for documentation

Rock/Workflow/Action/People/PersonInDataView.cs
```diff
@@ -0,0 +1,146 @@
...
+{
+    /// <summary>
+    /// Sets boolean based on if person is in dataview.
+    /// </summary>
+    [ActionCategory( "People" )]
+    [Description( "Sets an attribute True or False depending on if the person is in selected dataview." )]
+    [Export( typeof( ActionComponent ) )]
+    [ExportMetadata( "ComponentName", "Person In DataView" )]
+
+    [WorkflowAttribute( "Person", "Workflow attribute that contains the person to add the note to.", true, "", "", 0, null, new string[] { "Rock.Field.Types.PersonFieldType" } )]
```
nairdo: I believe the description ("...to add the note to.") is incorrect here.

PR #2043, opened 2017-02-28
Update RegistrationInstanceDetail.ascx

jreciomoreno:
The grid "gGroupPlacements does not display's properly in the Group Placement Panel when the "EnableResponsiveTable" is set to false, setting it to true fixes the issue.
Also it is helpful to allow sorting on the grid by setting to true the "AllowSorting" setting.

# Contributor Agreement

# Context

The group placement grid doesn't show properly when there are several columns on the grid. Changing the EnableResponsiveTable property to true fixes the issue.

# Goal

Will fix how the group placement grid shows on the screen.

# Strategy

Yes.

# Possible Implications

No

# Screenshots

Before
image

After
image

# Documentation

### DIFF
--- a/RockWeb/Blocks/Event/RegistrationInstanceDetail.ascx
+++ b/RockWeb/Blocks/Event/RegistrationInstanceDetail.ascx
@@ -377,7 +377,7 @@
                             </div>
                         </div>
                         <Rock:ModalAlert ID="mdGroupPlacementGridWarning" runat="server" />
-                        <Rock:Grid ID="gGroupPlacements" runat="server" DisplayType="Light" EnableResponsiveTable="false" AllowSorting="false" RowItemText="Registrant" ExportSource="ColumnOutput">
+                        <Rock:Grid ID="gGroupPlacements" runat="server" DisplayType="Light" EnableResponsiveTable="true" AllowSorting="true" RowItemText="Registrant" ExportSource="ColumnOutput">
                             <Columns>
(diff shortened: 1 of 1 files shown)


mikepetersonccv:
@edmistj It looks like EnableResponsiveTable causes Pickers to not show correctly when they are inside a grid.  Here is what it looks with EnableResponsiveTable is set to False 

image

But here is what happens when it is set to true 
image

edmistj:
Looks like in Chrome this works by expanding the grid to fit the picker. @mikepetersonccv can you look at what it would take to get this working in IE?

azturner:
@MikeDPeterson , was there an update to Jon's last comment about getting ResponsiveTable working in IE (noticed you unassigned yourself same day :))? Just wondering if this has been resolved without needing this change.

MikeDPeterson:
@azturner @edmistj if I remember correctly, this was due to a CSS issue. Something about the Overflow classes being different when the Responsive Table classes were added.  If you want, I can spend some time getting to the bottom of it

edmistj:
@garrettjohnson can you review this PR and see what the best course of action might be.

edmistj:
➤ Garrett Johnson commented:

[Jon Edmiston](https://app.asana.com/0/20866866924296/list) I think we talked about this in a 1:1, apologies for not closing this. Basically his solution is not a fix. We need to look further upstream at how pickers are architected, when we talked you said we'd need to wait beyond V8 to implement something like that

edmistj:
Going to close this PR as it fixes one issue, but inadvertently introduced another. Will have to circle back to find a different solution.

PR #2046, opened 2017-03-02
Update SignatureDocumentTemplateService.Partial.cs

jreciomoreno:
The objects used to set the Signature Document properties that is being saved are currently used in two different contexts this causes the code execution to throw an exception.

By creating new objects based on the values of the objects being passed as a parameter into the function resolves the issue.

# Context
Clicking the "Send Signature Request" button raised Invalid Operation Exceptions because an object was being used in two different contexts.

# Screenshots

"Send Signature Request" button on Registration Detail block.
image

### DIFF
--- a/Rock/Model/SignatureDocumentTemplateService.Partial.cs
+++ b/Rock/Model/SignatureDocumentTemplateService.Partial.cs
@@ -150,13 +150,16 @@ namespace Rock.Model
 
                         if ( document == null )
                         {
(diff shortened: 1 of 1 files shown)


azturner:
@jreciomoreno I believe this other pull request: https://github.com/SparkDevNetwork/Rock/pull/2178/files fixes same issue so closing this pull request. Feel free to submit again if that is not the case.

PR #2048, opened 2017-03-06
switch to newer async twilio client

jbaxleyiii:
# Context
During a sunday service, we did a big push for a campaign backed by text to workflow. We had over 6000 text in which at peak, locked up our rock instance and dropped communication from over half. The waiting on twilio response caused our instance to spike with CPU and timeout on all requests
 
# Goal
This makes the waiting for the POST to twilio async to not block the main thread.

# Strategy
After updating the twilio client, we are able to use the `CreateAsyncMessage`. This changes the `Send` method to be an async method.

# Possible Implications
There shouldn't be any

### DIFF
--- a/Rock/Communication/Transport/Twilio.cs
+++ b/Rock/Communication/Transport/Twilio.cs
@@ -30,6 +30,8 @@ using Rock.Model;
 using Rock.Web.Cache;
 
 using Twilio;
+using TwilioTypes = Twilio.Types;
+using Twilio.Rest.Api.V2010.Account;
 
 namespace Rock.Communication.Transport
 {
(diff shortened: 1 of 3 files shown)


cabal95:
I'm not that familiar with the async keyword, but is it safe/proper to add the async keyword to a method that is an override from the base class?

I'm a bit concerned that the new API appears to have no way to run with multiple accounts as it seems to do everything with static methods. Sadly their documentation for the new API is non-existent. Out of the box I don't think this will be an issue, but it could be moving forward if a plug-in provides it's own Twilio connectivity, a single Init call could disable core SMS support.

Mind you I don't have a better solution seeing as the old API (the one currently in use) will be deprecated in a few months anyway.

cabal95:
Actually it looks like all the TwilioClient.Init() method does is setup a shared TwilioRestClient that is used in all the other static methods if none is provides. So this could be updated to create a TwilioRestClient with the account information and then pass that into the CreateAsyncMessage method: https://github.com/twilio/twilio-csharp/blob/master/src/Twilio/Rest/Api/V2010/Account/MessageResource.cs#L78

edmistj:
Successfully compiled locally (appveyor failed it) and James has tested it on their servers. Pulling it in.

PR #2053, opened 2017-03-07
Make twilio webhook async and allow for future fully async transports

jbaxleyiii:
# Context
Even though the sending is now async, the webhook was not, so we still saw large load spikes

# Goal
To make the webhook for twilio faster

# Strategy
IHTTPASYNC YAY

Changed the IHttpHandler for Twilio to be async

### DIFF
--- a/Rock/Communication/MediumComponent.cs
+++ b/Rock/Communication/MediumComponent.cs
@@ -107,14 +107,8 @@ namespace Rock.Communication
             var transport = Transport;
             if ( transport != null && transport.IsActive )
             {
-                if (transport.IsAsync)
-                {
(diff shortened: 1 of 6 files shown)


edmistj:
@jbaxleyiii if it helps we're planning on making all of the webhooks async like the text to workflow plugin change.

jbaxleyiii:
@edmistj this is the code we are using in production now. It sent around 10,300 in 45 mins with no duplicates.

edmistj:
@jbaxleyiii What should we do with this PR. We're trying to knock-out PRs tomorrow.

jbaxleyiii:
Should be great to merge! It's what we are using in production

PR #2061, opened 2017-03-11
React

jbaxleyiii:
Context
In order to migrate the native app (and web app) platform that NewSpring is currently developing, we need support for rendering React components as Rock blocks. This will open up a lot of potential performance improvements, as well as pave the way for a universal app to be part of Rock core.

# Goal
Shared code between web and app through react. 

# Strategy
Using webpack as a bundler, we have created independent bundles for each block to be lazy evaluated depending on which block is loaded. There are still some questions with this listed below.

# Possible Implications
Currently the react block controller makes an api call to `/api/Pages`. If this isn't available for security reasons, then the block won't load. To get around this, we could include block paths on a page as part of the Rock.settings client side js initialization?

This could impact the building process of Rock. There are a couple of options here.

1. Commit the production bundle files (just like DLLs in this case) and when react dev is done, ensure `npm run build` is done prior to commiting
2. Require node as a build dependency for Rock
3. Find a way to build using msbuild (there is an old c# webpack project that may be a starting point?)

This PR creates an example counter block (under Blocks/Examples) to showcase the current integration.

Currently supported:
- No changes to Rocks blocks system
- Interopt between react and webform blocks
- Server side rendering
- Setting initial props of the react block from the server
- hot module reloading in development mode
- split bundles based on block needs
- testing framework included
- linting and static typing included
- server side rendering

Next Steps (integration wise)
- GraphQL support (along with a GraphQL endpoint for rock 🔥)
- Theme support
- Inter block communication
- Client side routing)
- React only theme

cc @edmistj @azturner


jbaxleyiii:
@nairdo I was unable to reduce the packages any? I'm not sure why, but they all seem needed? Also, I have been unable to fix the line endings on the site.master. Could you see if you could fix it? Even copy pasting from develop on github didn't fix it.

Also, I think this is a good first take prior to the Dynamic block additions. They serve two different application types

edmistj:
Ready to pull in after review.

mikepetersonccv:
@jbaxleyiii We started getting javascript errors in IE with this change.  It ended up being an issue in IE and only when DocumentMode was set to IE10.  To fix the issue I committed https://github.com/SparkDevNetwork/Rock/commit/72e5c9c7389ee6fb50897d8650bb0d1d9beeae3a.  This changes Rock to no longer specify X-UA-Compatible, and it sounds like that mode is now deprecated anyways.  The bad news is that any custom themes that don't ship with rock will have to manually remove X-UA-Compatible (see the commit to see what I mean).  It might be nice if you could look into a way to make it work for themes that still have IE10 mode specified.

PR #2064, opened 2017-03-13
New Lava Property field options to Group and Person reports.

cabal95:
# Contributor Agreement

**Yes**

# Context

Inability to display City of a Group's Meeting Location without jumping through various formatting hoops and multiple report columns.

# Goal

Provide a new report field type that allows more flexibility when using the Lava formatter.

# Strategy

* Implemented new LavaBoundField control to the Rock.Web.UI.Controls.Grid namespace that allows for the use of Lava to format the specific field value. Extra merge values are also supported (such as providing the standard merge options).
* Implemented new abstract LavaPropertySelect class in Rock.Reporting.DataSelect namespace that provides the UI and logic to the user. Reflection is used to identify all the available properties on the entity type that the user can select from. Also allows for selecting the sorting property to use when the selected property is a linked Entity.
  * Example: When reporting on a Group you can select the property `Schedule` and then the sorting-property of `WeeklyTimeOfDay` to allow for sorting of Groups based on the time of day of their schedules.
* Implemented new concrete GroupLavaPropertySelect class in the Rock.Reporting.DataSelect.Group namespace.
* Implemented new concrete PersonLavaPropertySelect class in the Rock.Reporting.DataSelect.Person namespace.

# Tests

n/a

# Possible Implications

Should be no impact on backwards compatibility, all new code.

# Screenshots

2017-03-13 10_00_04-virtualbox

2017-03-13 10_02_11-virtualbox

2017-03-13 10_04_14-virtualbox

2017-03-13 10_05_40-virtualbox

Lava Property fields are available as Sort fields since there is a concrete database field behind it. Sorting is ignored if the property selected is an IEnumerable, which cannot be sorted.

2017-03-13 09_59_41-virtualbox

2017-03-13 09_56_07-virtualbox

# Documentation

I do not believe any specific documentation needs to be created as I have not seen any place where all the reporting field types are documented.


cabal95:
Changing this to a work in progress. I need to dig into the Grid internals a bit. For some reason the GetExportValue of my BoundField is not being called which causes the exported data to be incorrect during Excel export.

cabal95:
I seem to have terrible luck with Appveyor and those iCal unit tests - but the latest commit fixes the Excel export. I implemented the same concept the CallbackField uses to allow the BoundField to format the excel output data. This could probably be moved to an Interface declaration at some point in the future to make it easier for other DataBounds to do the same without having to modify the Grid control.

Ready to go again.

edmistj:
I'm not seeing the use case for this. We do have a Lava field type. I believe it's limited to seeing only columns on the report but could you add the column as your property and then hide it? Or could the concept of the "Property" be added to the current field as an "Additional Property"?

cabal95:
It's possible this could be merged with the existing Lava item, but I'm not sure. This one uses a subclass of RockBoundField since it is working with real database fields while the existing Lava column uses a subclass of RockTemplateField. They receive very different data passed to them when they format for display and export.

The existing deficiency is that the report system only makes a relatively small percentage of properties available on the various entities we report on.

For my example above, currently there is no way to include the Schedule property/column on a Group report because it just isn't available to be selected anywhere. This component makes available _all_ properties of _any_ entity without any extra coding(see note) by using reflection.

It also allows the user to be able to sort by those properties which they may not be able to do now even if they include the property and use Lava to format it. Example of this is that for a report we needed right away I had to write a Group Schedule DataSelect for reporting so that I could include the Group's schedule in the report. I could sort by `Schedule` but that really does me no good (actually I think I tried and it might have thrown an exception). But I _can_ sort on `Schedule.WeeklyTimeOfDay` with this component so that the report shows everything sorted by start time.

Note: Because of the way data selects are designed a 3-line class needs to be defined for each entity type because a data select is a singleton rather than created as needed. As a singleton it is not passed the entity type it is working with in the required methods.

cabal95:
Updated component names to `Group Formatted Field` and `Person Formatted Field` to reduce confusion with existing component names. Per offline conversation with @edmistj.

PR #2065, opened 2017-03-13
New Lava filter ResponseWrite for sending raw text to client

cabal95:
# Contributor Agreement

**Yes**

# Context

Somebody was trying to find a way to use Lava to send a response to an AJAX query but could not do so because there would always be other content on the page, even with a blank template. Seemed like a useful feature.

# Goal

New filter to send a raw response to the http client.

# Strategy

New Lava filter that takes a string value as it's input with an optional parameter for the `Content-Type` header.

Sample code snipped that returns a JSON encoded string to the client listing the names of all the security groups the currently logged in person is a member of.

```
{% capture json %}
[{%- assign groups = CurrentPerson | Groups:'1','Active' -%}
{%- for member in groups -%}{{- member.Group.Name | ToJSON -}}{% if forloop.last == false %},{% endif %}{%- endfor -%}
]{%- endcapture %}
{{ json | ResponseWrite:'application/json' }}
```

Using this filter Admins can provide simple API systems for their pages that would allow for asynchronous AJAX responses without having to write C# code or rely on the full API system.

To help prevent the admin from locking themselves out of the page, there are two checks done that prevent the response being written and instead write it as output from the Lava filter.
1. Postback - On postback the filter writes the content as HTML encoded text. This check is done so that after clicking Save when editing a block the admin is not instantly kicked out of the full page.
2. ?ResponseWrite=0 - URL Parameter ResponseWrite=0 disables the filter as well.

# Tests

Test not provided, filter requires valid HttpContext to function properly.

# Possible Implications

None.

# Screenshots
_Provide us some screenshots if your pull request either alters existing UI or provides new UI. Arrows and 
labels are helpful._

None

# Documentation

Needs documentation in the Lava docs.

> Sends the text to the client and immediately ends the web request. This can be used if you need
> to send raw data to the client as a response to an AJAX request.

### DIFF
--- a/Rock/Lava/RockFilters.cs
+++ b/Rock/Lava/RockFilters.cs
@@ -693,6 +693,55 @@ namespace Rock.Lava
             return input.Substring(start, length);
         }
 
+        /// <summary>
+        /// Clears the response buffer and writes the input to the response stream then ends
+        /// the request.
+        /// </summary>
+        /// <param name="input">The input value to be sent to the client in raw form.</param>
+        /// <param name="contentType">The Content-Type value to be used if provided.</param>
+        /// <returns>A formatted string to display if the response write did not happen.</returns>
+        public static object ResponseWrite( object input, string contentType = null )
+        {
+            if ( input == null )
+            {
+                return null;
+            }
+
+            var context = HttpContext.Current;
+            if ( context == null )
+            {
+                return null;
+            }
+
+            var responseWrite = context.Request.QueryString["ResponseWrite"];
+            var rockPage = context.Handler as RockPage;
+
+            /* Allow the filter to be disabled by the query string for disaster recover. */
+            if ( !string.IsNullOrEmpty( responseWrite ) && responseWrite.AsBoolean() == false )
+            {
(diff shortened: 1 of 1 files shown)


edmistj:
Would it be cleaner to implement this like the ~/AppCode/GetChannelFeed.cs where we have a defined type that allows you to define a lava template?  Loading up a full page and blocks only to return something other than a page seems odd. There's also the Lava REST endpoint https://www.rockrms.com/page/751

Thoughts?

cabal95:
No arguments that it's not exactly the most efficient way to go about it. Putting this on a page with 14 blocks on it is probably a terrible idea, but the same thing also happens every time a postback event happens. So whether you are using a postback event to populate data or using this filter the entire page loads on the server either way.

There are times when using the API and/or a web handler is not readily feasible. Example is when the raw data we want to send back is directly related to the variables passed into the lava by the block. It may already be doing filtering on those objects for us that would be troublesome to try and replicate and keep in sync. And sometimes it's data that just isn't available another way. i.e. some data passed to Lava is a private/custom object type defined by the block itself.

Having said all that, I'm not opposed to making this a plug-in instead or something instead of core if that would be preferred. This was based off something another person was trying to accomplish but couldn't and I thought "that's an interesting idea, I don't know of any other way to accomplish that" so wrote this up.

PR #2066, opened 2017-03-14
Changed UI to support inactivating check-in groups

trydyingtolive:
# Contributor Agreement
- [x] _Have you filled out and sent your [Spark Contributor Agreement](http://www.rockrms.com/Content/RockExternal/Misc/Contributor%20Agreement.pdf) to secretary [at] sparkdevnetwork.org?_

# Context
Occasionally we need to retire check-in groups and to reduce clutter we would like to inactivate them and have them not clutter up the check-in interface. If we choose to delete a group all the attendance records are essentially lost.

# Goal
My goal is to make is easy to inactivate and reactivate check-in groups.

# Strategy
I modified the configuration controls and blocks to check to see if the group has attendance records before deleting. If the group does not have attendance records it is simply deleted. However, if it does have attendance records it inactivates the group instead of deleting it outright. Inactive groups are hidden by default, but can be shown by selecting a check box. Once a group has been inactivated, it can be deleted or reactivated.

Additionally I added an option to attendance analytics to view or not view inactive groups. This means you can still run reports on check-in groups you are no longer using.

# Tests
We've been running a version of these changes at Southeast for a while.

# Possible Implications
If a custom check-in workflow piece doesn't check to see if the group is active before adding it to the check-in state it could cause problems for that plug-in.

# Screenshots
inactivate1

inactivate2

inactivate3

# Documentation
https://www.rockrms.com/Rock/BookContent/10/102 may need to be updated with in attendance analytics descriptive image with an explanation of the toggle (all groups/active groups). The screenshots of the checkin configuration block and attendance analytics may need to be updated as well.

### DIFF
--- a/Rock/Web/UI/Controls/Checkin Configuration Controls/CheckinAreaRow.cs
+++ b/Rock/Web/UI/Controls/Checkin Configuration Controls/CheckinAreaRow.cs
@@ -38,7 +38,6 @@ namespace Rock.Web.UI.Controls
 
         private LinkButton _lbAddArea;
         private LinkButton _lbAddGroup;
-        private LinkButton _lblDeleteArea;
 
         /// <summary>
         /// Gets the group type unique identifier.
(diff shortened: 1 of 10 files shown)


edmistj:
Mark I like this concept. The UI would be more consistent though if we kept deletes as deletes (with a strong warning) and Inactive as checkboxes. Would you be able to make the edits below (keeping the Show Inactive Groups Options too).

photoshop_2017-03-14_22-08-45

Also it would be awesome (and simple) if you could add tooltips on these buttons.

photoshop_2017-03-14_22-14-02

Let me know your thoughts.

trydyingtolive:
Sounds great! Changes made.

edmistj:
Awesome thanks!

eddieholeman:
@edmistj Has this issue had any further development since these March postings?

dcs619:
@eddieholeman it's been merged into develop, which means it'll be in v7.  You can play around with it [here](http://prealpha.rockrms.com/page/201?CheckinTypeId=14).

eddieholeman:
@dcs619 Thanks for the clarification.  I am still learning the process.  :-)

PR #2067, opened 2017-03-14
+ Event registrations will no longer occasionally create a duplicate address on a family. (Fixes #1680)

### LINKED ISSUE #1680: Event registration creates duplicate addresses
When asking for address on an event registration, the address entered becomes "Home" and whatever was in home becomes "Previous" even if the two are identical, and even when the address was pre-populated when the user is logged in. 

If the address on the registration matches the Home address on the person's family, no update or addition should be made to addresses.

chrismcgrath:
# Contributor Agreement

Yes

# Context

When we have an event registration that asks for a person's address, occasionally if they entered the same address we already have on file for them, Rock would move the existing address record to Previous and create a new, identical Home record for them. I found that this was due to the fact that Rock assumes there will only be one Location record in the database for each address, but due to multiple data imports, many addresses had been duplicated in the Location table. When Rock tries to find the existing Location record, it would often find a different one than the one attached to the family, and attach it as well, which resulted in having a duplicate Home and Previous address.

# Goal

To prevent unnecessary address duplicates from occurring.

# Strategy

Before searching for the first found Location record that has a matching address, the Location Get method now accepts a group, which it will check for a match first. If it finds a match there, it will use that address. If not, then it will search the database for a match.  This not only prevents the duplicate problem, but should perform faster when there is a match since it does not have to query the entire Location table to find it.

Incidentally, I have also removed the initial search that happened before standardizing the address. My logic is that any address that can be standardized should already be standardized, and it is unlikely that most manually entered addresses would be in the correct standard form. So this eliminates a generally unnecessary database query each time an address is added.  If there is some reason why this logic needs to stay in, by all means put it back, its removal is not a necessary part of this pull request.

# Tests

# Possible Implications

# Screenshots

# Documentation

### DIFF
--- a/Rock/Model/LocationService.Partial.cs
+++ b/Rock/Model/LocationService.Partial.cs
@@ -31,17 +31,37 @@ namespace Rock.Model
         /// Returns the first
         /// <see cref="Rock.Model.Location" /> where the address matches the provided address, otherwise the address will be saved as a new location.
         /// </summary>
-        /// <param name="street1">A <see cref="System.String" /> representing the Address Line 1 to search by.</param>
-        /// <param name="street2">A <see cref="System.String" /> representing the Address Line 2 to search by.</param>
-        /// <param name="city">A <see cref="System.String" /> representing the City to search by.</param>
-        /// <param name="state">A <see cref="System.String" /> representing the State to search by.</param>
-        /// <param name="postalCode">A <see cref="System.String" /> representing the Zip/Postal code to search by</param>
-        /// <param name="country">The country.</param>
+        /// <param name="street1">A <see cref="string" /> representing the Address Line 1 to search by.</param>
+        /// <param name="street2">A <see cref="string" /> representing the Address Line 2 to search by.</param>
+        /// <param name="city">A <see cref="string" /> representing the City to search by.</param>
+        /// <param name="state">A <see cref="string" /> representing the State to search by.</param>
+        /// <param name="postalCode">A <see cref="string" /> representing the Zip/Postal code to search by</param>
+        /// <param name="country">A <see cref="string" /> representing the Country to search by</param>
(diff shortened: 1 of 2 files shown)


Rock/Model/LocationService.Partial.cs
```diff
@@ -42,29 +41,15 @@ public partial class LocationService
         /// <returns>
         /// The first <see cref="Rock.Model.Location" /> where an address match is found, if no match is found a new <see cref="Rock.Model.Location" /> is created and returned.
         /// </returns>
-        public Location Get( string street1, string street2, string city, string state, string postalCode, string country, bool verifyLocation = true )
+        public Location Get( string street1, string street2, string city, string state, string postalCode, string country, bool verifyLocation = true, Group group = null )
         {
```
azturner: @chrismcgrath Same comment as other pull request. This is a breaking change, but unfortunately with the existing optional `verifyLocation` param, no way to avoid that, but can you convert to use method overrides so that future changes can be done without it being a breaking change.

Rock/Model/LocationService.Partial.cs
```diff
@@ -42,29 +41,15 @@ public partial class LocationService
...
-                ( t.Country == country || ( ( country == null || country == "" ) && ( t.Country == null || t.Country == "" ) ) ) );
-            if ( existingLocation != null )
-            {
-                return existingLocation;
-            }
-
-            // If existing location wasn't found with entered values, try standardizing the values, and
-            // search for an existing value again
+            // Standardize the location for comparison to other locations
             var newLocation = new Location
```
azturner: We should probably keep this. Would rather take the small db hit to reduce chance address already exists without being standardized.

chrismcgrath:
Made the changes as requested!

PR #2072, opened 2017-03-16
Added parameter to Person Address Update workflow action to allow the existing address(es) of the specified type to be retained as a Previous Address.

chrismcgrath:
# Contributor Agreement

Yes

# Context

Updating a person's address via the Person Address Update workflow action caused us to lose the previous address when we would rather keep it for data quality purposes.

# Goal

To give the workflow creator the option of whether the action will overwrite the existing address or save it as a new address while changing the type of the existing address(es) to Previous. If there are multiple addresses of the specified type it will change all of those to Previous, but still leave other types alone.

It will also allow the user to specify True, False, or blank for the Mailing and Mapped attributes. True or False will set these fields directly on the new address, while blank will get the values from the existing address and use those.

# Strategy

If the Save Previous Address parameter is set to True, it will call GroupService.AddNewGroupAddress(), which handles the updating and creating of addresses. Also added some additional parameters to that function to allow for the mapped/mailing locations to be specified and for an additional line to be added to the history indicating where the update came from.

# Tests

# Possible Implications

# Screenshots

# Documentation


Rock/Model/GroupService.Partial.cs
```diff
@@ -760,7 +760,8 @@ public static Group SaveNewGroup( RockContext rockContext, int groupTypeId, Guid
         /// <param name="country">The country.</param>
         /// <param name="moveExistingToPrevious">if set to <c>true</c> [move existing to previous].</param>
         public static void AddNewGroupAddress( RockContext rockContext, Group group, string locationTypeGuid,
-            string street1, string street2, string city, string state, string postalCode, string country, bool moveExistingToPrevious = false )
+            string street1, string street2, string city, string state, string postalCode, string country, bool moveExistingToPrevious = false,
+            string modifiedBy = "", bool isMailingLocation = true, bool isMappedLocation = true )
```
azturner: @chrismcgrath We've found that optional parameters on public methods are difficult to maintain backward compatibility. For example, this change would break compatibility, but because of the optional parameter that was already there, there's not really any way around that. Because of that can you change this to use method overrides instead. Like this...
```
        public static void AddNewGroupAddress( RockContext rockContext, Group group, string locationTypeGuid, int? locationId )
        {
            AddNewGroupAddress( rockContext, group, locationTypeGuid, locationId, false );
        }

        public static void AddNewGroupAddress( RockContext rockContext, Group group, string locationTypeGuid, int? locationId, bool moveExistingToPrevious )
        {
            AddNewGroupAddress( rockContext, group, locationTypeGuid, locationId, moveExistingToPrevious, "", true, true );
        }

        public static void AddNewGroupAddress( RockContext rockContext, Group group, string locationTypeGuid, int? locationId, bool moveExistingToPrevious, string modifiedBy, bool isMailingLocation, bool isMappedLocation )
        {
            if ( locationId.HasValue )
            {
                var location = new LocationService( rockContext ).Get( locationId.Value );
                AddNewGroupAddress( rockContext, group, locationTypeGuid, location, moveExistingToPrevious );
            }
        }
```
That will still be a breaking change this time, but if we then need to add additional params in future, we can do it without it being a breaking change.

Rock/Workflow/Action/People/PersonAddressUpdate.cs
```diff
@@ -46,6 +46,7 @@ namespace Rock.Workflow.Action
         new string[] { "Rock.Field.Types.LocationFieldType", "Rock.Field.Types.AddressFieldType" } )]
     [WorkflowTextOrAttribute( "Is Mailing Location", "Attribute Value", "The value or attribute value to indicate if the location is the mailing address. Only valid values are 'True' or 'False' any other value will be ignored. <span class='tip tip-lava'></span>", false, "", "", 5, "IsMailing" )]
     [WorkflowTextOrAttribute( "Is Mapped Location", "Attribute Value", "The value or attribute value to indicate if the location should be mapped location. Only valid values are 'True' or 'False' any other value will be ignored. <span class='tip tip-lava'></span>", false, "", "", 6, "IsMapped" )]
+    [BooleanField( "Save Current Address as Previous Address", "Determines whether this will overwrite an existing address of the specified type or change its type to Previous and add a new address.", order: 7, key: "SavePreviousAddress" )]
     public class PersonAddressUpdate : ActionComponent
```
azturner: Can you change the default for this attribute to be False (less impact on existing users).

azturner: Nevermind, actually keep default as 'True'

Rock/Model/GroupService.Partial.cs
chrismcgrath: Committed these changes.

PR #2091, opened 2017-03-24
+ History will be properly created when adding a new person from the Edit Family screen. (Fixes #1726)

### LINKED ISSUE #1726: Creation of person from Edit Family page is not logged in history
### Description

When a new person is added from the Edit Family page, nothing is inserted into the History recording that person's data or the time it was entered. This has left us with lots of people created in our system but no record of who did it or when, which is often necessary when trying to correct bad data.
### Steps to Reproduce
1. Added a new family with two people from the Add Family screen. Each one has history entries to show when they were created and what their values were.
2. Click on Edit Family from one of these new people and choose Add Person
3. Enter new person's details and Save

**Expected behavior:** 

History is created just like it is when adding a person from the New Family page

**Actual behavior:**

No history is created.
### Versions

[RockDiagnostics-EZRS8J.txt](https://github.com/SparkDevNetwork/Rock/files/445570/RockDiagnostics-EZRS8J.txt)
[RockDiagnostics-ROCK1.txt](https://github.com/SparkDevNetwork/Rock/files/445571/RockDiagnostics-ROCK1.txt)

Rock Version: Rock McKinley 5.1
Client Culture Setting: en-US

This is happening on our local install and on the rock demo site (diagnostics attached for both).

chrismcgrath:
# Contributor Agreement

Yes

# Context

Investigating the creation of Person records, we would often find records that had no information in the History screen about when they were created, by whom, or what their values were at creation. We found that this was happening when people were added from the Edit Family screen.

# Goal

To create History records when a person is added from the Edit Family screen.

# Strategy

The ID used to save the history was the GroupMember ID rather than the Person's ID. Fixed that. 

Also removed extraneous history that was being saved to a -1 ID. 

# Tests

# Possible Implications

# Screenshots

# Documentation

### DIFF
--- a/RockWeb/Blocks/Crm/PersonDetail/EditGroup.ascx.cs
+++ b/RockWeb/Blocks/Crm/PersonDetail/EditGroup.ascx.cs
@@ -1067,9 +1067,6 @@ namespace RockWeb.Blocks.Crm.PersonDetail
 
                     foreach ( var groupMemberInfo in GroupMembers )
                     {
-                        var memberChanges = new List<string>();
-                        var demographicChanges = new List<string>();
-
                         var role = _groupType.Roles.Where( r => r.Guid.Equals( groupMemberInfo.RoleGuid ) ).FirstOrDefault();
                         if ( role == null )
                         {
@@ -1084,9 +1081,6 @@ namespace RockWeb.Blocks.Crm.PersonDetail
                             Person person = null;
                             if ( groupMemberInfo.Id == -1 )
                             {
-                                // added new person
-                                demographicChanges.Add( "Created" );
-
                                 person = new Person();
 
                                 person.TitleValueId = groupMemberInfo.TitleValueId;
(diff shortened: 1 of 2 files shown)


RockWeb/Blocks/Crm/PersonDetail/EditGroup.ascx.cs
```diff
@@ -1132,25 +1126,14 @@ protected void btnSave_Click( object sender, EventArgs e )
                                 return;
                             }
 
-                            if ( _isFamilyGroupType )
-                            {
-                                if ( person.RecordStatusValueId != recordStatusValueID )
```
nairdo: I'm not sure I understand why this code was removed.  This case occurs when adding an existing person into the edited family and when their _current/old_ record status is different than what is being set on the edit form.

RockWeb/Blocks/Crm/PersonDetail/EditGroup.ascx.cs
```diff
@@ -1246,20 +1229,20 @@ protected void btnSave_Click( object sender, EventArgs e )
                                     }
                                 }
                             }
+
```
nairdo: I'm not super familiar with this code yet and I haven't gotten to the bottom of this change yet... can you briefly describe why you moved this to happen only inside the '// existing group members' block?

RockWeb/Blocks/Crm/PersonDetail/EditGroup.ascx.cs
chrismcgrath: Got it, thanks for the explanation, I'll put these lines back as they were. To the best of my recollection I thought I had determined that these lines were either creating bad data or would never get hit, but I just confirmed that they do get hit in your scenario.

RockWeb/Blocks/Crm/PersonDetail/EditGroup.ascx.cs
chrismcgrath: With the code above removed, this code did not need to be run in all instances. I'll be putting this back as it was as well, thanks for the review.

chrismcgrath:
I reverted the changes I made that @nairdo pointed out were incorrect. This should be ready to go now.

PR #2106, opened 2017-03-29
Fixed bug where clicking the group type label selects all groups

trydyingtolive:
# Contributor Agreement
YES

# Context
One of our admins noticed that clicking the group type label in attendance analytics selects all checkboxes instead of only those for that grouptype

# Goal
It will restore intended functionality.

# Strategy
I changed the javascript to only go so high in the tree before looking for checkboxes to check/uncheck

# Possible Implications
None


edmistj:
@azturner can you look at this. It appears that the issue is fairly new (seems to work on the demo and a recent local branch of mine). I'd prefer the original  javascript as it is less likely to break with DOM changes (.closest) vs (parent().parent())

azturner:
Actually fixed this here: https://github.com/SparkDevNetwork/Rock/commit/c4d8e548b754c20f23f355d75e73c2196888126d by adding back a line that was inadvertently removed when merging a pull request.

PR #2142, opened 2017-04-10
Adding a new IRockOwinStartup interface for allowing plugins to implement OWIN functionality.

cbump:
# Contributor Agreement
Yes

# Context
We wanted to be able to have plugins hook into OWIN startup in addition to the Application Startup that is already in Rock.  OWIN provides additional functionality and a deeper level of access to IIS that isn't available via standard the standard application startup.

# Goal
This adds the ability for a plugin to hook into OWIN.

# Strategy
This is essentially the same functionality that already exists with the IRockStartup interface but uses the OWIN architecture instead of the standard Application Startup that the other interface uses.

# Tests
I don't think we need to provide any additional testing.

# Possible Implications
To my knowledge, this shouldn't affect anything.

# Screenshots
N/A

# Documentation
We might want to provide a section in the documentation explaining how to add startup functionality with a plugin and when to choose IRockStartup vs IRockOwinStartup: https://www.rockrms.com/Rock/Developer/BookContent/26/26


azturner:
@cbump I am curious though about what properties you needed from the IAppBuilder param that Owin gives you?

cbump:
I'm still trying to get a plugin in for a hosted OAuth server into the Rock shop: https://docs.microsoft.com/en-us/aspnet/aspnet/overview/owin-and-katana/owin-oauth-20-authorization-server

You can see it working if you do this: 
1. Go here - https://secc.servicereef.com/
2. Click "Sign In" at the top right of the page.
3. Click "Southeast Account" and you'll be presented with a login that's actually hosted in Rock.

We are also planning to use OAuth for mobile authentication as well so we can store a standard OAuth refresh token on a device rather than having to store a password or some other custom token.

PR #2170, opened 2017-05-01
Add the ability to re-order campuses

shivambareria:
# Contributor Agreement

Yes

# Context
Added the ability to re-order campuses.

# Strategy
addeda new [Order] field to the Campus.cs model.


azturner:
@shivambareria, Block changes were great. Not sure how the migration file was created though. It caused an error when trying to update database. I deleted the existing migration and then added it back using 'add-migration' in the package manager console, and then was able to update-database...

image

PR #2173, opened 2017-05-03
Changed tag styles to accomidate longer tags

trydyingtolive:
# Contributor Agreement
YES! 

# Context
We have tagged some folks in as Missional Community Leaders and others as Missional Community Coaches. When you look at the persons profile the tag names get cut off and you can't tell the difference.

# Goal
The goal is to make it so that you can see the full tag name on mouseover.

# Strategy
The text span width is limited 120px. When mousing over the tag it expands the max-width to 1000px. This allows for a smooth transition that isn't possible by just by removing the max-width. 

# Tests
Tested in FF, Chrome, IE, and Chrome mobile.

# Possible Implications
It should just change the styling of the tags.

# Screenshots
tags

### DIFF
--- a/RockWeb/Styles/rock-components/tags.less
+++ b/RockWeb/Styles/rock-components/tags.less
@@ -31,17 +31,27 @@ div.tagsinput {
     }
 
     span {
+      cursor: default;
       display: block;
       float: left;
       max-width: 120px;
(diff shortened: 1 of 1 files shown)


edmistj:
Nice!

PR #2181, opened 2017-05-06
Workflow action to send an email to a group filter by role

shivambareria:
# Contributor Agreement

Yes

# Context
Workflow action to send an email to a group filter by role

# Strategy
Updated the SendEmail workflow action so that it can also use an attribute that has a field type of "Rock.Field.Types.GroupRoleFieldType" for the "Send To Email Address".

### DIFF
--- a/Rock/Workflow/Action/Communications/SendEmail.cs
+++ b/Rock/Workflow/Action/Communications/SendEmail.cs
@@ -40,7 +40,7 @@ namespace Rock.Workflow.Action
         new string[] { "Rock.Field.Types.TextFieldType", "Rock.Field.Types.EmailFieldType", "Rock.Field.Types.PersonFieldType" } )]
     [WorkflowTextOrAttribute( "Send To Email Addresses", "Attribute Value", "The email addresses or an attribute that contains the person or email address that email should be sent to. <span class='tip tip-lava'></span>", true, "", "", 1, "To",
         new string[] { "Rock.Field.Types.TextFieldType", "Rock.Field.Types.EmailFieldType", "Rock.Field.Types.PersonFieldType", "Rock.Field.Types.GroupFieldType", "Rock.Field.Types.SecurityRoleFieldType" } )]
-    [WorkflowAttribute( "Send to Group Role", "The Group Role or an attribute that contains the Group Role to which email should be sent to. <span class='tip tip-lava'></span>", false, "", "", 2, "GroupRole",
+    [WorkflowAttribute( "Send to Group Role", "An optional Group Role attribute to limit recipients to if the 'Send to Email Address' is a group or security role.", false, "", "", 2, "GroupRole",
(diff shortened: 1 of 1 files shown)


Rock/Workflow/Action/Communications/SendEmail.cs
```diff
@@ -176,6 +183,16 @@ public override bool Execute( RockContext rockContext, WorkflowAction action, Ob
                                         action.AddLogEntry( "Invalid Recipient: No valid group id or Guid", true );
                                     }
 
+
+                                    if ( groupRoleValueGuid.HasValue )
+                                    {
+                                        GroupTypeRole groupTypeRole = new GroupTypeRoleService( rockContext ).Get( groupRoleValueGuid.Value );
+                                        if ( groupTypeRole != null )
+                                        {
+                                            qry = qry.Where( m => m.GroupRoleId == groupTypeRole.Id );
```
azturner: Instead of doing the query to GroupTypeRoleService a couple lines above, you can just modify this Where clause to use the Guid value...
```
    qry = qry.Where( m => m.GroupRole != null && m.GroupRole.Guid.Equals( groupRoleValueGuid.Value ) );
```

Rock/Workflow/Action/Communications/SendEmail.cs
```diff
@@ -40,9 +40,11 @@ namespace Rock.Workflow.Action
...
     [WorkflowTextOrAttribute( "Send To Email Addresses", "Attribute Value", "The email addresses or an attribute that contains the person or email address that email should be sent to. <span class='tip tip-lava'></span>", true, "", "", 1, "To",
         new string[] { "Rock.Field.Types.TextFieldType", "Rock.Field.Types.EmailFieldType", "Rock.Field.Types.PersonFieldType", "Rock.Field.Types.GroupFieldType", "Rock.Field.Types.SecurityRoleFieldType" } )]
-    [TextField( "Subject", "The subject that should be used when sending email. <span class='tip tip-lava'></span>", false, "", "", 2 )]
-    [CodeEditorField( "Body", "The body of the email that should be sent. <span class='tip tip-lava'></span> <span class='tip tip-html'></span>", Web.UI.Controls.CodeEditorMode.Html, Web.UI.Controls.CodeEditorTheme.Rock, 200, false, "", "", 3 )]
-    [BooleanField( "Save Communication History", "Should a record of this communication be saved to the recipient's profile", false, "", 4 )]
+    [WorkflowAttribute( "Send to Group Role", "The Group Role or an attribute that contains the Group Role to which email should be sent to. <span class='tip tip-lava'></span>", false, "", "", 2, "GroupRole",
```
azturner: Since this will always be an attribute (it's not a group role or an attribute), can you change description to be something like this instead...
```
"An optional Group Role attribute to limit recipients to if the 'Send to Email Address' is a group or security role."
```

PR #2184, opened 2017-05-07
Add Bulk Workflow Capabilities to Bulk Update

shivambareria:
# Contributor Agreement

Yes

# Context

Add Bulk Workflow Capabilities to Bulk Update

# Goal
Bulk Update multiple Workflows

# Strategy
Added a new section to the Bulk Update block (Blocks/CRM/BulkUpdate.ascx) that will list available workflow types that can be launched for the selected individuals.

Only a subset of workflow types should be available to choose from. This would be configured through a block setting.


RockWeb/Blocks/Crm/BulkUpdate.ascx.cs
```diff
@@ -673,22 +697,40 @@ protected void btnComplete_Click( object sender, EventArgs e )
...
+                    foreach ( ListItem item in this.rlbWorkFlowType.Items )
+                    {
+                        if ( item.Selected )
+                        {
+                            wotkFlowTypes = wotkFlowTypes + item.Text + ",";
+                        }
+                    }
+                    changes.Add( string.Format( "Activate the Workflow <span class='field-name'>{0}</span>",
+                       wotkFlowTypes ) );
+                }
```
azturner: Again, there are some extension methods that make delimiting list of strings (i.e. so comma is not after last item) and pluralizing text easier...
```
                    var wotkFlowTypes = new List<string>();
                    foreach ( ListItem item in this.rlbWorkFlowType.Items )
                    {
                        if ( item.Selected )
                        {
                            wotkFlowTypes.Add( item.Text );
                        }
                    }

                    changes.Add( string.Format( "Activate the <span class='field-name'>{0}</span> {1}.", 
                        wotkFlowTypes.AsDelimited(", ", " and " ), 
                        "workflow".PluralizeIf( wotkFlowTypes.Count > 1 ) ) );
```

RockWeb/Blocks/Crm/BulkUpdate.ascx.cs
```diff
@@ -46,7 +47,8 @@ namespace RockWeb.Blocks.Crm
     [Description( "Used for updating information about several individuals at once." )]
 
     [AttributeCategoryField( "Attribute Categories", "The person attribute categories to display and allow bulk updating", true, "Rock.Model.Person", false, "", "", 0 )]
-    [IntegerField( "Display Count", "The initial number of individuals to display prior to expanding list", false, 0, "", 1  )]
+    [IntegerField( "Display Count", "The initial number of individuals to display prior to expanding list", false, 0, "", 1 )]
+    [WorkflowTypeField( "Workflow Actions", "The workflows to make available for bulk updating.", true, false, "", "", 2 )]
```
azturner: Can you change the name/key of this setting to "Workflow Types". (instead of "Workflow Actions").

RockWeb/Blocks/Crm/BulkUpdate.ascx.cs
```diff
@@ -1297,11 +1339,32 @@ protected void btnConfirm_Click( object sender, EventArgs e )
                 }
                 #endregion
 
-                    pnlEntry.Visible = false;
+                #region workflow
+                var personsService = new PersonAliasService( rockContext );
+                var persons = personsService.Queryable()
```
azturner: This looks like its selecting a list of PersonAlias records (vs Person records). You can just use the existing `people` list that is created earlier in this method.

PR #2185, opened 2017-05-08
Added location type options to Postal Code and Geofence filters

trydyingtolive:
# Contributor Agreement
YES!

# Context
Earlier I created two data views for filtering people by geofence and one by postal code. We use these often, but they pick up all location types. I would like be able to filter only by home address or past address etc...

# Goal
By adding in option to select location type, we can search by only certain location types.

# Strategy
I added a drop down option with the group location types. If one is selected the filter only applies to that location type. If none is selected it applies to all.

# Tests
I tested all options and everything seems good to go. I also tested against existing dataviews we are using.

# Possible Implications
I wrote the enhancement in such a way to be backwards compatible with existing dataviews already using this filter.

# Screenshots
image
image


mikepetersonccv:
@trydyingtolive looks good!

PR #2197, opened 2017-05-12
Support 0.0.0.0:9100 syntax in printer IP addresses

cabal95:
# Contributor Agreement

**Yes**

# Context

Lack of ability to have multiple printers on one IP address (e.g. a single Airport Express with multiple USB printers plugged in).

# Goal

Add support for printer devices to be defined with 0.0.0.0:9100 syntax. e.g. 10.0.0.42:9102.

This will have a "side effect" of allowing remote server based printing without a VPN with the proper router configuration. You can setup 5 printers in Rock to print to your public IP address, each on a different port (say 9100 - 9104). Then your router can do reverse NAT to direct 9100 -> 10.0.0.42:9100, 9101 -> 10.0.0.43:9100, 9102 -> 10.0.0.44:9100, and so on.

# Strategy

The code for printing a test label and a full during check-in label have been updated to support the new syntax.

# Tests

Would love to come up with a way to unit test printing to a physical printer, but alas I got nothing.

# Possible Implications

If somebody was crazy enough to define an IP address in IPv6 notation then this would break printing to that printer. I'm not sure the iOS app even supports IPv6 notation either.

# Screenshots

n/a

# Documentation

Check-in docs might need a footnote added here https://www.rockrms.com/Rock/BookContent/10/102#printers

### DIFF
--- a/RockWeb/Blocks/CheckIn/Config/EditLabel.ascx.cs
+++ b/RockWeb/Blocks/CheckIn/Config/EditLabel.ascx.cs
@@ -199,10 +199,22 @@ namespace RockWeb.Blocks.CheckIn.Config
                 if ( device != null )
                 {
                     string currentIp = device.IPAddress;
-                    var printerIp = new IPEndPoint( IPAddress.Parse( currentIp ), 9100 );
+                    int printerPort = 9100;
+                    var printerIp = currentIp;
+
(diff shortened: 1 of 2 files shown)


cabal95:
Meant to note that similar logic has already been added to the iOS app in the latest update.

PR #2203, opened 2017-05-17
Extended the Current ModelMap block to group model on RockDomain Attribute.

shivambareria:
# Contributor Agreement

Yes

# Context

Extended the Current ModelMap block to group model on RockDomain Attribute.

# Strategy

Created RockDomain Attribute to include Name and IconCssClass.
Extended the existing ModelMap block to group on the basis of Value defined in RockDomain Attribute.


RockWeb/Blocks/Examples/ModelMap.ascx.cs
```diff
@@ -258,7 +407,7 @@ private MClass GetPropertiesAndMethods( Type type, bool includeInherited = true
                     Name = p.Name,
                     IsInherited = p.DeclaringType != type,
                     IsVirtual = p.GetGetMethod() != null && p.GetGetMethod().IsVirtual,
-                    IsLavaInclude = p.IsDefined( typeof( LavaIncludeAttribute ) ),
+                    IsLavaInclude = p.IsDefined( typeof( LavaIncludeAttribute) ),
```
azturner: This should also check to see if [DataMember] is defined on property as well.

RockWeb/Blocks/Examples/ModelMap.ascx.cs
```diff
@@ -246,10 +389,16 @@ private MClass GetPropertiesAndMethods( Type type, bool includeInherited = true
             MClass mClass = new MClass
             {
                 Name = type.Name,
+                Category = new Category
+                {
+                    Name = type.IsDefined( typeof( RockDomainAttribute ), true ) ? type.GetCustomAttribute<RockDomainAttribute>( true ).Name : "Other",
+                    IconCssClass = type.IsDefined( typeof( RockDomainAttribute ), true ) ? type.GetCustomAttribute<RockDomainAttribute>( true ).IconCssClass : "fa fa-file-o"
+                },
```
azturner: Can you do this in a way where you don't have to call IsDefined and GetCustomAttribute twice for each model.  Something like:
```
var attr = type.GetCustomAttribute<RockDomainAttribute>()
...
Category = new Category
{
    Name = attr != null ? attr.Name : "Other";
    IconCssClass = attr != null ? attr.IconCssClass : "fa fa-file-o";
}
```

RockWeb/Blocks/Examples/ModelMap.ascx.cs
```diff
@@ -56,15 +84,138 @@ protected override void OnLoad( EventArgs e )
...
+                    SelectedCategory = qry.Select( a => a.Key ).First();
                 }
+
+                var displayCategoryList = qry.
+                                    Select( a => new
+                                    {
+                                        Category = new Category
+                                        {
+                                            Name = a.Key,
+                                            IconCssClass = a.Select( g => g.Name ).First(),
```
azturner: Not sure this is actually getting the IconCssClass. Shouldn't it be something like 
```
IconCssClass = a.Select( g=> g.Category.IconCssClass ).FirstOrDefault()
```

azturner:
Didn't pull this in because wasn't sure how the last two commits would get applied since it was a merge from core. Instead, just applied the first three commits to core here: https://github.com/SparkDevNetwork/Rock/commit/54ba2810335941b7c0d90c2eb4b33c33b927fde6

PR #2224, opened 2017-05-28
Clone a Security Role

shivambareria:
# Contributor Agreement

Yes

# Context
Cloning Security Role

# Strategy
Cloned the properties of Security Group with all the authorization rules.


Rock/Model/AuthService.Partial.cs
```diff
@@ -57,5 +57,19 @@ public IQueryable<Auth> GetAuths( int entityTypeId, int? entityId, string action
...
         }
+
+        /// <summary>
+        /// Returns an enumerable collection of <see cref="Rock.Model.Auth"/> entities by <see cref="Rock.Model.EntityType"/> and entity Id.
+        /// </summary>
+        /// <param name="groupId">A <see cref="System.Int32"/> represent the groupId of the security Role.</param>
+        /// <returns>
+        /// An enumerable list of <see cref="Rock.Model.Auth" /> entities that secure a specific entity.
+        /// </returns>
+        public IQueryable<Auth> GetAuths( int groupId )
```
azturner: Can you rename this to `GetByGroup`. Ideally it it would just an overload of the `Get` method with different method signature (parameter list), but with just a single int parameter, need to use a method name that is unique.

RockWeb/Blocks/Groups/GroupDetail.ascx
```diff
@@ -277,6 +277,7 @@
                             <span class="pull-right">
                                 <asp:HyperLink ID="hlAttendance" runat="server" CssClass="btn btn-sm btn-default" ToolTip="Attendance"><i class="fa fa-check-square-o"></i></asp:HyperLink>
                                 <asp:HyperLink ID="hlMap" runat="server" CssClass="btn btn-sm btn-default" ToolTip="Interactive Map"><i class="fa fa-map-marker"></i></asp:HyperLink>
+                                <asp:LinkButton ID="btnCopy" runat="server" CssClass="btn btn-default btn-sm fa fa-clone" OnClick="btnCopy_Click" ToolTip="Copy Group" />
```
azturner: Can you change the ToolTip text to `Copies the group and all of its associated authorization rules`

RockWeb/Blocks/Groups/GroupDetail.ascx.cs
```diff
@@ -825,6 +825,102 @@ protected void btnCancel_Click( object sender, EventArgs e )
...
+            var rockContext = new RockContext();
+            var groupService = new GroupService( rockContext );
+            var authService = new AuthService( rockContext );
+            var attributeService = new AttributeService( rockContext );
+
+            int groupId = hfGroupId.ValueAsInt();
+            var group = groupService.Queryable( "GroupType" )
+                    .Where( g => g.Id == groupId )
+                    .FirstOrDefault();
+            group.LoadAttributes( rockContext );
```
azturner: Should probably move this line so that it's inside the next if statement.

RockWeb/Blocks/Groups/GroupDetail.ascx.cs
```diff
@@ -825,6 +825,102 @@ protected void btnCancel_Click( object sender, EventArgs e )
...
+                    groupService.Add( newGroup );
+                    rockContext.SaveChanges();
+
+                    newGroup.LoadAttributes( rockContext );
+                    if ( group.Attributes != null && group.Attributes.Any() )
+                    {
+                        foreach ( var attributeKey in group.Attributes.Select( a => a.Key ) )
+                        {
+                            string value = group.GetAttributeValue( attributeKey );
+                            Guid guidValue = value.AsGuid();
```
azturner: Not sure this line is needed.

RockWeb/Blocks/Groups/GroupDetail.ascx.cs
```diff
@@ -825,6 +825,102 @@ protected void btnCancel_Click( object sender, EventArgs e )
...
+                        foreach ( var qualifier in attribute.AttributeQualifiers )
+                        {
+                            var newQualifier = qualifier.Clone( false );
+                            newQualifier.Id = 0;
+                            newQualifier.Guid = Guid.NewGuid();
+                            newQualifier.IsSystem = false;
+                            newAttribute.AttributeQualifiers.Add( qualifier );
+                        }
+
+                        Rock.Attribute.Helper.SaveAttributeEdits( newAttribute, entityTypeId, qualifierColumn, newGroup.Id.ToString(), rockContext );
```
azturner: Using the Helper.SaveAttributeEdits isn't really needed here. You should just be able to use something like `attributeValueService.Add( newAttribute );` and then call `rockContext.SaveChanges()` again to save the new values.

RockWeb/Blocks/Groups/GroupDetail.ascx.cs
```diff
@@ -825,6 +825,102 @@ protected void btnCancel_Click( object sender, EventArgs e )
...
+                        newAuth.CreatedByPersonAliasId = null;
+                        newAuth.CreatedDateTime = RockDateTime.Now;
+                        newAuth.ModifiedByPersonAlias = null;
+                        newAuth.ModifiedByPersonAliasId = null;
+                        newAuth.ModifiedDateTime = RockDateTime.Now;
+                        authService.Add( newAuth );
+                    }
+                    rockContext.SaveChanges();
+                } );
+                ShowDetail( newGroup.Id );
```
azturner: Because all of the auth rules are cached by Rock, you should add a call to `Rock.Security.Authorization.Flush()` here after saving changes.

Issue #2171, opened 2017-05-01
[Feature] New block for viewing effective security for entities


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

It is not uncommon for new users, or even us experienced users, to click things in the wrong order and lock ourselves out of a page, group, or whatever. When that happens currently the only way to fix it is to go into SQL and hunt down the offending record and delete it by hand.

While this is something those knowledge in SQL can do, new users are not so readily able to figure this out on their own. It can also be difficult to walk them through the process, especially if a mis-typed DELETE statement is entered and wipes out their entire Auth table. And to be honest I think most of us are hesitant to start throwing out DELETE statements at our own production database even if we have experience. We've probably all accidentally hit enter too soon or something and cried.

### Proposal

A new block that could be installed in Power Tools that allows `RSR - Rock Administration` users to see the effective permissions for either themselves or a specific person. This is a generic use block that the user would select the entity type to be inspected and enter the entity Id (as either an integer or Guid). This means it would still be a somewhat advanced usage block that brand new users would probably not know exactly what to enter, but after using it once or twice (with guidance from the community or documentation) they should be able to make use of it on future issues.

<img width="468" alt="trace security" src="https://cloud.githubusercontent.com/assets/1119863/25597037/4db9a77c-2e81-11e7-957b-80b1d6050cc4.png">

The block offers three elements of functionality.

1. It allows you to view effective permissions of any person in the database, not just yourself. This comes in handy when somebody that should have access does not and you don't see the obvious reason. You can view the effective permissions that person has to the entity in question to figure out where the Deny rule is coming from.
2. Shows only the effective permissions rather than all inherited and explicit permissions. Sometimes I miss the fact that somebody is denied access to something because they are in a group that is Deny access due to ordering. Additionally, the source entity where the explicit permission is defined is displayed.
[...]

PR #2270, opened 2017-06-22
Implement new block to inspect security permissions on entities

cabal95:
# Contributor Agreement

**Yes**

# Context

See #2171.

# Goal

New block for inspecting the effective permissions a user has to an entity (such as a Page). Also provides the ability to pop the trunk to regain access if the administrator has locked themselves out of a page or entity.

# Strategy

Wrote a spiffy looking block.

It searches the database for the permissions the specified user has and displays the one that matches the user in question. Shows the admin what access the user has to something as well as where that access is coming from (what parent entity explicitly defines that access).

If the user does not have access the admin can "pop the trunk" for that user (usually themselves).

# Tests

n/a

# Possible Implications

Inherits permissions from the "Security" page, which by default is RSR - Rock Administrators only. It would be bad for the admin to allow other users access to this block as they could give themselves access to anything.

# Screenshots

See referenced issue.

# Documentation

Probably could use a mention in the Hero Admin Guide https://www.rockrms.com/Rock/BookContent/9/98#securitysettings on how to use the block to check a user's access to something as well as regain access to (usually a page) they have locked themselves out of.


edmistj:
I manually pulled this PR in due to the nature of the migration.

PR #2281, opened 2017-06-27
+ Added search option for Additional Accounts to Transaction Entry

treyhendon:
# Contributor Agreement
Yes

# Context
When there are numerous Additional Accounts, it's difficult to find the desired Account.

# Goal
Provide a search feature for the Transaction Entry block which allows for all Additional Funds to be displayed in a quick search tool.

# Strategy
Added a block setting to populate a new repeater in the Transaction Entry block.  The search functionality is powered by JavaScript showing/hiding the returned Accounts.

# Tests
N/A

# Possible Implications
Should not affect anything.

# Screenshots
fundsearchquickdemo720

# Documentation

### DIFF
--- a/RockWeb/Blocks/Finance/TransactionEntry.ascx
+++ b/RockWeb/Blocks/Finance/TransactionEntry.ascx
@@ -37,6 +37,17 @@
                                         </asp:Repeater>
                                         <Rock:ButtonDropDownList ID="btnAddAccount" runat="server" Visible="false" Label=" "
                                             DataTextField="PublicName" DataValueField="Id" OnSelectionChanged="btnAddAccount_SelectionChanged" />
+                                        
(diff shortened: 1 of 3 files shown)


edmistj:
Thanks for the PR. We discussed this in our weekly core dev meeting and we're not going to pull it into core as is due to the UX not being consistent with the rest of core. Using an Account Picker might be a better option or we have a new Harvest Chosen picker control in v7 which does this very thing in a concise way (https://harvesthq.github.io/chosen/). If you'd like to propose an alternate option it would be best to let us review a mock-up before developing. That way we could collaborate on a final solution before the code is developed.

treyhendon:
I'll run the Harvest Picker by the church who requested and see if that hits their need.  Thanks!

Issue #1776, opened 2016-09-21
[EN] Rock Cleanup job for orphaned Attribute Values


dcs619:
As our database is growing, I've been monitoring the largest tables and looking for ways to clean them up (see #1670).  Our 3rd largest table is `AttributeValue` and I don't see anything set to clean that automatically.  

My thought is to create a job to JOIN any Attribute with the Entity that _should_ exist according to  Attribute.EntityTypeId, Attribute.EntityTypeQualifierColumn and Attribute.EntityTypeQualifierValue.  If that Entity doesn't exist, then any AttributeValues for that Attribute can be deleted.

One use case for this is a Group that's been deleted, but has been given custom/inherited attributes and those attributes are never deleted.  The GroupMembers are deleted, but the attribute values for the members aren't.  I'm sure there are other situations with orphaned AttributeValues too.

cabal95:
Can anybody think of a reason this wouldn't work?  It seems a pretty straight forward process. I don't think we even need to check the EntityTypeQualifierColumn and EntityTypeQualifierValue.  If the EntityType of the Attribute does not have an entity table row whose Id matched EntityId it should be safe to delete.  The Qualifiers just limit which _attributes_ apply to an entity, not so much the attribute values (though indirectly they do).

So, if Attribute.EntityType + AttributeValue.EntityId does not match to a record, it should be safe to delete the AttributeValue - correct?

azturner:
@cabal95 Yes, as long you are only looking at attribute values that are associated to an attribute that has an entity type that is an actual entity (IsEntity=1) that is a safe assumption.

dcs619:
I thought there was a case where the EntityTypeQualifierValue was needed  -- something to do with Group Member attribute values that were inherited from a parent grouptype?  

There are a lot of IsSystem (core) attributes that reference non-entities, maybe anything marked `IsSystem` should be excluded as well?

```
select 
    a.Name Attribute, a.IsSystem SystemAttribute, av.value Value, av.IsSystem SystemAttributeValue, e.name EntityType
from attribute a
inner join entitytype e
    on a.entitytypeid = e.id
inner join attributevalue av
    on av.attributeid = a.id
where 
    e.isentity = 0
```

MikeDPeterson:
[...]

Issue #2272, opened 2017-06-23
Deleting a Registrant record doesn't delete attribute value


BEGamache:
### Prerequisites
* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

Not exactly a bug or issue but I noticed when a registrant is deleted from a registration the registration attribute(s) for that registrant aren't removed from the AttributeValue table. Not certain if removing the registrant's attribute values when the registrant is removed should occur or if leaving it the way it is is fine.

Just thought I'd mention it.

### Versions

* **Rock Version:** 6.5

dcs619:
Think this is a known issue (#1776), any attributes added to entities won't be deleted when the entity is deleted.

Issue #1979, opened 2017-01-16
[EN] Workflow Attributes are orphaned when workflow is deleted


JimMichael:
### Prerequisites

* [x] Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
* [x] Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
* [x] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

I'm not sure if this is a bug, an architectural defect, or Working As Designed, but we've discovered that Workflow Attributes don't seem to be deleted when a Workflow is deleted (at least when those attributes are files), and instead "hang around" (orphaned) and we are concerned that over time the db (or file system) will be bloated with many attributes no longer used but are still taking up space. 

### Steps to Reproduce

1. Create a new File Type with File System storage type.
2. Create a simple workflow with a File attribute (of your new file type) along with a User Entry Form that exposes this attribute
3. Run the workflow and upload a file to the WF attribute. Observe that it is now sitting on the file system (~/App_Data/Files unless you changed that in the File Type)
4. Delete the workflow and observe that the file remains on the file system, thus "orphaned" from the workflow that created it and not referenced by anything else in Rock.

**Expected behavior:** 

Workflow attributes would be deleted when the WF that owns them is deleted. Person attributes do work correctly in this regard, but Workflow Attributes do not.

**Actual behavior:** 

Workflow attributes are orphaned. We have only tested this with WF attributes stored on the file system but assume it happens to db-stored attributes as well. Does some sort of cleanup job need to be created to search for such attributes and remove them when there are in this orphaned state? Or are we mis-interpreting what we're seeing?

### Versions

* **Rock Version:** 6.1
* **Client Culture Setting** en-US

dcs619:
I think this is a known issue, see #1776.  The files are at least cleaned up but Attribute Values aren't.

JimMichael:
Thanks David, sorry I missed that one in my search. Are you saying that (attribute) files ARE eventually cleaned up? I've had test attributes/files sitting out there 4+ days and they haven't gone away (yet.) What process would "eventually" clean them? I do like the idea in #1776 of a job that tries to join attributes to their entity and whack them if it can't.

[...]

PR #2282, opened 2017-06-28
Add functionality to the Rock Cleanup job to remove orphaned attributes from deleted entities

dcs619:
# Contributor Agreement

# Context

# Goal

# Strategy

# Tests

# Possible Implications

Ideally [Rock.Rest.Controllers.MetricYTData](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Rest/Controllers/MetricsController.partial.cs#L170) wouldn't be a public `IModel, IEntity, IHasAttributes` object with no context services.  Otherwise the [Where](https://github.com/SparkDevNetwork/Rock/compare/develop...feature-rock-cleanup#diff-50bd58becd76842cd7f87733ac3df1d7R675) clause on L175 would be much cleaner.

### DIFF
--- a/Rock/Jobs/RockCleanup.cs
+++ b/Rock/Jobs/RockCleanup.cs
@@ -26,6 +26,7 @@ using Rock.Data;
 using System.Data.Entity;
 using Rock.Web.Cache;
 using Rock.Field.Types;
+using System.Reflection;
 
 namespace Rock.Jobs
 {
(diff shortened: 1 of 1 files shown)


cabal95:
Looks good to me.

This is one of those things that as many of us as possible should test on our live-data dev systems before it rolls out to production. Logic looks sound to me, but any time mass deletes are involved... well, better safe than sorry. :)

dcs619:
To clarify, this will fix #1776, #1979, and #2272 which are all the same problem: deleted entities don't remove their associated attribute values.

PR #2293, opened 2017-07-03
Feature Disable Name Edit

treyhendon:
+ Added ability to disable First and Last Name editing in Public Profile Edit Block.

# Contributor Agreement
Yup!

# Context
Users are modifying names in the Public Profile Update block in ways that caused data integrity issues.  i.e. `John and Jane` in the First Name Field or completely changing the name to another family member.

# Goal
Allow the data team to decide if the First and Last Name fields should be enabled when End Users edit.

# Strategy
Added an Advanced Block setting to Disable Name Edit.

# Tests
None

# Possible Implications
Defaults to allow editing (like existing behavior) so no impact to most users.

# Screenshots
disablenameedit

# Documentation
None


edmistj:
Thanks! This block setting can be moved to the regular block settings. We don't add too many to the advanced tab.

edmistj:
@treyhendon are you able to move this block setting?

azturner:
@treyhendon,  did you see Jon's note about moving block setting?

cc: @dcs619

PR #2302, opened 2017-07-07
Fixed FundraisingParticipant block to not throw exception when home address missing

cabal95:
# Contributor Agreement

**Yes**

# Context

If a contributor to a fundraising trip participant has no home address then an exception is thrown because the address is not checked for null before being referenced.

# Goal

Add null checking.

# Strategy

Treat a null home address like an anonymous gift (i.e. show an empty string instead of the address).

# Tests

n/a

# Possible Implications

n/a

# Screenshots

n/a

# Documentation

n/a


cabal95:
Found another bug. Second commit fixes the contributions data not showing if the updates tab was also not visible. The button was marked as hidden but the content not set explicitly visible.

PR #2314, opened 2017-07-11
Added support for editing Device attributes.

cabal95:
# Contributor Agreement

**Yes**

# Context

Could not add attributes to devices unless I went through SQL which is not ideal.

# Goal

Allow administrators to have a safe and comfortable user interface that allows them to view and edit device attributes. Our goal should always be to make administrators feel like they are empowered on their own installation of Rock. Whenever possible we should do so in a way that makes them feel safe in providing powerful features to their users. If we fail at this task then we should be ashamed of ourselves.

# Strategy

Copied and pasted a lot of code from the content channel blocks.

# Tests

n/a

# Possible Implications

Administrators may be stuck with their new found freedom of empowerment.

# Screenshots

image

image

image

# Documentation

n/a

### DIFF
--- a/RockWeb/Blocks/Core/DeviceDetail.ascx
+++ b/RockWeb/Blocks/Core/DeviceDetail.ascx
@@ -11,6 +11,7 @@
         <asp:Panel ID="pnlDetails" CssClass="panel panel-block" runat="server" Visible="false">
 
             <asp:HiddenField ID="hfDeviceId" runat="server" />
+            <asp:HiddenField ID="hfTypeId" runat="server" />
 
             <div class="panel-heading">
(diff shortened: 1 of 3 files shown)


azturner:
@cabal95, after merging this change, the Delete button on the list does not appear to be working. I suspect you'll need to remove that column from the markup and add it dynamically after adding the attribute columns and rebinding the event when you do the add. Can you take a look and submit a PR with fix.

cabal95:
@azturner Will do. I'll take a look at the WorkflowList block, I think it does a similar display so I'll use that as a template for the ShowOnGrid functionality.

I assumed that would work as-is since it didn't throw any errors, but guess not.

PR #2315, opened 2017-07-11
Added Content Channel Cache

shivambareria:
# Contributor Agreement
Yes

# Goal
Added Content Channel Cache that is required by the rendering engine


MikeDPeterson:
@shivambareria Looks good, but probably needs ParentAuthority to be defined.  See CategoryCache.cs as an example of another Cache object that does this, and then ContentChannel.cs to see what it should be

MikeDPeterson:
Looks good!

PR #2321, opened 2017-07-13
Fix SMS Replies Not Matching If Your Defined Value Number Contains a Space

arranf:
# Context
The defined values containing our Twilio SMS numbers mysteriously had spaces added for formatting purposes. 

# Goal
To prevent someone having to waste time tracking down a issue that's very difficult to diagnose! Twilio also gives you spaces in your number by default.

# Strategy
Strip whitespace from the `toPhone` number before querying.

# Possible Implications
None

### DIFF
--- a/Rock/Communication/Medium/Sms.cs
+++ b/Rock/Communication/Medium/Sms.cs
@@ -169,6 +169,7 @@ namespace Rock.Communication.Medium
             errorMessage = string.Empty;
             
             string transportPhone = string.Empty;
+            toPhone = toPhone.RemoveSpaces();
 
             using ( Rock.Data.RockContext rockContext = new Rock.Data.RockContext() )
             {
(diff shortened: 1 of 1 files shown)


azturner:
@arranf , this looks like it's fixing issue if the number received from Twilio by the webhook contains a space and defined value number does not. Is it possible people may have intentionally already added a space then to their defined value to match the number returned by Twilio (not sure if that would work since same number would be used for sending)? If so, this would then break that since stripping space would no longer match. Would it be better to just change the compare to remove spaces from both values? i.e.
```
var matchValue = definedType.DefinedValues.Where( v => v.Value.RemoveSpaces() == toPhone.RemoveSpaces() ).OrderBy( v => v.Order ).FirstOrDefault();
```

arranf:
Sorry, that was me trying to do two tasks at once and doing a poor job of
it. The Twilio number is always sent in E.164 format which won't contain
spaces, I meant to do v.Value.RemoveSpaces().

arranf:
Okay, I've updated my commit to have `.RemoveSpaces()` on the correct variable 🤦

azturner:
This was fixed in the Communication 2 update here: https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Communication/Medium/Sms.cs#L226

But also manually made this change to the v1.6.10 branch here:
https://github.com/SparkDevNetwork/Rock/commit/5eecddedfd028bdc5d6913baecc3e15236756bf4

Issue #2330, opened 2017-07-16
Check-in completely breaks and chews up CPU when attendance codes are exhausted.


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

HDC discovered this little nugget. They had changed their attendance code length to 2 characters, alpha only. To an average user that would mean 676 (26 * 26) possible codes. But since only 17 characters are available for alpha use, that limits them to 289 possible codes.

What visually happened is that after selecting the people to check-in and tapping the Next button, the screen would sit there for awhile and then eventually go back to the welcome screen. What was happening behind the scenes is that IIS would spike to 100% CPU and stay that way until the request timed out and killed off the thread.

Now for the bug. There are two places in AttendanceCodeService that use a `while` loop with no last-resort escape.

https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Model/AttendanceCodeService.Partial.cs#L122
https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Model/AttendanceCodeService.Partial.cs#L136

Both of these while loops can go on forever if all possible codes have been used. Suggested fix would be a Stopwatch that throws an error after say 5 seconds. At the very least this way an exception would be recorded in the exception log saying what the problem is.

### Steps to Reproduce

1. Set check-in code length/type to alpha only 1 character length.
2. Check-in 17 kids.
3. Try to check-in number 18 and watch IIS twitch uncontrollably.

**Expected behavior:** 

An error should be reported rather than silently breaking.

**Actual behavior:** 

People state dumbfounded at the screen while staff scrambles to start manually check-in kids in.

### Versions

* **Rock Version:** 6.6
* **Client Culture Setting:** en-US

arranf:
I think a long term fix for this could include some sort of warning about the number of possible attendance codes.

arranf:
@dcs619 Is this something you're still working on?

chrismcgrath:
[...]

PR #2358, opened 2017-07-19
Fixed infinite loop when too few check-in codes can be computed

dcs619:
# Contributor Agreement

# Context

See initial discussion at #2330.  I've limited the number of loops to the possible number of checkin codes: [# of alpha or numeric character codes] ^ [desired code length].  Since the Generate() method may pick the same no-good twice, this may stop the loop before it finds the last possible code, but the list would be close to exhausted before the loop does stop.

@cabal95 do you mind testing?

### DIFF
--- a/Rock/Model/AttendanceCodeService.Partial.cs
+++ b/Rock/Model/AttendanceCodeService.Partial.cs
@@ -67,7 +67,7 @@ namespace Rock.Model
 
         /// <summary>
         /// Returns a new <see cref="Rock.Model.AttendanceCode" /> with a specified number of alpha characters followed by
-        /// another specified number of numeric characters.  The numeric character sequence will not repeat for "today" so 
+        /// another specified number of numeric characters.  The numeric character sequence will not repeat for "today" so
         /// ensure that you're using a sufficient numericLength otherwise it will be unable to find a unique number.
         /// Also note as the issued numeric codes reaches the maximum (from the set of possible), it will take longer and
         /// longer to find an unused number.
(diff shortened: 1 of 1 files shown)


cabal95:
Yes, I'll work on testing this out.  All I will be testing is checking in on a non-patched branch until it fails and then switch to the this branch to make sure it lets me continue checking in.

cabal95:
That "works", in that it now gives some feedback about the fact it failed. An exception is logged and displayed on the client:

image

I think we should check if the `attendanceCode` is null and if so raise an exception with a message that explicitly states why the exception happened rather than just the generic one that might not make sense to an system administrator. The above exception is from this line where the `GetNew()` method returns null since it couldn't generate any attendance code.

https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Workflow/Action/CheckIn/SaveAttendance.cs#L81

Maybe something as simple as:
```
    if ( attendanceCode == null )
    {
        throw new Exception( "All possible Attendance Codes have been used. Please increase code length." );
    }
```

dcs619:
Yep, I didn't test the workflow action could handle NULL @cabal95. I'll add that in.  I'm not a huge fan of throwing an exception but if I can't find a way to pass a graceful message back that at least would log for a Rock admin to follow up with.

cabal95:
A graceful message would be nice, but I'm personally not that worried about it.  I mean this is the first time we have run into this issue in the history of Rock AFAIK so it's not like this is an error that will pop up a lot.

azturner:
@dcs619 This looks good, but it looks like from the comments that you were going to make another change to handle the null value returned. Are you still planning to make that change?

dcs619:
@azturner Yes it's still on my list. I'm trying to get back to a "clean" working branch so I can test the changes

nairdo:
Greetings @dcs619 , I think David T is passing this to me for follow-up or closing.  It looks like some changes (such as https://github.com/SparkDevNetwork/Rock/commit/4f321c6812023fe3a49527bf5fb1b97b5fdbb1ab#diff-187bba5e6f8c59c9d3c0c92c3f2d5176) since your original PR are causing a merge conflict so you might want to re-PR over the latest stuff when you test.

nairdo:
I'm closing this pull-request, but feel free to re-submit a new one once it's ready.

Issue #34, opened 2012-01-06
Moving blocks to layout zones


jonedmiston:
Currently you can't move a block to a layout zone (only a page zone).

Issue #33, opened 2012-01-06
Layout Block Cache Issue


jonedmiston:
When I add a new layout block from a page it doesn't show up on other pages with the same layout until I restart the app.

PR #2362, opened 2017-07-20
Fixed Line Break Issue in Contact Information Column

shivambareria:
# Contributor Agreement
Yes

# Context
Fixed Line break issue in Contact Information Column inside Business List Block

image

### DIFF
--- a/RockWeb/Blocks/Finance/BusinessList.ascx
+++ b/RockWeb/Blocks/Finance/BusinessList.ascx
@@ -30,8 +30,8 @@
                             <Rock:RockTemplateField>
                                 <HeaderTemplate>Contact Information</HeaderTemplate>
                                 <ItemTemplate>
-                                    <%# Eval("PhoneNumber") %><br />
-                                    <%# Eval("Email") %>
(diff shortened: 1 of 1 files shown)


RockWeb/Blocks/Finance/BusinessList.ascx
```diff
@@ -30,8 +30,8 @@
                             <Rock:RockTemplateField>
                                 <HeaderTemplate>Contact Information</HeaderTemplate>
                                 <ItemTemplate>
-                                    <%# Eval("PhoneNumber") %><br />
-                                    <%# Eval("Email") %>
+                                    <%# Eval("PhoneNumber") %>  <%# !string.IsNullOrEmpty( (string)Eval( "PhoneNumber" ) )? "<br />":string.Empty %>
+                                    <%# Eval("Email") %> <%# string.IsNullOrEmpty( (string)Eval( "PhoneNumber" ) )? "<br />&nbsp; ":string.Empty %>
```
azturner: @shivambareria Not sure why the email would ever need the `<br/>&nbsp`, and it's also checking  if PhoneNumber is empty.  Probably just need the change on line #33 (and not the change on #34).

edmistj: @azturner I think this was part of the requirements to always have that column have two lines so all the rows would be the same height (most will have both a phone and email).

PR #2415, opened 2017-08-26
Added User Login Filter on Person

shivambareria:
# Contributor Agreement
Yes

# Goal
Added related person data view for user login.
+ Need a list of people who have a login
+ Has logged in since xxxxxx (sliding date range)
+ Created since
+ Login type


azturner:
@shivambareria Rather than adding a Person dataview filter with those specific criteria to evaluate on the User Login table, can you just make this be a simple `UserLoginDataViewFilter` (in the "Related Data View Filter" section). This would be just like the existing `HistoryDataViewFilter` or `LocationDataViewFilter`. These just include people who are part of a filter from a different entity type.  That way we have more flexibility to filter on anything specific to the User Login data (including the four things mentioned in the description above). Does that make sense?

Issue #2104, opened 2017-03-28
[Enhancement] Update Attribute Lava Tag


NuclearTacos:
### Prerequisites
* [x] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Back Story (Optional)

As I continue to use Workflows, and make systems that work beyond a singular linear workflow, I find myself in need of being able to set Attributes for Entities that aren't supported by an Activity or UI.

The use-case that brought this to my attention the most was making a Workflow/Connection system to help shepherd people into a Small Group when they don't know which specific group they want, just the parameters they want (like age range, day of week, etc).

I had Workflow A be a form Workflow to take the traits they were looking for, then it created a Connection Request.  That Connection Request started Workflow B (so that the workflow would be accessible from the Connections page) and that Workflow would access those Attributes that we entered into Workflow A and filter the available Small Groups by that criteria.

The hitch was there wasn't a direct way to pass the Attributes from Workflow A to B **because** Workflow B wasn't _started_ by Workflow A, but rather the Connection Request.  I handled this by including the ID of Workflow A in the Connection Request Comment, then using an entity command to access Workflow A and it's attributes from Workflow B.

This is obviously far from ideal, and what would be amazing is to be able to set the attributes of the Connection request in Workflow A.  I mean, I can create the Attributes through the UI, I can retrieve their values in Lava, but I can _only_ set them in SQL and then I have to clear the cache to make sure everything works the way I'd expect it to.

### Proposed Enhancement

I think this is a ripe target for a Lava Tag.  The main goal would be to have a tag that allows you to update a target Attribute's value for an Entity. I think that it warrants its own security setting in line with how Lava SQL and Entity Commands have to be turned on.  I also think that it should adhere to the security settings of the one who saves it (e.g. You can't use it to edit an attribute that you couldn't edit otherwise due to security settings).
#### Example Implementation
Even though these examples are using a Person entity (which is supported with Workflow actions and the UI, I can produce as many use-cases as you'd like to make your decision.
[...]

PR #2458, opened 2017-09-13
+ Added Workflow Actions for setting a Property or Attribute of any Entity.

Tschrock:
# Contributor Agreement

👍 

# Context

As discussed in #2104, It would be great to have a way of setting attributes (and properties) on any entity inside a workflow. This would allow us to work with any entity we want, instead of being limited to the existing Actions or resorting to SQL 😨 to update things.

# Goal

This change adds two new Workflow Actions for setting a property or attribute value for any entity.

# Strategy

I've added two new workflow actions - **Set Entity Property** and **Set Entity Attribute**

Both actions take an Entity Type, the Entity's Id or Guid, and the key and value of the attribute/property to set.

### Set Entity Property
This action sets a property on an entity. Property values are converted to their appropriate type and empty values are set depending on the `Empty Value Handling` option (Explanation here: https://github.com/SparkDevNetwork/Rock/issues/2104#issuecomment-328620103). Entity Framework and SQL Server ensure values are still valid (You can't change an Entity's Id for example).

### Set Entity Attribute
This action sets an attribute's value for an entity.

# Tests

🤷‍♂️ 

# Possible Implications

🚫 

# Screenshots

### Set Entity Property
firefox_2017-09-13_14-22-22

### Set Entity Attribute
firefox_2017-09-13_14-23-51

# Documentation

📖

Both of these new actions will need documented with the other workflow actions (https://www.rockrms.com/WorkflowActions)

# Migrations

🚫

### DIFF
--- a/Rock/Rock.csproj
+++ b/Rock/Rock.csproj
@@ -1905,6 +1905,8 @@
     <Compile Include="Workflow\Action\Utility\InteractionAdd.cs" />
     <Compile Include="Workflow\Action\Utility\CreateShortLink.cs" />
     <Compile Include="Workflow\Action\Utility\LavaRun.cs" />
+    <Compile Include="Workflow\Action\Utility\SetEntityAttribute.cs" />
+    <Compile Include="Workflow\Action\Utility\SetEntityProperty.cs" />
     <Compile Include="Workflow\Action\WorkflowControl\Redirect.cs" />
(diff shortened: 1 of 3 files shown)


edmistj:
I'd say your 'soon' is pretty fast! Thank you!!!

PR #2462, opened 2017-09-14
Change Login Block To Honour Locked & Unconfirmed External Accounts

### LINKED ISSUE #2115: Google Auth - Locked Out - No Error Message
### Prerequisites
*  Put an X between the brackets on this line if you have done all of the following:
    * [ ] Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * [X] Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

When a User logs in using Google Auth, it creates a UserLogin with a name Google_<number>. If we lock this UserLogin and the user tries to login using Google Auth using the same email ID, user is taken back to the Login Page, without any Error message. However, if a DB user is Locked out, there is a message displayed, which is configured on the Login Block.

Unable to reproduce on Demo Site, as Demo site is not setup for Google Auth.

**Expected behavior:** 
The same Locked Out message should be displayed as the DB User.

**Actual behavior:** 
No Error Message is displayed, but the user is taken back to the login screen.

### Versions

* **Rock Version:** [6.2]  
* **Client Culture Setting** [en-US]

arranf:
# Context
Fixes #2115 

# Goal
Change the Login block to check for locked and unconfirmed external accounts

# Strategy
I took the logic for checking for confirmation/locked accounts from the login method from the internal login button event and moved it to a separate method which is called by both the `lbLogin` and the `btnLogin`. The method signature and naming for that method are a bit weird because I'm bad at naming and I wanted to preserve the `UserLoginService` if it was an internal auth.

# Tests
Not sure this is testable but if it is I'll happily write a test.

# Possible Implications
If someone has locked an external auth account previously there would've been no consequences. After this change, they'd be locked out.

# Documentation
🤷‍♂️

### DIFF
--- a/RockWeb/Blocks/Security/Login.ascx.cs
+++ b/RockWeb/Blocks/Security/Login.ascx.cs
@@ -50,10 +50,10 @@ Thank you for logging in, however, we need to confirm the email associated with 
     [CodeEditorField( "Locked Out Caption", "The text (HTML) to display when a user's account has been locked.", CodeEditorMode.Html, CodeEditorTheme.Rock, 100, false, @"
 Sorry, your account has been locked.  Please contact our office at {{ 'Global' | Attribute:'OrganizationPhone' }} or email {{ 'Global' | Attribute:'OrganizationEmail' }} to resolve this.  Thank you. 
 ", "", 5 )]
-    [BooleanField("Hide New Account Option", "Should 'New Account' option be hidden?  For site's that require user to be in a role (Internal Rock Site for example), users shouldn't be able to create their own account.", false, "", 6, "HideNewAccount" )]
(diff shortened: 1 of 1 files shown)


azturner:
Merged this, but also made another update so that userlogin doesn't need to be queried again when using normal db login: https://github.com/SparkDevNetwork/Rock/commit/6ebbd7133cafef32debbd9320e3ff91df535b10e

Issue #2433, opened 2017-09-06
[EN] Adjust Financial Gateway Interface To Know If Transaction is Event or Donation


arranf:
## Description
Rock has a natural divide between payments for events and gifts that are donations and some Financial Gateways would like to process the payments differently depending on which type the payment is.

## Strategy
A relatively non-invasive and non-breaking change would be to add a `PaymentType` enum as a property to the PaymentInfo object to give some context to the gateway.

I'll take the lead on PRing this if you guys agree.

edmistj:
There is a Defined Type called 'Transaction Type' that is used for that purpose. The two default values (Contribution and Events) have system guids. You might have everything you need with this (it's tied to the Financial Transaction model). 

I would note that some churches have more than 2 types. The Transaction Entry block has been extended in several ways to allow for it to act as a 'check-out' function. You'd probably want to have some fallback logic for cases when the Transaction Type isn't Contribution or Event.

arranf:
@edmistj The problem with `Transaction Type`  is that it's only added to the transaction after the payment is processed by the gateway. Perhaps moving be that information to the `PaymentInfo` class would be a better route?

mcnoodles:
This would also be helpful for us. our finance people currently rely on the simple donation dashboard as we're transitioning."

dcs619:
I'm working with an international Rock church that could use this change as well.  

For the gateway to know what type of transaction it's processing, you have to modify core functionality in TransactionEntry, RegistrationEntry, and ScheduledTransactions to add prefix or specific GLCode support.  The PR @arranf submitted would be a much cleaner approach.

nairdo:
@MikeDPeterson  resolved for v10 via https://github.com/SparkDevNetwork/Rock/commit/e8340eaeafe9387a69c1be8d92c8b3490c162ba2

taylorbrooks:
Thank you!!!

PR #2508, opened 2017-10-02
Add TransactionType to PaymentInfo to Make Type Accessible to Gateways

arranf:
## Context
#2433 

## Goal
To make the type of transaction available to gateways

## Strategy
I've added a `int? TransactionTypeId` property to the base `PaymentInfo` class. I've also added code in each block that I could find that creates transactions to populate it.

## Possible Implications
This should be backwards compatible.


mcnoodles:
This would help us tremendously!  Thanks Aaran!

taylorbrooks:
womp womp

nairdo:
@taylorbrooks No... We've had another incident with the sync tool (Unito) we use.  We recently separated our Issues and Pull requests into two syncs and it overreacted.  I'm doing cleanup now. :(

MikeDPeterson:
Merged locally with https://github.com/sparkdevnetwork/rock/commit/e8340eaeafe9387a69c1be8d92c8b3490c162ba2

Issue #2548, opened 2017-10-17
Add Account filters to blocks on Contributions tab


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

We plan to duplicate the `Contributions` tab on the Person Details page but add block settings to limit the displayed accounts to a subset. This will allow us to give limited access to accounting information.

### Versions

* **Rock Version:** 6.9
* **Client Culture Setting:** en-US

edmistj:
Sounds good, so the Account Filters would be block settings?

cabal95:
Yep. `AccountsField` attribute on the block. Value not required.

One additional change that I am making (as I work through this so far) is to make the `Detail Page` for the Pledge List block optional (possibly the other blocks if the same thing come sup). If not set, the rows are not clickable and the Add button does not show up.

edmistj:
Ok sounds good.

PR #2549, opened 2017-10-17
Added Account filter block setting to various Finance blocks

cabal95:
## Contributor Agreement

**Aye Sir!**

## Context

Needed ability to display giving information, limited to a specific set of accounts, to a small group of staff members. Also see #2548. 

## Goal

Add ability for the admin to configure the following blocks to limit results to those that are related to the specified accounts.

* Transaction List
* Scheduled Transaction List
* Pledge List
* Transaction Yearly Summary Lava

## Strategy

New Block setting to, optionally, select multiple accounts. This now lets us create a new page with these blocks on it, security limited to this group of people and accounts limited to just the accounts they can see.

Also a slight tweak to the UI on the Transaction List and Scheduled Transaction List blocks when in this filtered mode. Since those can display single-row results that cover multiple accounts, if the account name to be displayed is not "allowed" then it's name gets changed to `Other`. Screenshot sample below.

## Tests

n/a

## Possible Implications

n/a

## Screenshots

transactionlist

In this case, the original account name was `General Fund` and it got changed to `Other`. If nobody objects we could just not display these values at all, but my concern was then the amounts would not add up to the Amount column that is still visible.

## Documentation

n/a

## Migrations

n/a


azturner:
@cabal95 The TransactionList block also has an account filter that still displays all the accounts even when the block is configured to only show some accounts. Can you update this so that if a block setting is being used to limit the account list that the account filter is hidden.

azturner:
Same with PledgeList and ScheduledTransactionList

cabal95:
@azturner Should be ready.

PR #2560, opened 2017-10-24
Add an optional maximum length to Text field types

### LINKED ISSUE #2488: [EN] Add length limitation to Text FieldType
### Description

This should be a quick PR the next chance I have but wanted to get it here for any discussion.  It would be nice to be able to set a max length on the Text FieldType.

### Steps to Reproduce

1. Create a Person attribute of Text field type
2. Add a core category and Save
3. Go to the attribute and add a wall of Lorem Ipsum text
4. Groan with agony as you didn't want someone's life story stored in that attribute.

**Expected behavior:** 

Be able to set a max length for a text field, as this functionality is supported on all TextBox controls.

**Actual behavior:** 

Not able to set a max length for a text field.  Confirmed functionality not available in pre-alpha.

dcs619:
## Contributor Agreement

✅

## Context

Fix #2488.

## Goal

Adds support for the pre-existing Textbox property of Max Length to the Rock Text fieldtype.

## Screenshots

screenshot 2017-10-23 23 00 34

screenshot 2017-10-23 22 58 16

## Documentation

An attribute of type Text can now be set with a maximum input length.  Keep in mind that the user isn't prompted when they reach the end of the field; they just won't be able to type any more.

## Migrations

N/A

### DIFF
--- a/Rock/Field/Types/TextFieldType.cs
+++ b/Rock/Field/Types/TextFieldType.cs
@@ -66,8 +66,8 @@ namespace Rock.Field.Types
 
             var maxLength = new RockTextBox();
             controls.Add( maxLength );
-            maxLength.Label = "Max Length";
-            maxLength.Help = "Enter the max length of this field (0 is unlimited).";
+            maxLength.Label = "Maximum Length";
+            maxLength.Help = "Enter the maximum length of this field (leave blank for unlimited).";
 
             return controls;
         }
--- a/Rock/Field/Types/TextFieldType.cs
+++ b/Rock/Field/Types/TextFieldType.cs
@@ -30,10 +30,10 @@ namespace Rock.Field.Types
     [Serializable]
     public class TextFieldType : FieldType
     {
-
         #region Configuration
 
         private const string IS_PASSWORD_KEY = "ispassword";
+        private const string MAX_LENGTH = "0";
 
         /// <summary>
         /// Returns a list of the configuration keys
@@ -43,6 +43,7 @@ namespace Rock.Field.Types
         {
             var configKeys = base.ConfigurationKeys();
             configKeys.Add( IS_PASSWORD_KEY );
+            configKeys.Add( MAX_LENGTH );
             return configKeys;
         }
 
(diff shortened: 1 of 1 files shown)


edmistj:
@dcs619 could you make the configuration be 'leave blank for unlimited' vs setting it to 0.

dcs619:
@edmistj the code already ignores Max Length if it's blank or 0, but I'll adjust the help text. Do you want people to be able to set a Max Length of 0?

edmistj:
@dcs619 don't think allowing for 0 is necessary. Thanks for updating the help text.

azturner:
@dcs619 This change is going to cause an issue in the EncryptedTextField type since it is inheriting from the TextFieldType class and adds it's own configuration options (and assumes they are the second/third option). That class should probably be updated so that it does not inherit from TextFieldType and instead inherits from the base FiedlType class. It would then need to add the handling of the IS_PASSWORD option itself rather than getting that functionality from the TextFieldType.  That way you can add the additional option to the TextFieldType.

Rock/Field/Types/TextFieldType.cs
```diff
@@ -74,10 +81,19 @@ public override List<Control> ConfigurationControls()
...
-            { 
-                configurationValues[IS_PASSWORD_KEY].Value = ( (CheckBox)controls[0] ).Checked.ToString();
+            if ( controls != null && controls.Count > 0 )
+            {
+                if ( controls[0] != null && controls[0] is CheckBox )
+                {
+                    configurationValues[IS_PASSWORD_KEY].Value = ( (CheckBox)controls[0] ).Checked.ToString();
+                }
+
+                if ( controls[1] != null && controls[1] is TextBox )
```
azturner: This should also check to make sure that controls.Count > 1

Rock/Field/Types/TextFieldType.cs
```diff
@@ -111,6 +127,11 @@ public override void SetConfigurationValues( List<Control> controls, Dictionary<
                 {
                     ( (CheckBox)controls[0] ).Checked = configurationValues[IS_PASSWORD_KEY].Value.AsBoolean();
                 }
+
+                if ( controls[1] != null && controls[1] is TextBox && configurationValues.ContainsKey( MAX_LENGTH ) )
```
azturner: Same thing here

dcs619:
Fixed in 266b9f0. Closing.

Rock/Field/Types/TextFieldType.cs
dcs619: That was already checked at L86

Rock/Field/Types/TextFieldType.cs
dcs619: Checked at L124

Issue #2552, opened 2017-10-19
SQL Lockup with too many incoming SMS messages


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

Okay, so awhile ago we found an infinite loop with generating check-in security codes. We just found another. There is an infinite loop that happens to generate an SMS Response Code. There are 898 possible code numbers (100 - 999, subtract out `666`). And if all of those have been used in the past 30 days then the `while` loop never exits and SQL starts chewing up CPU cycles like crazy - and those messages never get passed on to the final person.

https://github.com/SparkDevNetwork/Rock/blob/8142b5fc1cbefa1545079df7c13fa6da0204dd17/Rock/Communication/Medium/Sms.cs#L195

I think a good solution would be to forget the random number generator and just go sequential until we find an available number. I'm not sure there is a reason to use a random number in this case.

If we do need random, I would say we use a loop counter and after X (say 2,000?) loops, increase the max random number by 1,000 as well.

This also means there are a few places with ResponseCode that, I think, use a 3-digit regex to find a response code that would need to be modified to look for a 3+ digit code instead.

### Steps to Reproduce

Send 898 message to a Twilio number that sends to a staff member. Then try to send one more.

### Versions

* **Rock Version:** 6.9
* **Client Culture Setting:** en-US

MikeDPeterson:
If it needs to be random, and with no risk of spins, perhaps Rock could pre-generate a List of response codes in memory, and sort them in random order. Then anytime it needs a new one, just take the next one from the list, and start from the beginning each time it gets to the end.  The pre-generated list could be something of reasonable size, perhaps as large as 1 to 9999999, or maybe as small as 100-999. If this solution works, maybe a similar approach could be used for checkin labels

cabal95:
There hasn't been any comments on this about it needing to remain random. So I am going to propose the following algorithm:

[...]

PR #2563, opened 2017-10-30
Reworked SMS Response Code generate to prevent infinite loops.

cabal95:
## Contributor Agreement

**Yes**

## Context

Issue #2552, when all 899 available response codes had been used in the past 30 days the system would loop forever trying to generate a new code.

## Goal

Increase number of available codes, decrease window for a code being considered "used" and rework the loop to have a final escape.

## Strategy

System now supports up to 5-digit response codes. It generates them in a "bucket system" of 100 at a time, between 100 and 90,000 (skipping `666` and `911`). These chunks are generated using the "smallest number available" logic. Once the chunk list is generated it is randomized. Because of this "bucket system" the randomized list will be at most 100 sequential numbers, so we may want to increase the bucket size if we find there is not enough entropy.

This randomized list is then stored in the cache and re-used until the list has been depleted, at which time a new list is generated.

When trying to find the next single available code, it will try 1,000 times to find a code before giving up. An exception is now thrown if that 1,000 limit is reached or no codes are found to be available when generating the cached list.

## Performance

For testing performance, I pre-generated codes 100 - 3,927 as recently used response codes to determine how much time would be used by the system to process incoming SMS messages.

* Took 5,673ms to find recently used codes (3,827 database rows).
* Took 51ms to find first available chunk of codes (i.e. 3,928 - 3,999).
* Took 0ms to randomize list.

These timings only applied on the first (simulated) incoming SMS message. Subsequent messages took <2ms to pull from the cache and check the database to ensure that one single code was still available.

## Tests

n/a

## Possible Implications

n/a

## Screenshots

n/a

## Documentation

n/a

## Migrations

n/a

### DIFF
--- a/Rock/Communication/Medium/Sms.cs
+++ b/Rock/Communication/Medium/Sms.cs
@@ -230,8 +230,7 @@ namespace Rock.Communication.Medium
 
                 if ( availableCodes.Any() )
                 {
-                    availableCodes.OrderBy( c => Guid.NewGuid() ).ToList();
-                    return availableCodes;
+                    return availableCodes.OrderBy( c => Guid.NewGuid() ).ToList();
                 }
             }
 
--- a/Rock/Communication/Medium/Sms.cs
+++ b/Rock/Communication/Medium/Sms.cs
@@ -15,6 +15,7 @@
 // </copyright>
 //
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.ComponentModel.Composition;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


azturner:
@cabal95 In testing/looking at this, it does not appear that the codes are being randomized at all. First two codes I got were @100 and @101. I'm assuming the `availableCodes.OrderBy( c => Guid.NewGuid() ).ToList();` is what you're using to try and randomize them, but that does not seem to do so.

cabal95:
@azturner Hmm. It did work for me. It's possible it was an incredible fluke.  I did test that style of code again just now on https://dotnetfiddle.net and it worked. That was also the recommended method I found on a few stack overflow posts, but I suppose it's possible something with a slightly different .NET version or something. Is it easy enough for you to test again with a debugger breakpoint there to see what is actually in the list after randomization? I'll see if I can test again on my system, but I'm not sure that branch will run against my database anymore.

Code used to test on dotnetfiddle.net:
```
using System;
using System.Collections.Generic;
using System.Linq;
					
public class Program
{
	public static void Main()
	{
		var list = new List<int> { 1, 2, 3, 4, 5 };
		
		Console.WriteLine( string.Join(", ", list.OrderBy( c => Guid.NewGuid() ) ) );
	}
}
```

Rock/Communication/Medium/Sms.cs
```diff
@@ -179,38 +197,92 @@ private void CreateCommunication( int fromPersonAliasId, string fromPersonName,
...
+                    .Where( i => !blacklist.Contains( i ) )
+                    .Select( i => string.Format( "@{0}", i ) )
+                    .Where( c => !activeCodes.Contains( c ) )
+                    .ToList();
+
+                if ( availableCodes.Any() )
+                {
+                    availableCodes.OrderBy( c => Guid.NewGuid() ).ToList();
+                    return availableCodes;
+                }
```
azturner: @cabal95 The orderby actually works, but the reordered list is not saved to a new list. If we just change these two lines to...
`return availableCodes.OrderBy( c => Guid.NewGuid() ).ToList();`
it works fine. I'll just pull this in and make that change.

cabal95: Oh, that will do it. I must have had that in two lines when I tested and then refactored it into two right before committing without retesting. Sorry 'bout that!

azturner: @cabal95 No Prob!  Thanks for doing all these pull requests. It's awesome!

PR #2593, opened 2017-11-13
Added Subtitle (Attribute Id) To Modal Dialog from Attribute Editor

shivambareria:
# Rock PR Policy
1. If you are submitting a PR for a logged Issue / Enchancement request please reference it in your commit
2. If your PR is for an enchancment that has not been discussed and approved by the core team please get that approval BEFORE submitting
the request. In fact, this approval should be recieved before writing the code to limit rework on your part. This will ensure that all code is working into the same vision and direction of the core project.

## Contributor Agreement
Have you filled out and sent your [Spark Contributor Agreement](http://www.rockrms.com/Content/RockExternal/Misc/Contributor%20Agreement.pdf) to secretary [at] sparkdevnetwork.org?
**Yes**

## Context
What is the problem you encountered that lead to you creating this pull request?

## Goal
What will this pull request achieve and how will this fix the problem?
Goal is to put the Attribute Id in Modal Dialog containing attribute editor for convenience.

## Strategy
How have you implemented your solution?
Looked for parent Modal dialog inside Attribute editor control and added subtitle to control if found.

## Tests
If your code is a new method or function (that doesn't need a mock database or SqlServerTypes library) and can be Xunit tested [see example](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests/Rock/Lava/RockFiltersTests.cs) be sure your pull request includes the corresponding unit tests in the Rock.Tests project. In all cases *you* MUST test your code before submitting a pull request.
N/A

## Possible Implications
What could this change potentially impact? Are there any security considerations? Where could this potentially affect backwards compatibility?
N/A

## Screenshots
Provide us some screenshots if your pull request either alters existing UI or provides new UI. Arrows and labels are helpful.
N/A

## Documentation
If your change effects the UI or needs to be documented in one of the existing [user guides](http://www.rockrms.com/Learn/Documentation), please provide the brief write-up here:
N/A

## Migrations
N/A

### DIFF
--- a/Rock/Web/UI/Controls/AttributeEditor.cs
+++ b/Rock/Web/UI/Controls/AttributeEditor.cs
@@ -1367,6 +1367,8 @@ namespace Rock.Web.UI.Controls
 
                 this.Qualifiers = qualifiers;
                 this.DefaultValue = attribute.DefaultValue;
+
+                SetSubTitleOnModal( attribute );
             }
 
             if ( objectType != null )
@@ -1482,6 +1484,24 @@ namespace Rock.Web.UI.Controls
 
                     }
                 }
+            }
+        }
+
+        /// <summary>
+        /// Set the Subtitle of modal dialog
+        /// </summary>
+        /// <param name="attribute">The attribute.</param>
+        protected void SetSubTitleOnModal( Model.Attribute attribute )
+        {
+            Control parent = this.Parent;
+            while ( parent.GetType() != typeof( ModalDialog ) )
+            {
+                parent = parent.Parent;
+            }
+            if ( parent.GetType() == typeof( ModalDialog ) )
+            {
+                ModalDialog modalDialog = ( ModalDialog ) parent;
+                modalDialog.SubTitle = string.Format( "Id: {0}", attribute.Id );
             }
         }
 
--- a/Rock/Web/UI/Controls/AttributeEditor.cs
+++ b/Rock/Web/UI/Controls/AttributeEditor.cs
@@ -1113,7 +1113,7 @@ namespace Rock.Web.UI.Controls
                 if ( entityType != null )
                 {
                     _cbIsAnalytic.Visible = entityType.IsAnalyticAttributesSupported( this.AttributeEntityTypeQualifierColumn, this.AttributeEntityTypeQualifierValue );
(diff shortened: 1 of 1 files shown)


Rock/Web/UI/Controls/AttributeEditor.cs
```diff
@@ -1486,6 +1488,24 @@ protected void CreateFieldTypeDetailControls( int? fieldTypeId, bool recreate =
         }
 
         /// <summary>
+        /// Set the Subtitle of modal dialog
+        /// </summary>
+        /// <param name="attribute">The attribute.</param>
+        protected void SetSubTitleOnModal( Model.Attribute attribute )
+        {
+            Control parent = this.Parent;
+            while ( parent.GetType() != typeof( ModalDialog ) )
```
azturner: @shivambareria Looks like this would cause an exception if the attribute editor was ever used without a parent modaldialog (parent would eventually be null).

There's an extension method that you can use instead for this: FirstParentControlOfType<T>.  That takes care of checking for null.

Rock/Web/UI/Controls/AttributeEditor.cs
```diff
@@ -1486,6 +1488,24 @@ protected void CreateFieldTypeDetailControls( int? fieldTypeId, bool recreate =
...
+            {
+                parent = parent.Parent;
+            }
+            if ( parent.GetType() == typeof( ModalDialog ) )
+            {
+                ModalDialog modalDialog = ( ModalDialog ) parent;
+                modalDialog.SubTitle = string.Format( "Id: {0}", attribute.Id );
+            }
```
azturner: Might only set subtitle too if it's blank (in case a block is using subtitle already for something else).

Rock/Web/UI/Controls/AttributeEditor.cs
```diff
@@ -1486,6 +1488,20 @@ protected void CreateFieldTypeDetailControls( int? fieldTypeId, bool recreate =
         }
 
         /// <summary>
+        /// Set the Subtitle of modal dialog
+        /// </summary>
+        /// <param name="attribute">The attribute.</param>
+        protected void SetSubTitleOnModal( Model.Attribute attribute )
+        {
+            Control parent = this.Parent;
```
azturner: Is this line still needed?

PR #2594, opened 2017-11-13
Update Lava Webhook endpoints to not fail when DefinedValue does not start with /

cabal95:
## Contributor Agreement

**Yes**

## Context

If a user defines the Value of a Lava Webhook endpoint as just `vcard` instead of `/vcard` then it will never be matched, as the code expects all endpoint values to begin with `/`.

## Goal

Make the beginning `/` optional.

## Strategy

If the DefinedValue value does not start with `/` or `^` then prepend '/' to the URL string.

## Tests

n/a

## Possible Implications

n/a

## Screenshots

n/a

## Documentation

n/a

## Migrations

n/a

### DIFF
--- a/RockWeb/Webhooks/Lava.ashx
+++ b/RockWeb/Webhooks/Lava.ashx
@@ -109,6 +109,11 @@ public class Lava : IHttpHandler
                 string apiMethod = api.GetAttributeValue( "Method" );
                 List<string> variables = new List<string>();
 
+                if ( !apiUrl.StartsWith( "/" ) )
+                {
+                    apiUrl = "/" + apiUrl;
+                }
+
                 //
                 // Check for match on method type, if not continue to the next item.
                 //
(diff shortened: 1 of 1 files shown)


RockWeb/Webhooks/Lava.ashx
```diff
@@ -109,6 +109,11 @@ public class Lava : IHttpHandler
                 string apiMethod = api.GetAttributeValue( "Method" );
                 List<string> variables = new List<string>();
 
+                if ( !apiUrl.StartsWith( "/" ) && !apiUrl.StartsWith( "^" ) )
```
azturner: @cabal95 Wouldn't this be better to add after the check below for apiUrl being null or empty?

cabal95: Yes... Yes i would. :) Done.

Issue #2427, opened 2017-09-04
Cannot save default attribute values for tiered field types.


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

When trying to edit an attribute with a tiered field type (such as `Group Type Group`), you cannot save the default value. It always returns to a blank state. This appears to be an issue with the `AttributeEditor` not allowing the underlying Field/Control to fully initialize before requesting the current value.

### Steps to Reproduce

1. Create a Global Attribute called `zztest` and set it's type to `Group Type Group` (or any other tiered field type where you select one value and then can select a second value).
2. Save the attribute. Now click the row to edit the actual value and save.
3. Notice the value does save.
4. Edit the attribute with the pencil and set a default value and save.
5. Edit the attribute again and notice the default value has not saved.
6. Waste 2 hours of your life trying to figure out what you did wrong when creating a custom field type before testing an existing one and realizing it isn't you.

**Expected behavior:** 

Should be able to save default values.

**Actual behavior:** 

The world explodes and all life comes to an end. The auto-save point is restored but since the bug is still in place you end up in a permanent loop of death and destruction.

### Versions

* **Rock Version:** 6.3, 6.8, prealpha
* **Client Culture Setting:** en-US

Issue #1687, opened 2016-08-09
"Group and Role" Workflow Attribute Type Not Editable


mikejed:
Having created a trigger to create a particular workflow type, I created a workflow attribute and ran the "Set attribute from Entity" action as the only action in the workflow to see what is passed when the workflow is triggered from a "Group Member Workflow". Since Group Role didn't appear to capture the data (though if I print the GUID stored in the attribute and search for that, I do find it in the [GroupRole] table), I changed the attribute to type "Group And Role".

Immediately upon saving the workflow type, the existing workflows disappear from the list of that workflow type.

Upon editing the workflow again and clicking the "Edit" button on the attribute, the following exception is thrown:

```
Error
The given key was not present in the dictionary.
   at Rock.Field.Types.GroupAndRoleFieldType.EditControl(Dictionary`2 configurationValues, String id)
   at Rock.Web.UI.Controls.AttributeEditor.CreateFieldTypeDetailControls(Nullable`1 fieldTypeId, Boolean recreate)
   at Rock.Web.UI.Controls.AttributeEditor.set_FieldTypeId(Nullable`1 value)
   at Rock.Web.UI.Controls.AttributeEditor.SetAttributeProperties(Attribute attribute, Type objectType)
   at RockWeb.Blocks.WorkFlow.WorkflowTypeDetail.ShowAttributeEdit(Guid attributeGuid)
   at RockWeb.Blocks.WorkFlow.WorkflowTypeDetail.gAttributes_Edit(Object sender, RowEventArgs e)
   at System.Web.UI.WebControls.LinkButton.RaisePostBackEvent(String eventArgument)
   at System.Web.UI.Page.ProcessRequestMain(Boolean includeStagesBeforeAsyncPoint, Boolean includeStagesAfterAsyncPoint)
```

ccofnow:
Reproduction:
http://rock.rocksolidchurchdemo.com/page/136?workflowTypeId=2016
1) Created a new workflow type with a single "Set attribute from Entity" action, single attribute. In this case, a person attribute to start
2) Allowed the workflow type to be launched from a person profile
3) Launched the workflow from a person profile
4) Verified the person showed up in the workflow grid
5) Edited the workflow attribute to be of type "Group"
6) Verified the workflow grid showed the person alias GUID instead of the person name (since the GUID does not represent a valid group)
7) Edited the workflow attribute to be of type "Group and role"
*8) Note that the workflow grid is now empty
[...]

PR #2595, opened 2017-11-14
Fixed Attribute Editor to allow editing tiered default values

cabal95:
## Contributor Agreement

**Yes**

## Context

See issue #2427 and #1687.

AttributeEditor would not allow you to edit default value of a tiered attribute such as Group Type and Group. Or rather, you could edit but it would not save the value correctly.

Also found two places where a key was access without being checked for existence which was causing errors when changing attribute types.

## Goal

Fix the bugs, make the attribute editor work.

## Strategy

Implemented `ContainsKey` checks and turned on view state for the default value placeholder.

@azturner originally wrote this code so mentioning him, even though it is 4 years old, just in case he remembers a reason why he originally set it that way. With ViewState disabled on the default value placeholder, it also disabled it on the child controls so the viewstate was not retained and thus current selections could not be determined correctly.

Edited a few other attribute types and no problems were found.

## Tests

n/a

## Possible Implications

n/a

## Screenshots

n/a

## Documentation

n/a

## Migrations

n/a


cabal95:
Note: this fix might fix an issue I had trying to build a plug-in related to attributes, so once it's accepted if a note could be added indicating which version it will first be available in that would be great. In the mean time I'll work on testing if this does fix other problem I ran into.

cabal95:
Marking this as In Progress. Have one more commit incoming. Fell down the rabbit hole on this and worked out the other viewstate related changes needed to fix ViewState with configuration controls as well. This allows (among other things) embedding fields within fields. Everything seems to be working, just need to run it through some more tests on existing field types to make sure nothing has broken.

Sample of embedded fields below is essentially an array field type. I need to work on the actual field type (not part of this PR) a bit more, but the basic idea is being able to select "child" field types and allow for configuring them using their native configuration controls.

attribute_demo

cabal95:
Okay this is ready and tested.  I will note that since this grew a bit more than the original scope I can go back and break it into two different commits if that would be better. One for the field type changes and one for the attribute editor fixes - while it is fixing a bug still, it's also certainly pushing the envelope towards enhancement status since nobody but me has run into the bug yet.

cabal95:
Thanks guys, this should allow for some pretty cool new attribute types I was unable to figure out how to implement with the previous ViewState issues.

PR #2626, opened 2017-12-14
Fixed bug that would throw an error if an email subject was too long to be stored in the database.

chrismcgrath:
## Contributor Agreement

Yes

## Context

Users can enter email subjects of any length, but if the subject is longer than the database column max (currently 100), Rock will throw an error when trying to send.

## Goal

To prevent throwing an error.

## Strategy

Get the value of the MaxLength attribute on the database column and only take that many characters for the subject.

## Possible Implications

A more ideal solution would be for the UI to prevent the too-long subject and give the user the chance to edit the subject to fit within the length limit. But since I don't know how to do that, and 100+ character subjects should be quite rare, this is what we have implemented for now.

### DIFF
--- a/Rock/Model/CommunicationService.Partial.cs
+++ b/Rock/Model/CommunicationService.Partial.cs
@@ -16,6 +16,7 @@
 //
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using Rock.Communication;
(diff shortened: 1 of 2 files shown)


azturner:
Added an extension method for this and added additional fields to check: https://github.com/SparkDevNetwork/Rock/commit/98f8d0782696b2cc657772f65237454a9c07c776

PR #2634, opened 2017-12-21
- Updated Rock Store to only show Icon Safe area of plugin image

cabal95:
## Contributor Agreement

**Yes**

## Context

While "Icon Safe Area" is not exactly a correct statement, I believe it conveys the idea. Basically we are recommending developers put an icon in the "red section" of the package banner (420x210 centered) and other information, such as possible text, outside of that "safe area". However on wider monitors, such as 1920x1080, the browser window is wide enough that the package icon displays outside this "safe area" and will display text.

While we don't have a good way yet to have a dedicated icon, this at least provides a way to ensure that the area outside the shaded red will not be displayed in "icon list" view modes without changing the height of the icon itself.

I'm perfectly okay to have this PR rejected since I didn't go through the normal issue-post first for discussion. But since it only ended up being 3 lines of CSS code it felt like it would be faster/easier to just post this with screenshot samples.

## Goal

Make the package icons display a little bit cleaner on wide browser windows.

## Strategy

Apply a max-width to the package summary div and auto margins so it will be centered in the column.

## Tests

n/a

## Possible Implications

n/a

## Screenshots

Screenshots were taken on a 1920x1080 monitor with browser window maximized.

Before:

screenshot-2017-12-21 rock shop rock rms

After:

screenshot-2017-12-21 rock shop rock rms 1

## Documentation

n/a

## Migrations

n/a

### DIFF
--- a/RockWeb/Styles/_store.less
+++ b/RockWeb/Styles/_store.less
@@ -54,6 +54,9 @@
   cursor: pointer;
   background-color: #fff;
   border: 1px solid #ccc;
+  max-width: 282px;
+  margin-left: auto;
+  margin-right: auto;
 
   .packagesummary-image-hover {
     width: 100%;


cabal95:
Forgot to mention, tested on Safari, Firefox and Edge and all appear to be fine.

Issue #2576, opened 2017-11-06
Reporting Dashboard Widgets do not support CampusScheduleContextSetter block


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

When using the Dashboard Widgets with a metric that has a campus and schedule partition, you can only filter by the first partition. So only one of the selectors offered by the CampusScheduleContextSetter block will be honored (usually the campus).

### Steps to Reproduce

1. Create metric with first partition of Campus, second partition of schedule.
2. Create a few metric values across both partitions.
3. Add the Line Chart Dashboard Widget to a page and configure for using context.
4. Add Campus Schedule Context Setter to the same page.
5. Try to set context on campus.
6. Try to set context on schedule.

**Expected behavior:** 

Both should work.

**Actual behavior:** 

Only campus works.

### Suggested Fix

The `LineBarPointsChartDashboardWidget` (boy that's a mouthful) class has a method called `GetPrimaryMetricPartitionEntityIdFromContext` (another mouthful) that gets the filtering information for the first partition found. Since this is a `private` method, I propose this be modified to return an array of results rather than just the first found partition. It is called from two places, one that specifically wants the first match (easily done with `FirstOrDefault` on the returned array) and the other that would want all partitions found in context.

### Versions

* **Rock Version:** 6.8
* **Client Culture Setting:** en-US

PR #2638, opened 2017-12-23
Updated dashboard widget code to support multiple partition contexts

cabal95:
## Contributor Agreement

**Oh yeah**

## Context

See issue #2576. Dashboard Reporting Widget blocks would only get one partition context from the current page even if more were available.

Aaaaaand i screwed up the commit message by not following the proper commit grammar. Sigh.

## Goal

Update base class to support multiple partitions.

## Strategy

All partition context information is returned instead of just the first one. Calling methods (only 2) decide if they want the first one or all of them.

## Tests

n/a

## Possible Implications

Users could suddenly see the data they originally wanted to see but never could.

## Screenshots

n/a

## Documentation

n/a

## Migrations

n/a

### DIFF
--- a/Rock/Reporting/Dashboard/LineBarPointsChartDashboardWidget.cs
+++ b/Rock/Reporting/Dashboard/LineBarPointsChartDashboardWidget.cs
@@ -486,8 +486,8 @@ namespace Rock.Reporting.Dashboard
                 int? result = null;
                 if ( GetEntityFromContextEnabled )
                 {
-                    MetricPartitionEntityId primaryMetricPartitionEntityId = GetPrimaryMetricPartitionEntityIdFromContext();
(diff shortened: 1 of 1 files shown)


edmistj:
@cabal95 in the future it might be a good idea to wait until the source issue is marked 'Available'. That will be the label that says the idea is approved. Just want to eliminate any time spent writing code that wasn't in the same direction as core. This one is good, but do realize that the dashboard widgets are not 100% ready for prime time (as noted in the docs). We're eventually looking to move the charts to Chart.js (away from Flot).

cabal95:
I'll keep it in my mind, but I'm also usually not too worried about "wasted" time for small patches like this. I needed to write this functionality into our forked blocks anyway so we could make use of them with our multi-partitioned metrics, so it wasn't much more than copy and pasting from our code into the DLL methods.

PR #2657, opened 2018-01-10
Remove dollar sign (character) from fees

jreciomoreno:
If a dollar sign is saved as part of the fee option (Id / Name), it causes the fee not to work during and event registration, this because the fee control id is malformed with an invalid dollar character that causes the find control logic not to find and charge the selected fee during registration.

# Rock PR Policy
1. If you are submitting a PR for a logged Issue / Enchancement request please reference it in your commit
2. If your PR is for an enchancment that has not been discussed and approved by the core team please get that approval BEFORE submitting
the request. In fact, this approval should be recieved before writing the code to limit rework on your part. This will ensure that all code is working into the same vision and direction of the core project.

## Contributor Agreement

## Context
Fees that have a dollar sign as part of the option Id didn't work on event registration.

## Goal
Added logic to remove dollar sign from fees in order to avoid this issue.

## Strategy
Yes

## Tests

## Possible Implications
Don't think so, it will only fix the fee id's that have dollar signs when these are edited.

## Screenshots

## Documentation

## Migrations

### DIFF
--- a/RockWeb/Blocks/Event/RegistrationTemplateDetail.ascx.cs
+++ b/RockWeb/Blocks/Event/RegistrationTemplateDetail.ascx.cs
@@ -1856,11 +1856,11 @@ namespace RockWeb.Blocks.Event
             fee.FeeType = rblFeeType.SelectedValueAsEnum<RegistrationFeeType>();
             if ( fee.FeeType == RegistrationFeeType.Single )
             {
-                fee.CostValue = cCost.Text;
+                fee.CostValue = cCost.Text.Replace("$", "");
             }
             else
             {
-                fee.CostValue = kvlMultipleFees.Value;
+                fee.CostValue = kvlMultipleFees.Value.Replace("$", "");
             }
             fee.AllowMultiple = cbAllowMultiple.Checked;
             fee.DiscountApplies = cbDiscountApplies.Checked;


cabal95:
It seems the NumberBox is allowing non-numeric characters, which I think is a bug. This might be best fixed by fixing the NumberBox to provide validation for numeric only characters `0-9,.`.

screen shot 2018-01-10 at 8 57 07 am

Verified the same thing happens on a generic Attribute of type Currency.

jreciomoreno:
It does validate for us when we hit the save button.

What I was referring to is when the Options is set to Multiple.

[cid:2C61629A-64B7-49A2-ABDF-FB8D39E67879@willowcreek.org]
When the “Option” under Costs has a dollar sign, this gets saved as part of the fee id causing the event registration page not to find the selected fee.

[cid:42522242-63B5-49F0-B1B6-8C3433429CE4@willowcreek.org]

On Jan 10, 2018, at 11:00 AM, Daniel Hazelbaker <notifications@github.com<mailto:notifications@github.com>> wrote:


It seems the NumberBox is allowing non-numeric characters, which I think is a bug. This might be best fixed by fixing the NumberBox to provide validation for numeric only characters 0-9,..

[screen shot 2018-01-10 at 8 57 07 am]<https://user-images.githubusercontent.com/1119863/34784950-8e4fa98c-f5e4-11e7-8c45-888b88ee5633.png>

Verified the same thing happens on a generic Attribute of type Currency.

cabal95:
Got ya, yes for multi this is probably needed, I wonder if we need to strip all non-numeric characters to handle non-US currencies?

Oddly enough, you said it does validate for you on single, but it doesn't validate for me on single. I entered just `asdf` for the fee amount and it happily saved that. 🙄

jreciomoreno:
Hmm… I tried initially removing all non alpha-numeric characters from it but it didn’t work, I think is because on the backend the caret (^) character  is needed to form the key value pair. All other characters shouldn’t cause an issue if are used as the key for the fee, is just the $ that causes an error when trying to find the fee control by id on the event registration block.

For the validation, it works for us, its  a currency box, is it using the same control type in your block?


On Jan 10, 2018, at 12:50 PM, Daniel Hazelbaker <notifications@github.com<mailto:notifications@github.com>> wrote:


Got ya, yes for multi this is probably needed, I wonder if we need to strip all non-numeric characters to handle non-US currencies?

Oddly enough, you said it does validate for you on single, but it doesn't validate for me on single. I entered just asdf for the fee amount and it happily saved that. 🙄

Issue #2642, opened 2017-12-28
Add support for SQL User Defined Functions in plugins


cabal95:
### Description

Rock's data model supports User Defined Functions defined in SQL to be used in .Where() queries. However, there is currently no way (that I can find) for plugins to register their own user defined functions to be used in their own .Where() queries.

Currently, these are defined by manually calling the internal classes that define these functions during the OnModelCreating. Suggest that we instead use reflection to find all classes that implement `IStoreModelConvention<EdmModel>` and add instantiate them and add them to the data model.

### Versions

* **Rock Version:** 6.8
* **Client Culture Setting:** en-US

edmistj:
I added this to the developer discuss (for our next meeting). What is the intent of the question? Are you looking for a green light to do a PR, or are you requesting that we make the change? Just want to make sure I understand the ask.

cabal95:
It seems like a pretty trivial (in terms of code) addition so I would probably do it myself unless you wanted to do it internally. In my particular use case I still have some other things to iron out even once that is in, but seemed like a potentially useful item for future plugins whether I end up using it for this particular project or not.

Biggest concern would be accidentally adding in some `IStoreModelConvention<EdmModel>` from random DLLs that shouldn't be, so maybe a new Rock interface would be needed to tag them for inclusion. So if you green light the concept I might ping back with a few implementation follow up ideas just to make sure it's well thought out for future uses.

edmistj:
This is good to go, but could you use `IRockStoreModelConvention<EdmModel>`.

PR #2686, opened 2018-01-19
- Added support for plugins to use custom SQL Functions in LINQ queries

cabal95:
## Contributor Agreement

## Context

As a plugin developer, could not use custom made SQL Functions with my LINQ queries. Also see #2642.

## Goal

Provide a way to do this.

## Strategy

Added new interface `IRockStoreModelConvention<T>` which inherits from `IStoreModelConvention<T>` and use reflection to add all matching class instances to the configuration.

## Tests

* Verified that the 3 existing implementations are loaded.
* Tested the ufnCrm_GetSpousePersonIdFromPersonId function as used by the DataView transformation to verify that existing implementations still work.
* Tested adding a new user defined function via a plugin works.

## Possible Implications

## Screenshots

## Documentation

## Migrations

### DIFF
--- a/Rock/Data/IRockStoreModelConvention.cs
+++ b/Rock/Data/IRockStoreModelConvention.cs
@@ -1,0 +1,26 @@
+﻿// <copyright>
+// Copyright by the Spark Development Network
+//
+// Licensed under the Rock Community License (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.rockrms.com/license
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


MikeDPeterson:
@cabal95 very nice! Looks good

nairdo:
@cabal95 is there anything you think should be mentioned or explained in the dev docs or plugin packaging guides as related to this change?

cabal95:
@nairdo Nothing exotic. It should "just work" the same way as if you included a Data View filter or Select or other things like that. It could probably use a mention somewhere (no idea where) that it is possible to register your own custom SQL functions and to look at the built-in ones for examples. I would hope if a dev is going so deep as to need a custom SQL function that they would have a pretty good idea of how to implement most of it already.

Issue #2607, opened 2017-11-28
Google reCAPTCHA support


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

I was looking at using Google's reCAPTCHA API for a side project. Upon looking at it, it seems like it is about as simple as can be and would be extremely simple to include in core with a small web control that includes the JS file, renders the "captcha" and validates the captcha when requested by the containing Block. The validation is a simple API request.

Possible use cases would be optional settings for User Entry Forms, Prayer Requests, Email Forms, possibly others. If the captcha is enabled, it would likely only make sense to _actually_ show it if there is not a logged in user.

While spam is not a major issue yet for most of us, I'd rather stay ahead of the curve and have something ready to implement in various blocks as the need arises.

Since this is something I have a use-case for, I'm happy to develop the pull request. Knowing it would probably not be widely available until v8 I would likely duplicate the code into my own project until it reaches core and then switch the functionality over to core code.

#### General Implementation Goals

1. API Key configuration (Either two global attributes or dedicated block)
2. Rock.Web.UI.Controls.ReCaptcha control provide the validation services.

> Note: Google reCaptcha is free, but requires you to activate it and provide two API keys to the implementation. It is similar in concept to Google Maps usage, but in my opinion easier to setup. Documentation would need to be written for how to set it up, which might make the dedicated block for configuration a better choice as it can (probably) verify that the API keys are valid and also have a link to the documentation for how to configure in Google. I can write a general overview of the configuration and include it if Amy or somebody else can do final formatting and screenshots.

### Versions

* **Rock Version:** git
* **Client Culture Setting:** en-US

edmistj:
Sounds great. Creating the key seems pretty easy. I'm torn on the new configuration page. It would make it easier, but would add a new page to the settings that are already getting a bit long. We have plans to start grouping the child pages on each page so maybe that solves that.

cabal95:
[...]

PR #2692, opened 2018-01-22
New developer control to provide a Captcha to validate a real person

cabal95:
## Contributor Agreement

## Context

See #2607. Wanted a central control for providing a Captcha to developers.

## Goal

Make a control that developers can use in their blocks.

## Strategy

Implemented new control using Google's reCaptcha system. Chose generic term `Captcha` as the control name so that if we ever change providers we don't have to change the control interface (hopefully).

### Developer information

Control can be used either `Required="true"` or `Required="false"`. When not required it will allow the user to submit the form, but will respond `false` to the `IsResponseValid()` method. This allows blocks to optionally use the control when it isn't strictly required but could help. For example, in a Prayer block it might be optional, but if used would automatically mark the request as approved.

The `IsResponseValid()` method ensures the response is valid and must be used. If Google detects a large amount of non-verified responses it will put a warning on the user's reCaptcha administration console.

It is recommended that all uses of the Captcha control be optional. This is because the administrator must manually enable an account with Google for it's use and as such we cannot be certain that they have done so. And they may choose not to altogether.

@edmistj My last paragraph brings up a question. If the control is used, and the admin has not configured the global attributes, should an error occur (as it would now) or should the control silently mark the user response as valid? That would really only be useful if a developer decided to "hard-code" enable a Captcha control, rendering the block useless to a user who did not have Captcha configured.

### Example Usage

```
<Rock:Captcha ID="cpCaptcha" runat="server" Required="false" Label="Verification" />
```
```
var isValid = cpCaptcha.IsResponseValid();
```

You can also specify the `SiteKey` and `SecretKey` in the Captcha declaration, or in code, if needed for some reason.

## Tests

## Possible Implications

Spam might decrease.

## Screenshots

Screenshots show the "Required=True" state of the control in both an invalid and valid state.

2018-01-22 14_13_19-virtualbox

<hr />

2018-01-22 14_13_53-virtualbox

## Documentation

We need documentation written up on how to register for reCaptcha use. It is pretty straight forward, the summary of steps is below:

* https://www.google.com/recaptcha/admin
* Register a new site (unless you already have one)
* Label = "RockRMS"
* Style = reCAPTCHA V2
* Domains = top level domains (e.g. rocksolidchurchdemo.com)

I learned something helpful while doing this. It is top level domains. So if you put in rocksolidchurchdemo.com, you do not also need to put in www.rocksolidchurchdemo.com. It acts as a wildcard for all child domains. This makes configuration even easier.

## Migrations

A migration is included (though not in the project file) to create the 2 Global Attributes to store the Site Key and Secret Key.


edmistj:
@cabal95 could `.IsResponseValid()` return a bool? where null is returned when not enabled. That would let the developer decide what to do.

cabal95:
That would be doable. In such a case if either SiteKey or SecretKey are empty then the control would not render (treated as if Visible=false) and `.IsResponseValid()` would return null.

I also just realized I didn't complete my original thought on this. The downside to something like this is less information for the admin when trying to figure out "why the bloody captcha isn't showing" and not realizing it's because they haven't put in the SiteKey or SecretKey, since no visible error would show. e.g. admin enables `Use Captcha` in the block settings, clicks save, and then nothing changes. No captcha is shown vs an error stating it's mis-configured.

cabal95:
Another possible take on this that might be better, is a new method of `.IsAvailable()` on the control that returns a `bool` indicating if the control is available (i.e. has the sitekey and secretkey). This would allow the developer to device what to do _before_ rendering the page.

edmistj:
@cabal95  that works

cabal95:
All set. Realized there was no reason to make it a method so there is a `.IsAvailable` property now.

edmistj:
@cabal95 could you write up some documentation for this feature. Amy can touch it up a bit, but at this point I'm not sure we'd even know where to start.

cabal95:
@edmistj Will do. Should I update the `UI Toolkit` wiki page as well?

edmistj:
@cabal95 that would be nice, thanks!

cabal95:
Wiki updated and documentation sent to Amy.

azturner:
Thanks for documenting this @cabal95 . We have standardized on prefixing all future core attributes with 'core_' so I updated these to reflect that: https://github.com/SparkDevNetwork/Rock/commit/6c2a887329684005473d8b46ece355c93f9520e1

cabal95:
👍 I thought that was the case but looked at another recent migration and didn't notice the core_, but probably looked at the one that didn't get the memo. :)

PR #2722, opened 2018-01-31
- Fix grades not working outside of en-US culture

arranf:
Separate ParsingGradeTransition logic
Add GlobalAttributeCache property "CurrentGraduationDate", replace all uses of GlobalAttributeCache.GetValue("GradeTransitionDate") with a GlobalAttributeCache property

## Context
Rock has historically not displayed grades correctly in client cultures that have a different format different from `MM/dd/yyyy` such as en-GB. In #2317 I fixed the `GlobalAttributeCache.CurrentGraduationDate ` to parse the `GradeTransitionDate` correctly however there are other places that manually read the `GradeTransitionDate` to calculate grade offsets - i.e. the GradeFilter, GradePicker, and a PersonProperty. This behaviour became apparent as the year rolled over.

## Goal
 This PR ensures that the grade offset is calculated independent of client culture in places I missed in #2317. 

## Strategy
Much of the code that was affected seemed to be repeating the pattern of calculating whether the transition date is this year or next year so I've added a  `CurrentGraduationDate` property to the GlobalAttributesCache which represents the date of the next graduation date. This simplifies the code elsewhere by a fairly significant margin.

## Areas In Need of Refactor
`GradePicker.GetJavascriptForYearPicker()` can be simplified as a result of the `CurrentGraduationDate` property.

## Possible Implications
This PR is compatible with the v6 branch.

### DIFF
--- a/Rock/Model/Person.cs
+++ b/Rock/Model/Person.cs
@@ -2293,12 +2293,7 @@ namespace Rock.Model
         {
             if (gradeOffset.HasValue && gradeOffset.Value >= 0)
             {
-                var globalAttributes = GlobalAttributesCache.Read();
-                var transitionDate = globalAttributes.GetValue("GradeTransitionDate").AsDateTime() ?? new DateTime( RockDateTime.Today.Year, 6, 1 );
-                transitionDate = new DateTime( RockDateTime.Today.Year, transitionDate.Month, transitionDate.Day );
-
-                int gradeOffsetAdjustment = (RockDateTime.Now.Date < transitionDate) ? gradeOffset.Value : gradeOffset.Value + 1;
-                return transitionDate.Year + gradeOffsetAdjustment;
+                return GlobalAttributesCache.Read().CurrentGraduationYear + gradeOffset.Value;
             }
 
             return null;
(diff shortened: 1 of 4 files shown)


edmistj:
Instead of writing logic for a specific global attribute we're thinking we should make a new field type to store month/day and it could then format correctly. What do you think?

arranf:
@edmistj I think that works well long term! Short term I think this needs a hotfix as if you're a UK church your grades are probably broken.

MikeDPeterson:
@arranf We ended up doing much of what your pull request does, but also updated the GradeTransitionDate to a Month/Day field type.  So, we'll close this pull request since much of it ended up in https://github.com/SparkDevNetwork/Rock/commit/e6f3f8f7f658a59b310e454e53a1bdadbce089a2 (but with some changes)

lpeterson1:
I'm not sure this is the best place to comment on this issue -- I know it's been through the wringer, but on both 7.4 and 8.0 prealpha, if "today" is the grade transition day, editing a person's grade results in an incorrect grade. If you move them a grade forward, the grad year stays the same, so their calculated grade doesn't change. Just giving a heads up -- I don't think this is a high priority.

PR #2744, opened 2018-02-06
Email template preview images delete fix.

trydyingtolive:
## Context
Email template preview images keep disappearing from our list of templates. We found that was because the images are being marked as temporary and being removed.

## Goal
To not have to upload our preview images anymore.

## Strategy
On save, if the image changes delete the previous image (if exists) and set the new image as not temporary.

## Tests
N/A

## Possible Implications
None.

## Screenshots
N/A

## Documentation
N/A

## Migrations
N/A


MikeDPeterson:
@trydyingtolive Looks good except that communications already created with the template will need the previous Logo Image, so it'll probably have to stick around forever. Other than that, this is great fix.  If you agree, I can go ahead and make that small tweak when I merge it, unless you want to.  We want to get this into 7.3 today, so I don't mind making that small change myself

trydyingtolive:
It won't hurt my feelings. Go right ahead.

PR #2762, opened 2018-02-13
New workflow actions for managing registrations

trydyingtolive:
## Context
We needed some workflow actions to help us automatically manage our registrations.

First we needed a way to let people fill out a form to request financial aid for camps. We wanted people to fill out a form and get a discount code so they could register for the camp while their form was being processed. This means they needed to both get a discount code and their discount needs to be increased or decreased after review.

Additionally we needed a way to handle early bird discounts without working at 12AM on a Saturday.

## Goal
This pull request adds three new actions to manage registrations. Adding the ability to create discount codes, update discount codes and change the cost of a registration.

## Strategy
Generate Discount Code creates a new discount code by generating a GUID and using only the appropriate number of characters. It saves that discount code to a workflow attribute for display later.

Update Discount Code takes an existing discount code and updates the value of that code with the option to retroactively update existing discounts.

Update Registration Template Cost sets the cost and minimum initial cost on a registration template only. This is to simplify the workflow action. 

## Tests
We are already running some of these in our production against a large registration. 

## Possible Implications
The biggest security implication is that someone who cannot change registration costs but can change registration, though if someone can do workflow they already have a lot of power.


edmistj:
@trydyingtolive We're going to pull in the first two actions. The action to 'Update Registration Template Cost ' I don't think will be needed as someone is working on a PR to allow Discounts to be applied automatically. This will provide an automated 'Early Bird Discount' feature.

PR #2773, opened 2018-02-15
Slingshot: Fix EF error on location import

dcs619:
## Context

See https://github.com/SparkDevNetwork/Slingshot/issues/9.  The code between `LocationImport` and `PersonAddressImport` wasn't consistent with truncating Street1 and Street2 to 100 characters.  This change makes that consistent.

Ideally the LocationImport and PersonAddressImport should use the same logic, but refactoring doesn't fit in my timeline 😸


MikeDPeterson:
@dcs619  Looks good. Thanks for fixing this!

Issue #2695, opened 2018-01-24
Delete button not really disabled in File Manager


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

In the CMS File Manager (also same issue in the embedded on in the HTML editors), when you have the root `Content` folder selected the delete button is grayed out. However, if you click on the disabled button anyway, it will still activate and ask you if you want to delete. Clicking yes will then produce an error about not being able to delete the folder (since it is in use), but it does successfully delete everything IN the content folder.

### Steps to Reproduce

1. Go to the CMS File Manager page
2. Delete the root Content folder
3. Weep for all those poor bits.

**Expected behavior:** 

Button should not be clickable when inactive.

**Actual behavior:** 

Button is clickable when inactive.

### Versions

* **Rock Version:** 6.8,7.1
* **Client Culture Setting:** en-US

Issue #2659, opened 2018-01-11
Paging buttons at bottom of Communication History grid are unnresponsive - v7beta


JimMichael:
### Prerequisites
* [x] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)? NA
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

The buttons at the bottom of the Communications History grid that let you page through the results are unresponsive. Checked a few other grids (eg Transactions List) and the paging buttons work fine, so it's something specific to this page.

### Steps to Reproduce

1. On a Rock instance with lots of communication history, open the Communications History page.
2. Try to go to any other page via the buttons at the bottom of the grid.

**Expected behavior:** 

Buttons work

**Actual behavior:** 

They don't. I could not test this on prealpha as there are no communications in the grid.

### Versions

* **Rock Version:** v7beta 1.7.0.51

PR #2819, opened 2018-02-23
- Updated File Browser to really disable the Rename, Move and Delete …

cabal95:
## Contributor Agreement

## Context

The File Browser only visually disabled the buttons but they still actually tried to do what they needed to. In the case of Rename and Move that was to cause an exception. In the case of delete, it would allow you to delete the entire Content folder and all child folders.

~See #2659~, See #2695.

## Goal

Truly and fully disable the buttons.

## Strategy

Added code on the C# side to ignore those buttons for the root element in case the JavaScript fails. Added JavaScript to disable the button clicks if they are marked as disabled.

## Tests

## Possible Implications

## Screenshots

## Documentation

## Migrations


cabal95:
Forgot to mention #2695 in the commit message, so it needs to be manually completed when merged.

PR #2831, opened 2018-02-26
Fixed Lava preventing items from appearing in EventItemOccurrenceList…

mikejed:
…Lava block

# Rock PR Policy

## Contributor Agreement

## Context
When adding EventItemOccurrenceListLava block to a page, it uses `{% include '~~/Assets/Lava/EventItemOccurrenceList.lava' %}` as the Lava Template. When configured to point at an event, it correctly shows the name of the event, but does not show any occurrences. The for loop uses `{% for occurrence in EventItemOccurrences %}`, which has a size of 0. Examining the debug output for EventItem shows that EventItemOccurrences is a property of EventItem, not something passed to the block by itself

## Goal
Allows the default content of the block to display content

## Strategy
Changed the for loop to iterate over `EventItem.EventItemOccurrences` instead of `EventItemOccurrences` directly, in each of (3) themes.

## Tests

## Possible Implications
If there is another block where an install is referencing this file and EventItemOccurrences is passed to the block directly rather than as a property of EventItem, this update would break that block

## Screenshots

## Documentation

## Migrations

### DIFF
--- a/RockWeb/Themes/Flat/Assets/Lava/EventItemOccurrenceList.lava
+++ b/RockWeb/Themes/Flat/Assets/Lava/EventItemOccurrenceList.lava
@@ -3,7 +3,7 @@
     <h3 class="panel-title">Register for {{ EventItem.Name }}</h3>
   </div>
   <div class="list-group">
-    {% for occurrence in EventItemOccurrences %}
+    {% for occurrence in EventItem.EventItemOccurrences %}
       <a class="list-group-item" href="{{ RegistrationPage }}?RegistrationInstanceId={{ occurrence.Linkages | Select: 'RegistrationInstanceId' }}">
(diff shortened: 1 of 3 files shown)


mikejed:
Fixes #2833

mikejed:
Since I submitted this, I've seen people talking about similar loops in other blocks, and it sounds like in them, EventItemOccurrences is available by itself. So perhaps the block needs to be revised to provide the same entities as other similar blocks, rather than changing the lava to something that would only work with this block.

nairdo:
@mikejed in light of your discussion with Mike P on issue https://github.com/SparkDevNetwork/Rock/issues/2833 and his subsequent fix, would you say this PR is no longer needed -- so I can close it -- or is this still a needed change?

nairdo:
I'm closing this pull-request for now, but feel free to re-submit a new one if you think it still needs to be implemented.

mikejed:
Nick, I will have to confirm next week that the PR is no longer necessary. I expect its closure and disregard to be entirely warranted though. Thanks

PR #2849, opened 2018-03-06
Fixed not able to save checkin areas due to validation

trydyingtolive:
## Contributor Agreement

## Context
While trying to save a new check-in area Rock (V8) threw the validation error:  "Lava template for group view is mandatory." This is because new check-in areas are group types which do not have a lava display template generated for them.

## Goal
Fix the validation issue.

## Strategy
I changed the GroupType validation so that if the entity doesn't have a lava template it will add on the lava template of it's inherited group type. If it's inherited group type doesn't have a template the validation will fail. 

I also did some work so that the CheckInArea control will add the inherited group type to the entity so that it can be properly validated.

## Tests
Tested creating new area and on existing areas. Tested creating and updating group types.

## Possible Implications
None

## Screenshots
N/A

## Documentation
N/A

## Migrations
N/A

### DIFF
--- a/Rock/Model/GroupType.cs
+++ b/Rock/Model/GroupType.cs
@@ -474,7 +474,7 @@ namespace Rock.Model
             set { _triggers = value; }
         }
         private ICollection<GroupMemberWorkflowTrigger> _triggers;
-        
+
         /// <summary>
         /// Gets or sets the group schedule exclusions.
         /// </summary>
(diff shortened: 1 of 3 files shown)


edmistj:
@trydyingtolive Thanks for bringing up the issue. We may want to solve this in a different manner though. I think we'll create a default template to use if none is provided.

PR #2853, opened 2018-03-07
Adding an option to allow for not displaying workflow entry in a panel.

cbump:
# Rock PR Policy

## Contributor Agreement

## Context
We often want to use the workflow entry block without displaying it in a panel.  It's pretty easy to hide it with CSS but that feels wrong.

## Goal
This adds a block attribute giving you the option to not show the panel.

## Strategy
The OnLoad method simply removes the css and hides the heading of the panel if the panel is set to not display.

## Tests
None

## Possible Implications
None that I know of

## Screenshots
With the panel:
image
Without the panel:
image

## Documentation
I don't think it needs documented but I'm not sure if documentation for the Workflow Entry block exists somewhere.

## Migrations
None.

### DIFF
--- a/RockWeb/Blocks/WorkFlow/WorkflowEntry.ascx
+++ b/RockWeb/Blocks/WorkFlow/WorkflowEntry.ascx
@@ -6,9 +6,9 @@
         <div class="row">
             <div id="divForm" runat="server" class="col-md-6">
 
-                <div class="panel panel-block">
+                <div class="panel panel-block" id="divPanel" runat="server">
 
-                    <div class="panel-heading">
+                    <div class="panel-heading" id="divPanelHeading" runat="server">
                         <h1 class="panel-title">
                             <asp:Literal ID="lIconHtml" runat="server" ><i class="fa fa-gear"></i></asp:Literal>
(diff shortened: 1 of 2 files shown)


edmistj:
I worry that this could turn into something that is wanted on more blocks. It would be nice to create a solution that would be reusable. Two options come to mind:

1. Create a new Interface to tag these types of blocks IPaneledBlock and a RockPanelControl. This could then auto-wire up a property on the block settings to allow for hiding the panel (as you have done).

2. This is a lot simpler and achieves most of the same goals using CSS. We create a CSS class of .hide-panel. Blocks already have a setting to add a class to them. So adding the 'hide-panel' class would in effect hide the display. The CSS would be:

```
.hide-panel .panel {
    border: 0;
    background-color: transparent;
    margin: 0;
    box-shadow: none;
}

.hide-panel .panel-heading {
    display: none;
}

.hide-panel .panel-body {
    padding: 0;
    background-color: transparent;
    border: 0;
}
```

edmistj:
@cbump just to update this request. We think it might be good to allow this on any block in a panel. We're going to make a new Rock control for supporting this. These blocks would then support an interface to allow hiding the panel through the Block Settings dialog.

In the meantime you might use the CSS above as a temporary solution.

Issue #2854, opened 2018-03-08
[EN] Display Parent Rosters in Group Toolbox for Children's Groups


dcs619:
### Description
I have code that I can PR and tweak for core that adds support for Children's Groups.  Group leaders can access parental information, not just information about the child they lead.

The "Parent Roster" tab and "Email Parents" button will display when the following criteria are met:
1) Enable Parent Support is turned on (turned off by default)
2) Parent Roster page block setting is set 
3) One or more of the group members are designated as a Child in their Family group
4) One or more of those Children have a designated Adult in their Family group.  If the Child's Family does not have an Adult, that Child's name will not display on the Parent Roster tab.

Attached are screenshots of the proposed functionality.  This is sponsored by [Fellowship Greenville ](https://www.fellowshipgreenville.org/).

New "Parents Roster" tab and "Email Parents" button on the Group Toolbox page:

![screenshot 2018-03-08 09 48 38](https://user-images.githubusercontent.com/1210933/37157979-38962394-22b8-11e8-9409-8575490c2bad.png)

New "Parents Roster" tab displaying a basic list of parents and their contact information (nothing can be edited):

![screenshot 2018-03-08 09 49 05](https://user-images.githubusercontent.com/1210933/37157978-38888144-22b8-11e8-9ae5-b6cdbd305324.png)

New GroupDetailLava.ascx.cs block settings:

![screenshot 2018-03-08 09 55 40](https://user-images.githubusercontent.com/1210933/37157977-3879c41a-22b8-11e8-898d-51b413e91360.png)

The Parent Roster page is a copy of External Site\Support Pages\My Account\Group Toolbox as a new page:

![screenshot 2018-03-13 12 56 08](https://user-images.githubusercontent.com/1210933/37357158-0d00f7cc-26be-11e8-9b3b-3b2e0895b051.png)

This has to be a separate page in order for the Lava to correctly calculate the content display.

### Affected Files
* Blocks\Groups\GroupDetailLava.ascx.cs
* Stark\Assets\Lava\GroupDetail.lava

### Versions
* **Rock Version:** v7.3
* **Client Culture Setting:** en-us

edmistj:
Looks good based on showing if "Enabled Parent Support" = true and any group member is a child. 

v8 has an age classification on the Person model that tells you if they are a child. It would be very efficient to use that.

dcs619:
[...]

PR #2855, opened 2018-03-08
Implement Parent Rosters in Group Toolbox for Children's Small Groups #2584

dcs619:
## Contributor Agreement
✅

## Context

This PR adds Parent Roster support for Children's Small Groups. This allows leaders to have access to parental information, not just information about the child they lead.  See discussion, explanation and screenshots at #2584.

## Possible Implications

If this is merged to v8, @edmistj suggested using the Person.AgeClassification field to target Groups instead of depending on the GroupMember.GroupRole of Child in family groups.

## Migrations

The block setting for Enable Parent Support is off by default so this does not include a migration to add a new "Parent Roster Tab" page beneath the current Group Toolbox page.  If that's something I need to do I can include it.


edmistj:
@dcs619 How did you decide to determine if someone is a child?

dcs619:
@edmistj When [GetGroupMemberPeople](https://github.com/SparkDevNetwork/Rock/pull/2855/files#diff-c5a03cc4dc3377f85c43183b3b8a2bd8R1138) is called with transformParent = true, it checks if a Person in the Group has a Family.ChildRole, and returns any People with the Family.AdultRole.  This is the same code used by Person.GetFamilies(), but called once for all group members instead of per person.

Short version: if a Person doesn't have a Parent, they can't be a Child.  I think Rock allows a Family with a single Child (no Adult), but in that case they don't need a row on the Parent Roster.

azturner:
@dcs619 Were you going to update this to use the AgeClassification (see discussion on #2854)

dcs619:
@azturner Yes, sorry I missed the notification on that discussion

Issue #896, opened 2015-03-11
Creating a login via api/UserLogins results in 500 with matching info


jhawkzz:
If a new login is created using api/UserLogins, using a POST with model UserLoginWithPlainTextPassword, Rock will throw a 500 if there's already a login created with the same username and password.

Creating a login this way with a unique username / password works fine.

edmistj:
@jhawkzz as far as you know is this still an issue?

cabal95:
Confirmed. It generates an error 500 complaining about duplicate key IX_UserName.

Either this would be considered correct behavior, since I don't think any of the API methods really check for SQL errors about primary key violations on INSERT, OR we should modify the API to do so (can it be done globally somewhere?) and return a `409 Conflict` status.

edmistj:
@cabal95 think we just need to check if the username exists before inserting and reply with a 409 with a nice message.

PR #2872, opened 2018-03-14
Fixed the UserLogin API to return a 409 status if the username already exists.

cabal95:
## Contributor Agreement

## Context

See #896. When creating a user login through the API, if the username already exists an error 500 is returned.

## Goal

Make the 500 go away.

## Strategy

Now check to see if username exists and if so return a 409 Conflict status instead.

## Tests

## Possible Implications

## Screenshots

## Documentation

## Migrations


MikeDPeterson:
@cabal95  Thanks for cleaning that up image

Issue #2880, opened 2018-03-16
Registration Entry block does not honor Fee Order


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

When you setup optional fees for an even registration, you can drag them around to be in whatever order you want. But when the fees are rendered on the Registration page, the Order is ignored.

### Steps to Reproduce

1. Create a couple fees.
2. Change the order around.
3. Notice they are out of order.

**Expected behavior:** 

Order should be honored.

**Actual behavior:** 

Order is ignored completely and utterly.

### Screenshots

![image 1](https://user-images.githubusercontent.com/1119863/37536249-4352abe0-2907-11e8-8522-98b5880e9985.png)
![image 2](https://user-images.githubusercontent.com/1119863/37536257-461d8d18-2907-11e8-8d55-4089cee713a3.png)

### Versions

* **Rock Version:** 6.8, 7.x, 8.x
* **Client Culture Setting:** en-US

cabal95:
This line seems to be the culprit, FWIW.

https://github.com/SparkDevNetwork/Rock/blob/develop/RockWeb/Blocks/Event/RegistrationEntry.ascx.cs#L3849

PR #2893, opened 2018-03-22
Added missing order on fees

jreciomoreno:
Fees should retain the order they were assigned when showing on the form.

## Context
Fees are not being ordered on the form.

## Goal
Orders the fees.

## Possible Implications
N.A.


cabal95:
This will fix #2880.

PR #2923, opened 2018-04-03
Added support for automated registration discounts

mkinyon:
## Context
There are times when an organization has an event that offers a discount.  Instead of providing a discount code, they want to have the discount apply automatically based on the configured limits.

*Scenario 1: Group Discounts*
An organization offers an event with a cost of $25 per person.  If two or more people are registered together, the cost is automatically reduced to $15 per person.

 *Scenario 2: Limited Time Discount*
An organization offers an event that has a limited time discount.  If the person registers before the deadline, they automatically receive a 20% discount.

## Strategy
This will add a new registration discount setting that will allow the discount to be applied automatically based on the configured limits:
image

When the registrar gets to the summary screen and they qualify for a discount, the discount will be applied automatically.  A message will also be displayed with the discount that was applied:
image

*Additional Details:*

- Since only one discount code can ever be applied at a time, the first automatic discount that matches will be applied.  The first one is determined by the order of the discount codes in the registration template settings.

- The registrar can still apply a discount manually.  In that case, the manually entered discount will override the automated one.  

## Possible Implications
There is no potential impact for existing users since this feature is disabled by default.  The registration block will function exactly the same unless an admin enables "Auto Apply Discount" on a discount.

## Documentation
This section of the documentation will need to be updated to reflect the changes.  Also, it looks like this section is missing the newly added limits.
https://www.rockrms.com/Rock/BookContent/29/118#discounts

## Migrations
This pull request will need a migration to add a new column to the RegistrationTemplateDiscount table.  Let me know if you want me to provide one.

### DIFF
--- a/Rock/Model/RegistrationTemplateDiscount.cs
+++ b/Rock/Model/RegistrationTemplateDiscount.cs
@@ -132,6 +132,15 @@ namespace Rock.Model
         [DataMember]
         public DateTime? EndDate { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the discount applies automatically.
+        /// </summary>
+        /// <value>
+        /// <c>true</c> if this discount applies automatically; otherwise, <c>false</c>.
+        /// </value>
+        [DataMember]
+        public bool AutoApplyDiscount { get; set; }
+
         #endregion
 
         #region Virtual Properties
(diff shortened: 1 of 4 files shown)


azturner:
@mkinyon In looking at this more, realized that it's displaying an error message when a automatic discount is applied. Looks like it has to do with the ValidateSummary seeing that a discount was applied, but textbox is blank.

image

PR #2943, opened 2018-04-18
- Added LoadAttributes() function for collections of entities

chrismcgrath:
## Contributor Agreement

## Context

I often want to select a set of entities based on their attribute values, and in order to do that I have to loop through each entity in the collection and load the AttributesValues for each. Or sometimes I just want to be able to easily make sure that all the objects in the query I just did have their attributes loaded.

## Goal

To simplify the code required to load the attributes in a collection of entities.

## Strategy

I created LoadAttributes() functions that accept an IEnumerable<Attribute.IHasAttributes> and calls the individual LoadAttributes() function for each item.

## Tests

## Possible Implications

My only thought is that this may not be the most efficient way to load the attributes of many instances of the same EntityType all at once, but since the existing LoadAttributes() function is sufficiently complex and I have not noticed any performance issues, I didn't feel it was worth trying to improve it at this time.

## Screenshots

## Documentation

## Migrations


chrismcgrath:
Woops, merge conflict from some unrelated code that only had some spacing fixed, should be fine now.

PR #2955, opened 2018-04-27
Fix incorrect iPad printing of columns on parent label

lpeterson1:
# Rock PR Policy

## Contributor Agreement

## Context

The stock parent label doesn't print correctly from the iPad client. Line feeds don't work, so the columns of names/codes or ages/codes don't work and stray ampersands are printed.

## Goal

The label prints correctly.

## Strategy

`\&` is a line break in ZPL. When printing from the iPad client the `\` that creates the columns is getting stripped out before the label is sent to the printer. Embedding hex `5C 26` in the merge data sends the line break intact and the label prints correctly. To fix it, in the label ZPL I added the `^FH_` command to each merge data line, indicating that the character is hex if preceded by an underscore. I also changed the merge field Lava so the `\&` is now `_5C_26`.

If for any reason it's not desireable to use the underscore as the hex indicator, it's also possible to use `+`, `=`, `~`, and possibly others. Many characters I tried do not work.

I also cleaned up the rest of the label, removing the `^JUS` command and a few unnecessary `^FH\` commands. I widened the columns to allow for longer names and codes without line breaks. This could all be left as is if desired (it doesn't affect the actual fix).

## Tests

I tested with both the iPad and Windows client and it works correctly on both.

## Possible Implications

The new label code will still work the way it always has when used with the old merge field Lava (works on Windows, broken on iPad). The old label code will not work correctly with the new merge fields.

## Screenshots

NA

## Documentation

I don't see that this bug was documented in the Checking Out Check-In guide, so I don't believe there are any necessary changes.

## Migrations

I think this will mean a migration, but I'm not clear how to handle that in this PR. I did change a file from the 1.5.0 migration. Let me know if I blew this part. :-)


lpeterson1:
@ethan-sparkdevnetwork I'm probably missing something you already did, but the icon label will need to be changed too. The new merge codes won't work correctly with the old label.

ethan-sparkdevnetwork:
The migration includes a change to the binary file stored in BinaryFileData. This has been added to V7.4 but not to 8.0.

ethan-sparkdevnetwork:
Removed changes to the label due to potential impact of other labels that use those merge codes.
This issue will get fixed in the iPad app instead of using a work-around.

Issue #2887, opened 2018-03-20
Rock Check Scanner finds wrong MICR check # on some checks


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

The Check number on checks MICR code is not read correctly from some checks by the Rock Check Scanner WPF application.

<img alt="image" src="https://user-images.githubusercontent.com/1119863/37682533-1b02d082-2c47-11e8-8789-91b82e996813.png">

Read MICR number: `d#########d ##########c1309 10                   `
MICR Number Parts found: `#########_##########_10`

As you can see it does read the MICR number correctly, but when trying to decipher the parts it mistakenly takes the `10` instead of the `1309`.

Suggested fix would be to change how the "extra information" for check number (since it is not part of the MICR spec) is handled. When multiple space separated numeric values are found, take the longer one as the check number.

I can work on a PR for this as it is an ongoing issue. We have a number of checks each week that have this problem.

**Expected behavior:** 

Check number should be read accurately from the check.

**Actual behavior:** 

It isn't.

### Versions

* **Rock Version:** 6.8
* **Client Culture Setting:** en-US

PR #2961, opened 2018-05-01
Fixed Check Scanner application to better handle check numbers

cabal95:
## Contributor Agreement

## Context

See #2887. Check scanner will sometimes pick a short 2 digit field in the MICR data instead of the longer (more probable) field for the check number.

## Goal

Pick the longest field in the MICR extra data.

## Strategy

Order by length of parts and take the last element.

## Tests

## Possible Implications

## Screenshots

## Documentation

## Migrations

### DIFF
--- a/Applications/Wpf/CheckScannerUtility/CheckScannerUtility/ScanningPage.xaml.cs
+++ b/Applications/Wpf/CheckScannerUtility/CheckScannerUtility/ScanningPage.xaml.cs
@@ -536,7 +536,7 @@ namespace Rock.Apps.CheckScannerUtility
                     string otherData = null;
                     if ( remainingMicrParts.Any() )
                     {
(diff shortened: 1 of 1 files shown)


cabal95:
@nairdo It's a "best guess" to get the check number since there is no standard that banks are required to follow (some banks put it as part of the Account number section, which we can't do anything about). We get a lot of checks that use check numbers like `1309 10`, `1829 10` and such. `10` is just some constant number they put in that has nothing to do with the check number (no idea what it means since it's bank specific). The previous code would always take the `10` as the check number.

It's not a perfect solution, just "slightly better". :)

MikeDPeterson:
@cabal95 looks good.  One suggestion is to add a comment above you changed line of code that explains the reasoning.  I'll go ahead and add the comment for you after I merge this.

PR #2993, opened 2018-05-15
Added support to optionally use the first registrant's information to…

mkinyon:
… auto complete and hide the registrar information.

## Context
The goal of this pull request is to streamline the registration process for events where you know that the person registering is also the registrar.  For example, if someone is registering themselves for an event and does not have a login they would need to enter their information twice.  The first time to collect their information and then a second time to fill in the registrar information.

This streamlines the process by giving the event coordinator the option to auto fill the registrar's information with the first registrant's information and also hide the registrar form altogether.

## Strategy
This will add a new registration template setting that will let the user change how registrar information is collected:
image

The different options are:

**Prompt For Registrar**
This is the default setting that will display the registrar information panel as it currently does today.

**Prefill First Registrant**
This setting will auto fill the registrar panel with the first registrant's information.  This is helpful in cases where the registrants do not login to register but you still want to auto fill their information.

**Use First Registrant**
This setting will also auto fill the registrar panel with the first registrant's information but will also hide the registrar panel altogether.  The only exception would be if the registration form does not collect the registrants' email address.  In that case, the panel will be displayed since that information is required for communications.

## Possible Implications
There is no potential impact for existing users since the default setting `Prompt For Registrar` uses the existing behavior.

## Documentation
This section of the documentation will need to be updated to reflect the changes:
https://www.rockrms.com/Rock/BookContent/29/118#generalsettings

## Migrations
This pull request will need a migration to add a new column called `RegistrarOption` to the RegistrationTemplate table. Let me know if you want me to provide one.


nairdo:
@mkinyon This is a great write-up, thanks.  I could not get this PR to work in that build of the develop branch, but it might be due to a bug with the new caching stuff.  (It does mostly work if you re-submit the PR against the latest develop branch.) 

Here are a few things based on my testing.

1) Can you change the Registration Template UI option to use the term "Pre-fill ..." (hyphenated) 
2) When using the Pre-fill First Registrant option, I think the "...is in the same family as" option should also auto-select the correct person's name in the radio button list. (I'd say select the name and then hide the radio button list altogether, but the person could change the _Completed By_ information -- and then you'd need to clear and show the radiobutton selection).
image
3) When using the Use First Registrant, the same situation as the previous item. 
4) I'm getting a "Non-static method requires a target." error after pressing the final button, but it may be an unrelated problem in the develop branch.
image

mkinyon:
Thanks for the information, Nick!  I’m on vacation this week so I take care of this the week of June 11th.

mkinyon:
@nairdo, I have taken care of the four issues that you mentioned above.  I also updated my feature branch with the latest code from develop.  Let me know if that creates a problem.

nairdo:
@mkinyon This looks good, tests well, and builds perfectly (once I added the migration for the new column).  I'll merge this later tonight when I have more time so I can also commit the separate migration through the develop branch immediately after the merge.

mkinyon:
@nairdo Thank you very much for reviewing this again!

Issue #2998, opened 2018-05-18
[EN] Smart Field Support for SignNow


treyhendon:
### Description

I wrote enhancement for SignNow to support Smart Fields and have code for PR ready.  Sponsoring church was [Lake Pointe](https://www.lakepointe.org/).

Added two settings (Attribute Key setting for the Merge Key Pair with the Smart Fields and Enabled Lava Commands) to the SignNow provider.

Here is the current code before I submit the PR: https://github.com/treyhendon/Rock/commit/18b50a579941f847d995f10e606e5462d2f2ee46#diff-4abf30929be9ad6a25713d2173af6ae1 

### Affected Files

* Rock.SignNow\SignNow.cs

### Versions

* **Rock Version:** 1.8.0.16  
* **Client Culture Setting:** en-us

edmistj:
We might hold pulling this in as we have some new strategies coming for digital signatures. Until we get those fully ironed out we'd prefer not make changes to the existing providers. Going forward, it would be great to have conversations about core submissions before coding so we can collaborate on direction.

treyhendon:
I get that.  Sorry this has been sitting in my code bank since last fall and I haven't been able to do any PRs.  Should we close this thread?

treyhendon:
@edmistj I was asked about Sign Now Smart Fields today by a church and wanted to see if there was any opportunity to reopen this conversation?

edmistj:
@treyhendon we're looking to reduce the core ties to SignNow (possibly eliminate it).

treyhendon:
Wow, would that remove the signature framework from Rock for the new core features? Or do you mean that Sign Now would be moved to a shop item and not install initially?  Asking because I'm wondering if it's worth sharing what we have with more churches or will it be ultimately useless? Thanks!

edmistj:
The details are still up in the air pending what we're able to achieve with the feature.

PR #2999, opened 2018-05-18
Added Smart Fields to SignNow

treyhendon:
## Context
Needed SignNow to have ability to populate form values.  Discussed in #2998

## Goal
Add the ability to populate fields in a signature document.

## Strategy
Using a key pair attribute on the document, users can populate SignNow Smart Field values using Lava commands.

## Possible Implications
No possible implications.

## Documentation
Sign Now Properties Photo 
6 - The Attribute Key of the Merge Document Template attribute that will contain Smart Fields and Values.
7 - The allowed Lava Commands the Smart Field Values can use.

### DIFF
--- a/Rock.SignNow/SignNow.cs
+++ b/Rock.SignNow/SignNow.cs
@@ -20,15 +20,17 @@ using System.ComponentModel;
 using System.ComponentModel.Composition;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


edmistj:
Closing for now as this does not fit with our go forward digital signing strategy.

Issue #2910, opened 2018-03-27
New Workflow Action: Process Incoming SMS


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

This was discussed briefly on Slack but wanted to get it here for better tracking. I have a custom Workflow Action that "injects" a text string into the two-way SMS system. You provide the From Number, To Number and Message body and it sends it off to the SMS handler to be treated like a normal incoming SMS, as if it just came in from Twilio to the standard handler. This allows you to have a phone number handled by the Text To Workflow processing, and then still pass the message on to the two-way system.

![image](https://user-images.githubusercontent.com/1119863/37998761-cc198cc0-31d4-11e8-81d2-12a3236cc5b0.png)

All 3 fields support Lava. I would expect this to land in the 8.0 release, but until then it is available as a plugin.

@edmistj had recommended the action be renamed to `Process SMS as Communication Response`. I'm going to tweak that a bit to be `Process as SMS Communication Response` as I think that is clearer to what it is actually doing. It's verbose, but gets the point across.

### Versions

* **Rock Version:** 8.0?
* **Client Culture Setting:** en-US

edmistj:
Sounds good. I'm good with the name tweak.

PR #3002, opened 2018-05-18
New workflow action for processing messages as SMS Responses

cabal95:
## Contributor Agreement

## Context

When processing an SMS message from a workflow, there is currently no way to them decide to pass it off to the Text2Person system. See #2910.

## Goal

Provide a Workflow action to allow this functionality.

## Strategy

New workflow action that accepts the following parameters:

* From Number (or Attribute)
* To Number (or Attribute)
* Message (accepts Lava)
* Error Attribute (optional)

The only one that probably needs explanation is the Error Attribute. If the SMS system returns an error then the error message will be placed in this attribute. Currently, the only error that would be returned is the standard `Could not deliver message. This phone number is not registered in the Rock Solid Church database.` error. This allows the admin to then check if this value is non-blank, and if so send it as a response to the user.

## Tests

## Possible Implications

## Screenshots

screenshot-2018-5-18 workflow configuration rock rms 1

## Documentation

## Migrations

### DIFF
--- a/Rock/Rock.csproj
+++ b/Rock/Rock.csproj
@@ -2129,6 +2129,7 @@
     <Compile Include="Workflow\Action\CheckIn\RemoveEmptyPeople.cs" />
     <Compile Include="Workflow\Action\CheckIn\SaveAttendance.cs" />
     <Compile Include="Workflow\Action\CheckIn\SetAvailableSchedules.cs" />
+    <Compile Include="Workflow\Action\Communications\ProcessAsSmsCommunicationResponse.cs" />
     <Compile Include="Workflow\Action\Communications\SendEmail.cs" />
     <Compile Include="Workflow\Action\Communications\SendEmailWithEvents.cs" />
     <Compile Include="Workflow\Action\Communications\SendNotification.cs" />
--- a/Rock/Workflow/Action/Communications/ProcessAsSmsCommunicationResponse.cs
+++ b/Rock/Workflow/Action/Communications/ProcessAsSmsCommunicationResponse.cs
@@ -1,0 +1,80 @@
+﻿// <copyright>
+// Copyright by the Spark Development Network
+//
+// Licensed under the Rock Community License (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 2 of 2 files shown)


edmistj:
For consistency can we name this 'SMS Communication Send'. Also, to confirm, does the From Number & To Number attributes take a person or phone number? This is how the Send Email action works. We'd want this to work the same way.

cabal95:
The two From Number and To Number are text attributes because when processing an INCOMING text message, the system needs specific phone numbers to work with.

Based on your comment, I think maybe you were thinking this was to send a message out, it is not. This lets the admin simulate an incoming message into the Person-to-Person SMS system.  Example use case:

I setup a workflow to handle incoming keywords, it does the work of figuring out which keyword they are sending. If no keyword is found, we assume it's meant to to be a person-to-person text message so we want to inject it into the person-to-person system. To do that, we need to know the from number, the to number, and the message the original SMS contained.

Based on that, I don't think "SMS Communication Send" makes sense, since we are not sending a communication out.

edmistj:
Ah OK. We need to get better context provided on what PRs are trying to accomplish. 

Based on this I would recommend the name "SMS Communication Process Response". We've standardized on the pattern "Noun - Verb". In some cases this reads a bit awkward, but it keeps the like actions together (grouped by noun). We've updated the 'ComponentNames' of previous actions to reflect that, but were not able to change the file/class names as there were already in use.

If you make those changes we can pull this in.

cabal95:
Action name has been renamed.

For my reference, was the original discussion context in the referenced issue not clear, or is it that I need to copy/paste the original context discussion from the referenced issues into the PRs as well?

nairdo:
@cabal95 Can you verify that this write-up is accurate and clear? https://www.rockrms.com/WorkflowActionCategory?Category=5#smscommunicationprocessresponse (it took me several attempts to really understand what this action is all about.  I finally decided that the example case was really needed to grasp the intent.)

cabal95:
@nairdo Yep, looks good to me. It is a bit of an odd one, so I agree with you about the example use case being the best way to convey how to use it.

Issue #2989, opened 2018-05-14
Site Crawler incorrectly skips links when robots contains noindex


cabal95:
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

The Rock Site Crawler that came with Universal Search will not follow links if the robots `noindex` option is specified. This is incorrect, links should be followed unless the robots meta specifies the `nofollow` option, or the link itself has a `rel="nofollow"` option.

I can submit a PR for this.

### Suggested Action

Add support for the `nofollow` flag in the robots meta tag. At the same time update the `ParseLinks` method to check for `rel="nofollow"` in the link and if found skip that individual link.

**Expected behavior:** 

I want to build a page that has links to other pages that should be indexed but would not normally be found during a site crawl (example, event pages whose links only show up after clicking a PostBack button, which cannot be indexed).

Additionally, there are a few pages on the site that we don't want indexed because they are little more than menu/link-only pages.

**Actual behavior:** 

These link-only pages are indexed because I cannot have the crawler follow links but not index the page itself.

### Versions

* **Rock Version:** 7.3
* **Client Culture Setting:** en-US

edmistj:
Can you provide a bit more info on your specific use case. I'm not sure I understand this line:
>The Rock Site Crawler that came with Universal Search will not follow links if the robots noindex option is specified.

Where is this specified?

cabal95:
> Can you provide a bit more info on your specific use case.

We have about 300 calendar events. But only a handful will be indexed because the Events page, the one with the mini calendar and short descriptions of the events, will only show the events for the current month. Unless the user clicks the little arrow to go to the next month. This is done via PostBack which means the crawler cannot get to this information, since it is a Javascript link. In fact only about 25 of these events showed up on the index after crawling the site.

[...]

PR #3015, opened 2018-05-22
+ Updated Site Crawler to honor the <meta name="rebot" value="nofollo…

cabal95:
## Contributor Agreement

## Context

See #2989 for full discussion.

We have pages that we do not want indexed for searching, but we do want links to be followed. An example of this would be menu pages. They contain links to other pages, but really have no content of their own - other than maybe a short statement of "Select your campus below". As such, these pages should not come up in search results.

With normal search engines / site crawlers, you can specify `<meta name="robot" value="noindex">` to indicate that the _current page_ should not be indexed. If you want to specify that the _current page_ should not be indexed **and** no links should be followed, you would specify `<meta name="robot" value="noindex,nofollow">`.

With the Site Crawler built into Rock, it currently does not honor the `nofollow` option. Instead it treats `noindex` to mean `noindex,nofollow`.

## Goal

The Site Crawler has been updated to treat `noindex` the way other index engines work which is to not index that page, but still follow links.

Additionally, before following links it now checks if `<meta name="robot" value="...">` contains the value `nofollow`. If so then no links are followed on the page. If that value is not found, then each individual link is checked to see if it has `rel="nofollow"` which indicates that one individual link should not be followed.

## Tests

## Possible Implications

## Screenshots

## Documentation

## Migrations

### DIFF
--- a/Rock/UniversalSearch/Crawler/Crawler.cs
+++ b/Rock/UniversalSearch/Crawler/Crawler.cs
@@ -181,7 +181,10 @@ namespace Rock.UniversalSearch.Crawler
                                 }
 
                                 IndexContainer.IndexDocument( sitePage );
+                            }
 
+                            if ( metaRobot == null || metaRobot.Attributes["content"] == null || !metaRobot.Attributes["content"].Value.Contains( "nofollow" ) )
+                            {
                                 // crawl all the links found on the page.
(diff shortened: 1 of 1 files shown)


cabal95:
Yeaaaa... noticed that misspelling the moment _after_ I clicked `push`. :)

PR #3021, opened 2018-05-25
Fixed missing icons in Verify Security block due to FontAwesome 5 changes

cabal95:
## Contributor Agreement

## Context

When Rock changed to FontAwesome 5, many icons in my personal plugins stopped working because the `fa fa-` classes were on the `a` tags instead of embedded in `i` tags inside the `a` tags.

Examples:

Works: `<a class="btn btn-sm"><i class="fa fa-unlock"></i></a>`

Doesn't work: `<a class="btn btn-sm fa fa-unlock"></a>`

Most things use the working style, but a few use the latter style that doesn't always work. I've been updating my plugins as I find these, but this one is in core.

If we do a 1.7.5 release, this should be cherry picked back.

## Goal

Fix the block to make the icon show up.

## Strategy

Change the format of the btn to use the `i` tag for the FontAwesome class.

## Tests

## Possible Implications

## Screenshots

## Documentation

## Migrations


cabal95:
Scratch this, new PR incoming with layout fixes too (thought they would conflict with v7 branch, but they actually are fine).

PR #3022, opened 2018-05-25
- Fix to break the cache of the FontAwesome files when a new version is installed

chrismcgrath:
## Context

I installed the latest version of FontAwesome to make use of some of the new icons, but they did not display properly in the browser until I cleared my cache.

## Goal

This will cause the browser to get the updated Font when new version files are installed so that individual users don't have to clear their caches to get the new icons.

## Strategy

I added the FontAwesome version number variable to the v= parameter on all the font URLs.  I left the edition in this parameter as well, so it will look something like ?v=free5.0.13


edmistj:
@garrettjohnson can you review and merge when you're done.

mkinyon:
FYI, I tested the font awesome fix on an affected system and it works great.  Thanks @chrismcgrath!

PR #3023, opened 2018-05-25
Fixed icon (v7) and layout issues (v8) with the Verify Security block.

cabal95:
## Contributor Agreement

## Context

> Note: There has been no issue created as this originally started as a one line fix for the icon and then I noticed the layout issue on v8 (since that is what my develop branch is on).

### Icon Missing

When Rock changed to FontAwesome 5, many icons in my personal plugins stopped working because the `fa fa-` classes were on the `a` tags instead of embedded in `i` tags inside the `a` tags.

Examples:

Works: `<a class="btn btn-sm"><i class="fa fa-unlock"></i></a>`

Doesn't work: `<a class="btn btn-sm fa fa-unlock"></a>`

Most things use the working style, but a few use the latter style that doesn't always work. I've been updating my plugins as I find these, but this one is in core.

See "Before" screenshot for example of missing icon in lower left.

### Layout Issues

The way I designed the layout of this block causes issues with the new theme changes in v8. I used some CSS classes in ways that I apparently should not have. See margin issues in the "Before" screenshot, arrow pointing on the left edge.

## Goal

Fix the icon and the layout issues.

## Strategy

To fix the icon, instead of the `fa fa-unlock` class going on the anchor tag, it now uses a standard `<i class="fa fa-unlock"></i>` inside the anchor tag.

To fix the layout issues, split the block layout from one panel into two panels, one for the results grid.

Tested the changes on both v8 and v7 to ensure this PR can be cherry picked back to a 1.7.5 release if we do one.

## Tests

## Possible Implications

## Screenshots

Before:

before

After (v7):

afterv7

After (v8):

afterv8

## Documentation

## Migrations

### DIFF
--- a/RockWeb/Blocks/Security/VerifySecurity.ascx
+++ b/RockWeb/Blocks/Security/VerifySecurity.ascx
@@ -2,48 +2,53 @@
 
 <asp:UpdatePanel ID="upnlContent" runat="server">
     <ContentTemplate>
-        <asp:Panel ID="pnlSecurity" runat="server" CssClass="panel panel-block">
+        <div class="panel panel-block">
             <div class="panel-heading">
                 <h1 class="panel-title"><i class="fa fa-lock"></i> Verify Security</h1>
             </div>
 
             <div class="panel-body">
-                <div class="grid grid-panel">
+                <Rock:NotificationBox ID="nbWarning" runat="server" NotificationBoxType="Warning" Dismissable="true" />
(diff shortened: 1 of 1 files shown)


cabal95:
Be sure to check it on the new v8 theme. I did the second commit because when I looked at it on v8 it was completely jacked up (my fault for using CSS classes that I shouldn't have, but looked correct on v7).

PR #3050, opened 2018-06-07
Fix up file handlers (Fixes #2599)

### LINKED ISSUE #2599: GetFile.ashx throws exceptions for missing files/invalid parameters
### Prerequisites
* [X] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

We've noticed a lot of exceptions in our logs from GetFile.ashx, both for files which have been deleted and requests with invalid parameters (Usually sql injection probes). These exceptions should be handled properly and return appropriate responses instead of throwing server errors. I've also noticed that the response for these errors is served with a 200 status code, which could cause other issues where a client interprets it as a successful response. (For example, trying to use the error html as image data when requesting an image file)

### Steps to Reproduce

A. Attempt to request a file that either does not exist or has been removed.
or
B. Attempt to request a file with a malformed parameter (the guid parameter isn't a guid, id parameter isn't an int, etc).

**Expected behavior:** 

Requests for missing or removed files should return a 404 Not Found response and shouldn't throw an exception.

Requests for files using malformed parameters should return a 400 Bad Request response and shouldn't throw an exception.

**Actual behavior:** 
Requests for a missing or removed file generate the following exception:
```
System.InvalidOperationException | System.Data | Invalid attempt to read when no data is present.
   at System.Data.SqlClient.SqlDataReader.CheckDataIsReady(Int32 columnIndex, Boolean allowPartiallyReadColumn, Boolean permitAsync, String methodName)
   at System.Data.SqlClient.SqlDataReader.TryReadColumn(Int32 i, Boolean setTimeout, Boolean allowPartiallyReadColumn)
   at System.Data.SqlClient.SqlDataReader.GetValueInternal(Int32 i)
   at System.Data.SqlClient.SqlDataReader.GetValue(Int32 i)
   at Rock.Model.BinaryFileService.EndGet(IAsyncResult asyncResult, HttpContext context, Boolean& requiresViewSecurity)
   at RockWeb.GetFile.EndProcessRequest(IAsyncResult result)
```

Requests for files using malformed parameters generate the following two exceptions:
```
System.Exception | App_Code.v-in4up7 | file id key must be a guid or an int
   at RockWeb.GetFile.BeginProcessBinaryFileRequest(HttpContext context, AsyncCallback cb)
   at RockWeb.GetFile.BeginProcessRequest(HttpContext context, AsyncCallback cb, Object extraData)
```
```
System.NullReferenceException | System.Web | Object reference not set to an instance of an object.
   at System.Web.HttpApplication.AsyncStepCompletionInfo.RegisterBeginUnwound(IAsyncResult asyncResult, Boolean& operationCompleted, Boolean& mustCallEndHandler)
   at System.Web.HttpApplication.CallHandlerExecutionStep.System.Web.HttpApplication.IExecutionStep.Execute()
   at System.Web.HttpApplication.ExecuteStep(IExecutionStep step, Boolean& completedSynchronously)
```

In addition, all of these requests seem to return a successful response (200 status code), but with an html error page as content.

### Versions

* **Rock Version:** checked on our production 1.6.5 install, a clean 1.6.9 install, and the pre-alpha server

Tschrock:
## Context
The original problem was that the get file handlers responded improperly to some requests and were generating a lot of exceptions in those cases (See #2599). I've also discovered some more issues with them in the process of fixing that issue.

## Changes:
1. `GetFile.ashx.cs`
  * Removed content file support (the `isBinaryFile` parameter).
    * The current implementation has been broken for a while.
    * I can't find it used anywhere.
    * It's unnecessary (These files can already be accessed from `/Content`).
  * Invalid parameter options now return a 400 Bad Request instead of throwing an exception.
  * Missing or deleted files now return a 404 Not Found instead of throwing an exception.
    * For this, `BinaryFileService.EndGet()` was modified to properly check if data could be read, and now returns null if a file is missing instead of throwing an exception (existing usages already expect this behavior).
  * Changed `IsReusable` to return `true`.
    * Since we don't store the current request state in the handler, we can tell IIS to reuse it instead of making a new one for each request. This should make it a tiny bit more efficient.
1. `GetImage.ashx.cs`
  * Content files (`?isBinaryFile=F`):
    * Fixed issues with path handling
    * Fixed parameter counting for determining if a content file should try to be resized.
    * Missing or deleted content files now return a 404 Not Found instead of an empty 200 Success.
    * Invalid content files now return a 400 Bad Request instead of an empty 200 Success.
    * Content files that don't have permission to be read now return a 403 Forbidden instead of an empty 200 Success.
  * Binary files (`?isBinaryFile=T`):
    * Invalid parameters now return a 400 Bad Request instead of an empty 200 Success.
    * Fixed an issue with file security
  * General:
    * Exceptions now return a 500 Internal Server Error instead of an empty 200 Success.
    * Files that fail to resize now fallback to their original content instead of throwing an ObjectDisposedException.
    * Changed `IsReusable` to return `true`.
      * Since we don't store the current request state in the handler, we can tell IIS to reuse it instead of making a new one for each request. This should make it a tiny bit more efficient.
1. `FileUploader.ashx.cs`
  * Content files (`?isBinaryFile=F`):
    * Fixed issues with path handling.
    * Changed handling to prevent overwriting existing files (If you actually need to replace a file, you should delete it in the file browser and re-upload it).
  * General
    * Fixed file type validation.
1. `ImageUploader.ashx.cs`
  * Fixed file type validation.

### DIFF
--- a/RockWeb/App_Code/FileUploader.ashx.cs
+++ b/RockWeb/App_Code/FileUploader.ashx.cs
@@ -144,6 +144,13 @@ namespace RockWeb
 
             string scrubedFileName = ScrubFileName( untrustedFileName );
 
+            if( string.IsNullOrWhiteSpace( scrubedFileName ) )
+            {
+                throw new Rock.Web.FileUploadException( "Invalid File Name", System.Net.HttpStatusCode.BadRequest );
+            }
+
+            /* Scrub the folder path */
+            string scrubedFolderPath = ScrubFileName( untrustedFolderPath );
 
             /* Determine the root upload folder */
 
@@ -168,7 +175,11 @@ namespace RockWeb
             string trustedPhysicalRootFolder = Path.GetFullPath( context.Request.MapPath( trustedRootFolder ) );
 
             // Treat rooted folder paths as relative
-            string untrustedRelativeFolderPath = untrustedFolderPath.TrimStart( Path.GetPathRoot( untrustedFolderPath ).ToCharArray() );
+            string untrustedRelativeFolderPath = "";
+            if ( !string.IsNullOrWhiteSpace( scrubedFolderPath ) )
+            {
+                untrustedRelativeFolderPath = scrubedFolderPath.TrimStart( Path.GetPathRoot( scrubedFolderPath ).ToCharArray() );
+            }
 
             // Get the absolute path for our untrusted folder.
             string untrustedPhysicalFolderPath = Path.GetFullPath( Path.Combine( trustedPhysicalRootFolder, untrustedRelativeFolderPath ) );
@@ -336,17 +347,32 @@ namespace RockWeb
         }
 
         /// <summary>
-        /// Scrubs a filename to make sure it doen't have any directories or bad characters
+        /// Scrubs a filename to make sure it doesn't have any directories or invalid characters
         /// </summary>
         /// <param name="untrustedFileName">The filename.</param>
         /// <returns>A scrubed filename.</returns>
         public string ScrubFileName(string untrustedFileName )
         {
+            // Scrub invalid path characters
+            untrustedFileName = ScrubFilePath( untrustedFileName );
+
             // Get the base filename
             string baseFileName = Path.GetFileName( untrustedFileName );
 
-            // Scrub invalid characters
+            // Scrub invalid file characters
             return Regex.Replace( baseFileName, "[" + Regex.Escape( Path.GetInvalidFileNameChars().ToString() ) + "]", string.Empty, RegexOptions.CultureInvariant );
         }
+
+        /// <summary>
+        /// Scrubs a file path to make sure it doesn't have any invalid characters
+        /// </summary>
+        /// <param name="untrustedFilePath">The file path.</param>
+        /// <returns>A scrubed file path.</returns>
+        public string ScrubFilePath( string untrustedFilePath )
+        {
+            // Scrub invalid path characters
+            return Regex.Replace( untrustedFilePath, "[" + Regex.Escape( Path.GetInvalidPathChars().ToString() ) + "]", string.Empty, RegexOptions.CultureInvariant );
+        }
+
     }
 }
--- a/RockWeb/App_Code/FileUploader.ashx.cs
+++ b/RockWeb/App_Code/FileUploader.ashx.cs
@@ -133,54 +133,82 @@ namespace RockWeb
             // validate file type (child FileUploader classes, like ImageUploader, can do additional validation);
             this.ValidateFileType( context, uploadedFile );
 
-            //
-            // Get filename and scrub invalid characters.
-            //
-            var filename = Path.GetFileName( uploadedFile.FileName );
-            filename = Regex.Replace( filename, @"[<>*%&:\\]", string.Empty, RegexOptions.CultureInvariant );
 
-            // get folderPath and construct filePath
-            string relativeFolderPath = context.Request.Form["folderPath"] ?? string.Empty;
-            string relativeFilePath = Path.Combine( relativeFolderPath, filename );
-            string rootFolderParam = context.Request.QueryString["rootFolder"];
+            // NEVER TRUST THE CLIENT!!!!
+            string untrustedFileName = uploadedFile.FileName;
+            string untrustedFolderPath = context.Request.Form["folderPath"] ?? string.Empty;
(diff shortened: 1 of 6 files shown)


RockWeb/App_Code/FileUploader.ashx.cs
```diff
@@ -133,54 +133,82 @@ private void ProcessContentFile( HttpContext context, HttpPostedFile uploadedFil
...
 
-            // Make sure the physicalContentFolderName doesn't have any special directory navigation indicators
-            if ( physicalContentFolderName != System.IO.Path.GetFullPath( physicalContentFolderName ) )
+            /* Combine the root and folder paths to get the real physical location */
+
+            // Get the absolute path for our trusted root.
+            string trustedPhysicalRootFolder = Path.GetFullPath( context.Request.MapPath( trustedRootFolder ) );
+
+            // Treat rooted folder paths as relative
+            string untrustedRelativeFolderPath = untrustedFolderPath.TrimStart( Path.GetPathRoot( untrustedFolderPath ).ToCharArray() );
```
MikeDPeterson: I get an exception here when untrustedFolderPath = "".  To reproduce this issue, use the FileUploader (ContentFileMode) in the Rock ControlGallery.

RockWeb/App_Code/GetFile.ashx.cs
```diff
@@ -39,84 +38,21 @@ public class GetFile : IHttpAsyncHandler
         /// <param name="extraData">Any state data needed to process the request.</param>
         /// <returns>An IAsyncResult that contains information about the status of the process.</returns>
         public IAsyncResult BeginProcessRequest( HttpContext context, AsyncCallback cb, object extraData )
-        {
-            try
-            {
-                // Check to see if this is a BinaryFileType/BinaryFile or just a plain content file (if isBinaryFile not specified, assume it is a BinaryFile)
-                bool isBinaryFile = ( context.Request.QueryString["isBinaryFile"] ?? "T" ).AsBoolean();
```
MikeDPeterson: I did find one spot in core that was using IsBinaryFile.  It's the CMS > FileManager, if a person were to click on the link on the FileUploader after uploading a file.  However, it is broken :).  Like you mentioned, it could just use a regular link.  FileUploader.cs would need to be fix though.  We could probably just fix FileUploader.cs LIne 648 to be `_aFileName.HRef = ResolveUrl( this.UploadedContentFilePath );`

Tschrock:
I'll take a look at these this afternoon.

Tschrock:
Sorry for the late update, we've been really busy lately. This should address the two issues that were brought up.

ethan-sparkdevnetwork:
Thanks Tyler. I merged this into 8.7 and will merge it to develop for the next pre-alpha release.

RockWeb/App_Code/FileUploader.ashx.cs
```diff
@@ -133,54 +133,93 @@ private void ProcessContentFile( HttpContext context, HttpPostedFile uploadedFil
...
+            if( string.IsNullOrWhiteSpace( scrubedFileName ) )
             {
-                // if a rootFolder was specified in the URL, decrypt it (it is encrypted to help prevent direct access to filesystem)
-                rootFolder = Rock.Security.Encryption.DecryptString( rootFolderParam );
+                throw new Rock.Web.FileUploadException( "Invalid File Name", System.Net.HttpStatusCode.BadRequest );
             }
 
-            if ( string.IsNullOrWhiteSpace( rootFolder ) )
+            /* Scrub the folder path */
+            string scrubedFolderPath = ScrubFileName( untrustedFolderPath );
```
nairdo: @Tschrock FYI -- In case you are using this in your system before you get our fix (https://github.com/SparkDevNetwork/Rock/commit/afc1aad6449d0de45c6f9f36e514165f132c9f21), this is probably supposed to be ScrubFile**Path** not _Name_.

PR #3071, opened 2018-06-15
+ Attendance Codes will now properly prevent duplicate and banned values when combining alpha and numeric characters

chrismcgrath:
## Context

When setting a check-in configuration to use 2 alpha and 2 numeric characters, I found that the code to generate characters applied its logic to prevent banned and duplicate codes to the alpha and numeric portions separately, but not against the combined code.

## Goal

All pieces of attendance codes will be generated at the same time and then compared against the lists of banned and existing codes, repeating the process until a valid one is generated.

## Strategy

Simplified the attendance code generation into a single statement inside a single while loop that checks against the disallowed codes.


Rock/Model/AttendanceCodeService.Partial.cs
```diff
@@ -158,40 +152,25 @@ public static AttendanceCode GetNew( int alphaNumericLength, int alphaLength, in
...
+            if ( !string.IsNullOrEmpty( lastCode ) )
+            {
+                var maxCode = lastCode.Substring( alphaNumericLength + alphaLength );
+                int nextCode = maxCode.AsInteger() + 1;
 
-                    numericCode = nextCode.ToString( "D" + numericLength );
-                }
-                else
+                // Let's just skip over this one...
+                if ( nextCode.ToString().Contains( "666" ) )
```
nairdo: I'm still reviewing all these code changes, but if you're going to skip number codes this way, you'll need to include 411, and 911 too.

chrismcgrath: This part of the code is actually totally broken, I reported the problem to Jon. Chose not to try to fix it as part of this PR, though.

dcs619: It seems inconsistent to check for specific strings here when the `NoGood` array already contains all the other combinations to avoid (including 666 and 911).

nairdo: I believe this is going to be changed to use 'contains' on the NoGood code checking.

SIDE NOTE:  Also, I'm creating new tests in the test project for testing AttendanceCodes -- and future PRs will also need to use-- and add new tests if necessary.  For now, after a dev runs the tests, they will need to have the 'Fact Skip' added back in so we don't break anyone's CI environment that is not set up with a db for testing purposes.  Later on, we'll figure out a way to keep these tests un-skipped so they will be run for everyone who's running the tests in CI.

nairdo: @chrismcgrath I'm now ready to try out your code (tomorrow) since we now have new tests (via 7d864d706756ab9deb17249a871e337d73b0bb49) for various edge cases.  If you want to run these tests in your dev environment, you can check out the new readme  (via 71d96e0c6f46a13d2f01bb2ab926f7b727a08937).

nairdo:
I'll close this PR for now but feel free to rebase against the develop branch and resubmit when you've had time to review and make some corrections.  Also try setting up a connection string in the Rock.Tests\app.config and then run the "skip" tests to verify.  Let me know if you have any questions about that stuff.

Issue #2650, opened 2018-01-05
Perform bulk updates in Parallel to improve performance


cabal95:
### Description

Currently the bulk update process runs sequentially. In theory it should be able to run multiple updates in parallel by processing chunks of people records at the same time. For example, if we are doing a bulk update on 10,000 people, currently that happens one after another. Person 1, then person 2, then person 3, and so on.

There should be no reason we can't operate in parallel, for example, running two tasks at the same time we would concurrently process:
* Person 1, Person 2, Person 3, ...
* Person 5001, Person 5002, Person 5003, ...

This should improve performance. Because we don't know how many CPUs the server has or what performance ability the server has there would be a new block setting (in Advanced) to set the number of parallel tasks, which would default to 0 to indicate "best safe guess". For example, if 4 CPUs are available then the default might be 4 parallel tasks. The admin can increase or decrease this by setting a value other than 0. For example, a value of 1 would revert to the current behavior. (@edmistj This is a change from what we discussed)

Some testing will need to be done to determine a safe default formula. Because of "waiting on the database" there is likely to be a lot of extra CPU cycles available per task. We don't want to go too low and leave the CPU idle and we don't want to go too high and max out the CPU for a long-running bulk update.

Also, some testing will need to be done to make sure no bulk update tasks could be modifying the same entity at the same time (for example, a group that the person is in). I don't believe any individual update options could do that but we will need to double check just to be sure and if so introduce locks on those type of updates so they don't happen at the same time.

**Expected behavior:** 

Bulk updates should be enjoyable.

**Actual behavior:** 

Bulk updates are like watching paint dry.

### Versions

* **Rock Version:** v7 alpha
* **Client Culture Setting:** en-US

arranf:
@cabal95 Is this something you're still working on?

cabal95:
Not actively, but it is still something I would like to hit at some point. If somebody else wants to take a stab before me I'd be happy to let them. I'll remove myself for the time being until I'm actively working on it.

PR #3119, opened 2018-07-10
Bulk Update block enhancements to performance and UI progress

cabal95:
## Proposed Changes

1. Update block to use multiple background threads when performing updates. By default it will use the number of "cpu threads", but the user can override with a block setting (up to 64). A "cpu thread" is the number of cores available, including hyperthreading. So a quad-core CPU with hyper-threading would use 8 worker threads. (See Note Below)
2. Updated block to use SignalR to give live feedback to the user about the progress. This also means no more timeouts when doing massive updates. Progress is given in 2.5 second increments.

Note on worker threads: It looks like C# managed code (i.e. Rock) is limited to a single CPU group by default. From what I understand of the docs, it looks like a dual-CPU system would have 2 CPU groups. This means the C# code (Rock) could only use 50% of the available system CPU. The "cpu thread" auto-calculation takes this into account so it will not over-subscribe if you have multiple sockets.

Fixes: #2650 

## Types of changes

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc
- [ ] I have added [REQUIRED tests](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests/README.md) to Rock.Tests that prove my fix is effective or that my feature works
- [ ] I have included necessary documentation (if appropriate)

## UI Example

bulk_update

## Further comments

I don't have a v8 massive database available for full-length testing. I was able to do some psuedo testing by introducing artifical delays and changing the batch size to 1; but I am going to see if somebody on v8 pre-alpha with a full database can test this by doing a massive update to see if it all works before I mark it as safe to merge.

## Documentation

Since we provide default value for the concurrent task count, no documentation should be needed.

### DIFF
--- a/RockWeb/Blocks/Crm/BulkUpdate.ascx
+++ b/RockWeb/Blocks/Crm/BulkUpdate.ascx
@@ -1,9 +1,45 @@
 ﻿<%@ Control Language="C#" AutoEventWireup="true" CodeFile="BulkUpdate.ascx.cs" Inherits="RockWeb.Blocks.Crm.BulkUpdate" %>
 
+<script src="/SignalR/hubs"></script>
+<script type="text/javascript">
+    $(function () {
+        var proxy = $.connection.rockMessageHub;
+
+        proxy.client.bulkUpdateProgress = function (completed, total) {
+            var $bar = $('#<%= pnlProcessing.ClientID %> .js-progress-bar');
+
+            $bar.prop('aria-valuenow', completed);
+            $bar.prop('aria-valuemax', total);
+            $bar.css('width', (completed.replace(',','') / total.replace(',','') * 100) + '%');
+            $bar.text(completed + '/' + total);
+
+            $('#<%= pnlProcessing.ClientID %> .js-progress-div').slideDown();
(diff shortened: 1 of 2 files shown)


cabal95:
Jim was kind enough to run some tests with this block. His test server was single-socket, dual-core VM so it processed with 2 threads in parallel and took about 20 minutes to run. Quoting him:

>It took about 20 minutes to run, pegged the CPU the whole time, but finished fine. I originally tried to update all 67k records but Rock wouldn't even load the bulkupdate page without blowing up with that many. While the CPU was pegged, Rock was still somewhat usable, but pretty sluggish (duh).

With that, I'm happy with the block as is. Though I will let the core team decide if the default worker thread count should be reduced to half the available cores (i.e. if you have 4 cores available, only use 2).

Alternatively, maybe on the "verify changes about to be made" screen, we should include an alert-warning box if they are updating more than 500 people letting them know this process may take a while and Rock's performance may be affected during the operation?

JimMichael:
Just for completeness sake, the test I successfully ran that took 20min was on 18,600 records.

PR #3218, opened 2018-08-27
Fix possible NullReferenceException when converting an Entity to a dictionary (Fixes #3165)

### LINKED ISSUE #3165: GET request to the PersonAlias endpoint fails.
### Steps to Reproduce
Issue a get request to the PersonAlias endpoint with the $expand=Person parameter.

**Expected behavior:**

I should received the PersonAlias JSON back with the Person information filled in.

**Actual behavior:**

An exception is thrown in Entity.cs in the ToDictionary() method because PersonAlias.AliasPerson doesn't have a visible getter.

### Versions
* **Rock Version:** 7.4 (1.7.4.58)

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1109768521967654">image.png</a>

Tschrock:
## Proposed Changes
Fixes a NullReferenceException in `Entity.ToDictionary()` that can occur when a property's getter isn't public (such as `PersonAlias.AliasPerson`).

Fixes: #3165

## Types of changes

* [X] Bugfix (non-breaking change which fixes an issue)

## Checklist

* [X] I have read the [CONTRIBUTING ](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc

## Further comments
N/A

## Documentation
N/A

### DIFF
--- a/Rock/Data/Entity.cs
+++ b/Rock/Data/Entity.cs
@@ -313,7 +313,8 @@ namespace Rock.Data
 
             foreach ( var propInfo in this.GetType().GetProperties() )
             {
-                if ( !propInfo.GetGetMethod().IsVirtual || virtualPropsWhiteList.Contains(propInfo.Name) )
+                MethodInfo getMethod = propInfo.GetGetMethod();
+                if ( getMethod != null && ( !getMethod.IsVirtual || virtualPropsWhiteList.Contains(propInfo.Name) ) )
                 {
                     dictionary.Add( propInfo.Name, propInfo.GetValue( this, null ) );
                 }


Rock/Data/Entity.cs
```diff
@@ -313,7 +313,8 @@ public virtual IEntity Clone()
 
             foreach ( var propInfo in this.GetType().GetProperties() )
             {
-                if ( !propInfo.GetGetMethod().IsVirtual || virtualPropsWhiteList.Contains(propInfo.Name) )
+                MethodInfo getMethod = propInfo.GetGetMethod();
+                if ( getMethod != null && ( !getMethod.IsVirtual || virtualPropsWhiteList.Contains(propInfo.Name) ) )
```
nairdo: Hi @Tschrock , since the virtualPropsWhiteList containing the propInfo.Name **is a valid condition of the _original_ IF block**, wouldn't this be more correctly written as:

```csharp
   if ( ( getMethod != null && !getMethod.IsVirtual ) || virtualPropsWhiteList.Contains( propInfo.Name ) )
```
...this way, even if the getMethod is null, if virtualPropsWhiteList contains the propInfo.Name, it can be added to the dictionary.  Of course, if propInfo.GetValue(...) returns null, then all bets are off again.

Tschrock: The problem with that logic is that the whitelist could potentially include properties that either
    1. have a private getter
or
    2. have no getter.

In case 1, it would be providing the value of a private property (something I assume you don't want to happen since it's private to start with).

In case 2, it would cause a new exception to be thrown (See the [microsoft documentation](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.propertyinfo.getvalue?view=netframework-4.7.2#System_Reflection_PropertyInfo_GetValue_System_Object_System_Object___): `PropertyInfo.GetValue(...)` will throw an `ArgumentException` if the property's getter is not found).

If you can guarantee that the whitelist will always include valid properties with public getters, then you could get away with that, but I'd rather be more defensive about things and remove the possibility of future issues.

nairdo:
Vince confirmed that Gerhard ended up fixing the underlying issue (#3165) this past October.

PR #3228, opened 2018-08-29
Fixes update queries

jreciomoreno:
Fixes a an error that gets thrown while executing the update queries checking for nulls or empty strings when the column type is different than a varchar.
Also added logic to check for null parameters being passed to the SQL command so these are passed as DBNull.values.

## Proposed Changes
The ProcessBIAnalytics job would fail to execute when processing non text/varchar attributes.
These code changes fix the null comparison performed on the update queries for any data type.

## Types of changes

## Checklist

* [ x] I have read the [CONTRIBUTING ](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc
* [x ] By contributing code, I agree to license my contribution under the [Rock Community License Agreement ](https://www.rockrms.com/license)

### DIFF
--- a/Rock/Jobs/ProcessBIAnalytics.cs
+++ b/Rock/Jobs/ProcessBIAnalytics.cs
@@ -1,4 +1,4 @@
-﻿// <copyright>
+// <copyright>
 // Copyright by the Spark Development Network
 //
 // Licensed under the Rock Community License (the "License");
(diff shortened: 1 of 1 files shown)


edmistj:
@jreciomoreno did you see any performance difference when running these adjustments?

nairdo:
@jreciomoreno After a small discussion here, someone spotted a potential issue with the change so it was reverted from v10/develop.  It appears it would always perform UPDATEs in situations where the value is NULL and continues to be null due to the `OR`:

```Sql
UPDATE ... WHERE...
AND ([{columnName}] IS NULL OR [{columnName}] <> @attributeValue)
```

vs

```Sql
UPDATE ... WHERE...
isnull([{columnName}],'') != @formattedValue
```

PR #3303, opened 2018-09-26
Update spCheckin_BadgeAttendance.sql

Adroken:
Optimized stored procedure to return in less than a second even with large quantities of attendance data.

## Proposed Changes

After importing attendance data we noticed a large slowdown in performance when loading person profiles.  This was tracked back to the attendance badge stored procedure.  I re-wrote the stored procedure to be more efficient.  
I removed the large if structure and replaced it with a case statement inside the familyMemberPersonIds temptable.  I also added another column to that temptable in order to indicate which entry in the table is the requesting person.
In addition I reversed the logic to show only the person if they are a child and the entire household if they are not a child.  The reasoning behind this is because we added several other adult categories which would be a more common occurrence than adding more child categories.
Finally, I noticed that there was a possible logic gap with the setting of startday and endday and moved the definition of startday below where endday is checked for having a date not in the future.  This will prevent the possibility of a startday that is after the endday.

Fixes: #

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc
- [X] Unit tests pass locally with my changes
- [N/A] I have added [REQUIRED tests](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests/README.md) to Rock.Tests that prove my fix is effective or that my feature works
- [X] I have included necessary documentation (if appropriate)

### DIFF
--- a/database/Procedures/spCheckin_BadgeAttendance.sql
+++ b/database/Procedures/spCheckin_BadgeAttendance.sql
@@ -18,7 +18,7 @@
 	<param name="Reference Date" datatype="datetime">A date in the last month for the badge (optional, default is today)</param>
 	<param name="Number of Months" datatype="int">Number of months to display (optional, default is 24)</param>
 	<remarks>	
-		Uses the following constants:
+		Looks up the following constants:
 			* Group Type - Family: 790E3215-3B10-442B-AF69-616C0DCB998E
 			* Group Role - Adult: 2639F9A5-2AAE-4E48-A8C3-4FFE86681E42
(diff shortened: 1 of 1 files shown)


Adroken:
@nairdo I've made the requested changes.  Please let me know if I need to do anything different to re-submit.

Thanks for the quick response!

PR #3365, opened 2018-10-24
ExceptionList will now show the most recent description for each exception

chrismcgrath:
## Proposed Changes
The ExceptionList block groups exceptions by the first 28 characters of the description, so each exception in the group can have different messages. The query that gets the exception list displays the most recent date but may not display the description from the most recent exception, which can be confusing when the description shown is for an old error that should no longer be happening, but it looks like it happened recently. 

This change will correct this so the most recent description is shown for each grouping.

## Types of changes

* [X] Bugfix (non-breaking change which fixes an issue)

## Checklist

* [X] I have read the [CONTRIBUTING ](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc


nairdo:
Thanks @chrismcgrath !

PR #3387, opened 2018-10-30
Fixed lava where filter to work with arbitrary data

trydyingtolive:
## Proposed Changes
When trying to sort a list of objects that came from a JSON source ( using `|FromJson`), `|Where` doesn't work if you are filtering by an integer.  This makes filtering by Id or some other number challenging. The reason is that JsonConvert.DeserializeObject converts all numbers as long and it doesn't match the provided int. I have changed the way the two objects are compared so that the expected outcome happens. Unit tests included.

## Types of changes

## Checklist

* [x] I have read the [CONTRIBUTING ](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc
* [x] I have added [REQUIRED tests ](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests/README.md) to Rock.Tests that prove my fix is effective or that my feature works

## Further comments
Related StackOveflow: https://stackoverflow.com/questions/27972266/compare-numeric-types-when-types-are-unknown

I also considered `.ToString()`-ing both sides. This would have the added benefit of matching strings to integers, with the downside of it just not feeling right.

Additionally here is some test lava:

```
{% capture str %}
[
    {"Id": 1, "Name":"Chuck"},
    {"Id": 2, "Name":"Mark"}
    ]
{% endcapture %}

{% assign json = str | FromJSON | Where: "Id", 1 %}

<ul>
{% for item in json  %}
<li>{{item.Name}}</li>
{% endfor %}
</ul>
```

### DIFF
--- a/Rock.Tests/Rock/Lava/RockFiltersTests.cs
+++ b/Rock.Tests/Rock/Lava/RockFiltersTests.cs
@@ -741,6 +741,58 @@ namespace Rock.Tests.Rock.Lava
 
         #endregion
 
+        #region Where
+
+        /// <summary>
+        /// For use in Lava -- should extract a single element form array
+        /// </summary>
+        [Fact]
+        public void Where_ByInt()
+        {
+            var input = new List<Dictionary<string, object>>
+            {
+               new Dictionary<string, object> { { "Id", (int)1 } },
+               new Dictionary<string, object> { { "Id", (int)2 } }
+            };
+            var output = RockFilters.Where( input, "Id", 1 );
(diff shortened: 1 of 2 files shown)


trydyingtolive:
@nairdo I checked, it seems we are not running it in production. We ran into some of these issues before I started on the fix, so we used some work arounds. I mostly made these changes because of some interesting lava output Chris Rae had showed me, not rather than a need for SECC. So there wasn't a pressing need for us to cherry-pick it like we would normally do.

PR #3428, opened 2018-11-15
Require all address fields when address validation is required

jreciomoreno:
## Proposed Changes
Fixes: The validation on the address control that currently isn't validating all fields if its marked as required.

## Types of changes

## Checklist

### DIFF
--- a/Rock/Web/UI/Controls/AddressControl.cs
+++ b/Rock/Web/UI/Controls/AddressControl.cs
@@ -151,6 +151,14 @@ namespace Rock.Web.UI.Controls
             set
             {
                 ViewState["Required"] = value;
+                EnsureChildControls();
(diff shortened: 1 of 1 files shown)


nairdo:
@jreciomoreno Although I don't believe we can currently support requiring a Zip/PostalCode, I believe this fix (https://github.com/SparkDevNetwork/Rock/commit/849c22d12e2789fa94f5fa7192e3c300399ba987) for Rock v8.7 should largely address the issue, would you agree?

We'll close this PR, but please re-open an issue if you decide the PostalCode must also be required when the Address attribute is required. In that case, the solution that gets worked out will also need to account for international addresses that may or may not have postal codes (see https://en.wikipedia.org/wiki/List_of_postal_codes).

PR #3433, opened 2018-11-16
Add Connection Request Attributes #3421

### LINKED ISSUE #3421: [EN] Attributes on Connection Requests
# Feature Request
**Describe the Feature Request**
I would like to be able to use Attributes on Connection Requests and have them neatly displayed and editable.  My use case is for storing information related to Follow-up requests from guest families.  It can be stored as person attributes but then requires additional clicks to find.

**Describe Preferred Solution**
I would prefer to be able to add attributes and have them show up in the Connection Request Detail block and be editable.  They should also be accessible via lava particularly within related workflows.

**Describe Alternatives**
The Entity Attribute Value block can support this but it is not positioned well and requires two edit buttons.

**Providing Code**
I will provide code if feature request is approved: working on it, so hopefully.

bobrufenacht:
## Proposed Changes
Adds Attribute display and edit to Connection Requests.
Closes #3421 

## Types of changes

* [X] New feature (non-breaking change which adds functionality, which has been approved by the core team)

## Checklist

* [X] I have read the [Contributing to Rock ](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc
* [X] Unit tests pass locally with my changes

## Documentation
Not sure if this is needed or not.  Default has no visible change.  If Connection Request Attributes are added using Entity Attributes, they will be displayed on the Connection Request Detail block below the Comment and above the Edit button and on the Edit dialog they will show above the Save button.


JimMichael:
Just a curiosity, but will this new functionality allow the new attributes to be exposed ("asked-for") in the public-facing Connection Request form?

bobrufenacht:
If I understand the question, yes, when editing a connection request the attributes will be displayed and editable on the edit screen.

JimMichael:
Ok, cool. By the description it sounded like they were only exposed/editable from the internal connections tools, not the public facing form that users fill out to create the connection itself. It will be nice to be able to get more information from the end user when they submit a connection!

bobrufenacht:
I did misunderstand your question.  I did not update the public facing block but I’d be happy to do so.  At present. The Connection Request Sign-Up block has minimal information.  We would need a way to specify which attributes would be on the public page and I didn’t think the Show in Grid option was appropriate to use that way.  It might need to be a block setting to indicate which attributes are optional or required on the public page.  Suggestions on how to implement that?

JimMichael:
Yeah, I think it would need to follow the pattern of the Family Pre Registration or Public Profile Edit blocks, where you select the attributes you want to expose... because you need a way to have "internal only" attributes that aren't publicly shown.

PR #3507, opened 2019-01-22
Create Person_BirthDate.sql

Adroken:
## Proposed Changes

The current BirthDate column in the Person table is not able to be persisted.  This new calculation method not only mirrors the existing computations (and uses slightly less CPU) but allows for it to be persisted which enables the ability for indexes as well as resulting in a speed increase from 2000ms to 500ms at scale.

Fixes: #

## Types of changes

* [X] New feature (non-breaking change which adds functionality, which has been approved by the core team)

## Checklist

* [X] I have read the [Contributing to Rock ](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc
* [X] Unit tests pass locally with my changes

### DIFF
--- a/database/ComputedColumns/Person_BirthDate.sql
+++ b/database/ComputedColumns/Person_BirthDate.sql
@@ -1,0 +1,5 @@
+ALTER TABLE rms.dbo.Person
+	DROP COLUMN BirthDate
+
+ALTER TABLE rms.dbo.Person
+	ADD BirthDate AS DATEFROMPARTS(BirthYear,BirthMonth,BirthDay) PERSISTED


edmistj:
Hi @Adroken! Can you provide some details on the nature of this change?

Adroken:
> Hi @Adroken! Can you provide some details on the nature of this change?

Sure!  So when doing check-in or checking requirements I noticed the database wanting to create an index including the birthdate calculated column.  This was not possible because the column is not persisted and the current function does not allow it to be persisted because SQL insists it could be different each time you run it.  Changing the function from the case statement with conversions into the DATEFROMPARTS() function results in the exact same end result, but allows the column to be set to persisted (which stores the value in the database until the underlying source is changed) and greatly improves the speed with which the data is returned as well as allowing the indexing of this column to further improve query performance.

edmistj:
Thanks, that helps a lot. We need this level of detail to ensure we understand the changes that are being proposed.

edmistj:
@nairdo should have Mike review.

nairdo:
We've discussed and identified an issue:

- If there are bad dates in the system (such as 2/29/2021), the migration (that would implement the column change) will fail because it can't create the computed column.  We can't have that situation.
- Also, when there is an invalid date in the system, that DATEFROMPARTS function will throw an error and you will not be able to load the person record.  However, even the current system has some issues when attempting to view a record with illegal birthdates.

The primary issue with this new code is: any corresponding migration would need to first identify and correct the data before it attempted to alter the BirthDate column.  Perhaps someday we can revisit this idea.

Adroken:
The migration could go with the original intent of splitting the month/day/year into integer columns which allowed the incorrect data and simply drop the year from the birthdate of any invalid birthdate combination.  This would allow the column to remain NULL while retaining the birthdate (without year unfortunately) of the individual.
Also I can’t seem to enter bad data into the front-end thankfully.
After having the change in place it also prevents me from updating the columns with bad data (throws an error and rolls back the transaction)
If you would like, I would be happy to write a script to appropriately cleanse out any bad birthdates by either removing the birth year or by setting the out of range birthdates to the last day of the month (this appears to be what the rock front-end does currently).  I’m thinking the source of most of these would be imported data as rock seems to be doing an excellent job safeguarding the validity of its data.

Thanks and let me know!

Adroken:
Actually I just wrote the script to identify and set the birthyear to null on invalid birthdays.
One that just corrects to the last day of the month would either be more complicated or just split out the logic into multiple update statements

`UPDATE P
SET birthyear = NULL
from rms.dbo.person p where
(birthmonth = 2 and birthday>28 and NOT ((BirthYear % 4 = 0 AND BirthYear % 100 <> 0) OR BirthYear % 400 = 0)) --Non Leap Years
or
(birthmonth = 2 and birthday>29 and ((BirthYear % 4 = 0 AND BirthYear % 100 <> 0) OR BirthYear % 400 = 0)) --Leap Years
or 
(Birthmonth in (9,4,6,11) and birthday>30) --Months with 30 days
or
(Birthday>31) -- Too many days
or 
(Birthmonth>12) -- Too many months
`

Adroken:
image
Here is our 11:30 service from before we implemented this change.

image 1
Here is our 11:30 service after implementing the change.  Please note the change in scale.  Peak requests per second were constant between the two periods at approx 1.1k

The only additional change you will notice is we have another query that is running every 15 minutes that I asked to move from the exact 15 minute ticks of the hour to 11:10, 11:25, etc.

@nairdo if you have an example migration script that I can pop that SQL into or if you would like me to re-write it to change the date to the last day of the month instead of wiping out the year on dates that have day issues I would be more than happy to.  Please let me know and I would be happy to assist with fully providing everything needed for this change.

edmistj:
@Adroken love the performance improvement your showing. We're thinking a trigger might be able to do the same thing and be safer if someone did have bad data due to a manual SQL update. You'd be surprised the things people do manually. What are your thoughts? Either way we will be adding the clean-up script you provided as a good clean-up step.

We do something similar when calculating AtrributeValue DateTime.

Thoughts? Again, love the effort in improving the performance.

Adroken:
@edmistj Sorry for the delay in response - was super busy over this way last week!
I think a trigger is a great solution to bring the best of both worlds to this issue.  Do you guys have one already fleshed out for this or would you like me to modify this request with one?

Thanks again!

edmistj:
@Adroken We don't yet have a trigger configured. If you have time to update this PR with one that would be amazing. If not let us know and we'll get something before the next release. The performance gains you show above are great!

MikeDPeterson:
We have a very similar thing already done for AttributeValue.ValueAsDateTime, so we'll just go ahead and code it and implement it

MikeDPeterson:
FYI, after running into a little trouble with the trigger approach, landed on a solution closer to @Adroken's original solution (calculated persisted). However, changed to TRY_CONVERT to avoid the risk of throwing exceptions on bad dates. So, it'll end up being something like this:

```
ALTER TABLE [Person]
DROP COLUMN BirthDate


-- Calculate Birthdate using TRY_CONVERT to guard against bad dates, and use 126 style so that it'll parse as style 126 (ISO-8601) which is deterministic which will let it be persisted
ALTER TABLE [Person] ADD [BirthDate] AS CASE 
		WHEN BirthYear IS NOT NULL
			THEN TRY_CONVERT(DATE, CONVERT(VARCHAR, [BirthYear]) + '-' + CONVERT(VARCHAR, [BirthMonth]) + '-' + CONVERT(VARCHAR, [BirthDay]), 126)
		ELSE NULL
		END PERSISTED


CREATE INDEX [IDX_BIRTHDATE] ON [Person] ([BirthDate])
```

sparkdevnetwork-service:
➤ grace hansen commented:

@ nick - Please assess and assign priority

sparkdevnetwork-service:
➤ Nick Airdo commented:

Jon Edmiston ( https://app.asana.com/0/20866866924296/list ) I looked at this with Mike.  His concern is: "It is not going to work."  He explained, when an invalid date occurs, that DATEFROMPARTS function will throw an error and you will not be able to load the person record.

Example: someone with a birthday of 2/29/2021 will cause the error. 

I added someone to my db with that birth date via an import from another system and he's correct that this throws an error...

SELECT DATEFROMPARTS(BirthYear,BirthMonth,BirthDay) FROM [Person]

returns:

Msg 289, Level 16, State 1, Line 1
Cannot construct data type date, some of the arguments have values which are not valid.

** On NewSpring's system this runs fine across all 424k people records indicating they have no invalid birth dates.

However, even without that proposed change my Rock no longer functions correctly while trying to view (or being logged in as) a person with an invalid birth date.  You get this exception:

ArgumentOutOfRangeException in mscorlib
Message
Year, Month, and Day parameters describe an un-representable DateTime.

Do we need to talk with Mike together, or just roll forward with that change?

sparkdevnetwork-service:
➤ Nick Airdo commented:

...Additionally I just discovered attempting to run that proposed ALTER TABLE dbo.Person will fail in the event that someone's database has bad birth dates in it.

So, this will a cause migration issue unless we first find and fix the bad data before the ALTER occurs.

sparkdevnetwork-service:
➤ grace hansen commented:

Nick Airdo ( https://app.asana.com/0/21779865363458/list ) -added a new stage to flag tasks that require a technical call. Thanks for all the good background included.

sparkdevnetwork-service:
➤ Jon Edmiston commented:

I would add the details here to the PR and close noting that we'd be open to an alternate solution if one exists.

sparkdevnetwork-service:
➤ grace hansen commented:

Nick Airdo ( https://app.asana.com/0/21779865363458/list ) - We are on-hold for input regarding the proposed solution from the community developer who provided the PR code.

sparkdevnetwork-service:
➤ Nick Airdo commented:

This report is from Life.Church. Jon and I talked about this just now and we're going to have Mike on this for part of the remaining time we've allocated to the Life.Church performance project.

Additionally, Jon also identified the GivingLeaderId column has significant performance implications too so we're going to re-engineer that one as well -- after the birthdate column change.

We are considering this to be part of our new "Add a Zero" initiative.

sparkdevnetwork-service:
➤ Mike Peterson commented:

fixed with https://github.com/sparkdevnetwork/rock/commit/ebd38dc323bcf55b612a1b6e9a957ee50a94c772

sparkdevnetwork-service:
➤ grace hansen commented:

Nick Airdo ( https://app.asana.com/0/21779865363458/list ) - Mike marked this complete. Do we want this fix to go through QA?

sparkdevnetwork-service:
➤ Nick Airdo commented:

grace hansen ( https://app.asana.com/0/958164134519812/list ) It is just due to it being connected on the Github PR system which needs to be closed (otherwise it could inadvertently be merged into the develop branch).  Since you would like it open, I will make a new subtask here for QA with the link to the commit that can be tested.

PR #3539, opened 2019-02-13
Add DataView API Endpoint

richarddubay:
## Proposed Changes
This change adds a Dataview API Endpoint to pull the persisted data views that a person belongs to. This would be beneficial for the Apollos Project to help us pull personalized content based on personas/data views into a feed or some other area of the app as quickly as possible.

I'd love for this PR to be the start of a discussion if necessary. Does this functionality work with the vision of Rock? Am I missing anything? What did I do that could be made better or faster? Etc.

## Types of changes

_NOTE: I can't say this has been approved by the core team. But it is a recommendation to add to the core Rock product._

## Checklist

* [x] I have added [required tests ](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests/README.md) that prove my fix is effective or that my feature works
* [x] I have included updated language for the [Rock Documentation ](https://www.rockrms.com/Learn/Documentation) (if appropriate)

### DIFF
--- a/Rock.Rest/Controllers/DataViewsController.Partial.cs
+++ b/Rock.Rest/Controllers/DataViewsController.Partial.cs
@@ -1,0 +1,66 @@
+﻿// <copyright>
+// Copyright by the Spark Development Network
+//
+// Licensed under the Rock Community License (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.rockrms.com/license
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+//
+using System.Collections.Generic;
+using System.Linq;
+using System.Web.Http;
+
+using Rock.Data;
+using Rock.Model;
+using Rock.Rest.Filters;
+
+namespace Rock.Rest.Controllers
+{
+    /// <summary>
+    /// DataViews REST API
+    /// </summary>
+    public partial class DataViewsController
+    {
+        #region DataViewByPerson
+
+        /// <summary>
+        /// Returns a list of dataviews that a person is a part of
+        /// </summary>
+        /// <param name="entityTypeId"></param>
+        /// <param name="entityId"></param>
+        /// <param name="categoryGuid"></param>
+        /// <param name="categoryId"></param>
+        /// <returns></returns>
+        [HttpGet]
+        [Authenticate, Secured]
+        [System.Web.Http.Route( "api/DataViews/GetPersistedDataViewsForEntity/{entityTypeId}/{entityId}" )]
+        public IQueryable<DataView> GetPersistedDataViewsForEntity( int entityTypeId, int entityId, System.Guid? categoryGuid = null, int categoryId=0 )
+        {
+            var rockContext = new RockContext();
+            // Get the data view guids from the DataViewPersistedValues table that the Person Id is a part of
+            var persistedValuesQuery = rockContext.DataViewPersistedValues.Where( a => a.EntityId == entityId && a.DataView.EntityTypeId == entityTypeId );
+            IQueryable<DataView> dataViewList = persistedValuesQuery.Select( a => a.DataView);
+            if (categoryGuid != null)
+            {
+                dataViewList = dataViewList.Where( a => a.Category.Guid == categoryGuid );
+            }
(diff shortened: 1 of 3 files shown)


edmistj:
Couple of thoughts:

- Since the endpoint returns data view information I think it should be in the DataView controller.
- We could abstract this a bit and pass in EntityTypeId and EntityId and now it can be used for all Data Views
- The name should acknowledge that it only looks at persisted Data Views something like GetPersistedDataViewsForEntity
- The return type should be the DataView model but if you return it as a IQueryable then OData should kick in and allow you to configure it to return just the Guid if that's what you want. Most people would want more so using the IQueryable you allow the caller to determine the return shape.
- The route should match the action name so /api/DataViews/GetPersistedDataViewsForEntity/{entityTypeId}/{entityId}?categoryGuid=[optional] (might be nice to also allow for CategoryId)
- I wouldn't do the check for the Person. Filtering by the EntityTypeId and EntityId should be enough.

Does that make sense. Others are welcome to chime in.

IsaacHardy:
@edmistj It seems that the DataViewsController is a CodeGenerated controller. Should we go about adding to it in a certain way based on its inherent automation?

richarddubay:
@edmistj These are great suggestions! We'll make these for sure.

edmistj:
@IsaacHardy yes you would implement them as a partial class. You can use the others as a pattern.

cabal95:
Just to add some extra details to what Jon said... the files in the `Rock.Rest/Controllers/CodeGenerated` directory are indeed automatically generated and shouldn't be modified. Because they are `partial` classes, you can define another `partial` class of the same name that has "user code" in them - these are the ones in the `Rock.Rest/Controllers` directory.

Since the `DataViewController.cs` in the `Rock.Rest/Controllers` doesn't exist yet, you would just create it. The `AuthController.cs` is probably your best one to look at as a template as it's basically empty (the one method it does have, you can rip out and replace with your own).

Hope this helps!

richarddubay:
@edmistj @IsaacHardy @cabal95 So one of the things that we're going to have to figure out is how to get this API working in a 1.8.5.2 environment (because that's what the Apollos Project is using as it's version of Rock at the moment ... we always want to make sure that we are compatible with whatever is the latest release version). 

The way that we wrote and are writing the code for this particular PR will work for where `develop` is and moving forward, but not at the level of the latest release. We're just getting back into Rock development lately so I'm not exactly sure when the problem will go away, but currently the 1.8.5.2 release doesn't like the querystring parameters on the endpoint. In particular the "?". So we're going to write this twice, I guess. Once for `develop` and once for `master`. 

Maybe this is a dumb question but I'm going to ask it anyway. Would it be beneficial for us to write this particular endpoint for this PR in a way that is compatible with the current release? Or just to write it twice? Our current plan is just write it twice and to move the `Rock.Rest.dll` with the "current release" way into our Rock instance so that we can use it now and then just let this PR eventually override the old endpoint when it gets released. Does that even make sense? Thoughts? Maybe this is a discussion better had someplace else?

richarddubay:
Actually, now that I'm looking at it again ... maybe the querystring stuff happens naturally when we pass parameters to the function. So the answer is just that I don't need to specify it in the route.

cabal95:
> maybe the querystring stuff happens naturally when we pass parameters to the function. So the answer is just that I don't need to specify it in the route.

I'm 99% sure is correct. Any query string parameters that match a function parameter name are automatically populated. You just need to provide a default value (e.g. `null`) so that you can detect when they didn't specify one and handle it appropriately (either with an error or by searching everything instead of that single category).

richarddubay:
@edmistj @cabal95 Hey gents! So @IsaacHardy and I got this moved to the DataViewController and we've made all the suggested updates as far as we can tell. The only thing that is causing us issues now is that the OData stuff doesn't seem to want to work. We can't run a `$select` and have it return just one thing. I'm not sure how to make that happen.

cabal95:
Meaning you tried something like `/api/DataView/GetPersistedDataViewsForEntity/10/42?categoryId=8&$select=Name` but it still returned a full DataView object?

richarddubay:
@cabal95 That's exactly right.

cabal95:
@richarddubay Okay, looks like you need to add a `[EnableQuery]` attribute on the method (similar to the `HttpGet` one you have) to enable the OData query filtering support.

Learned something new myself!

richarddubay:
@cabal95 That was a genius catch. Thank you so much!

richarddubay:
@edmistj @cabal95 Thanks so much for all your help in making this API endpoint so good. I really do appreciate it! Since this is a pretty custom endpoint, I'm going to go ahead and close this PR and attempt to move this to a plugin sometime down the road. Thanks again for all your help!

PR #3545, opened 2019-02-18
Adding a flag for allowing to not replace unsafe sender email addresses.

cbump:
## Proposed Changes
Mailgun automatically supports relaying email on behalf of any email address.  This PR will update the preferences for the MailGun transport to allow for the Safe Sender list to simply put a h:Sender header on the email and allow anything else to still be relayed.  Mailgun will use the "on Behalf of" functionality: https://help.mailgun.com/hc/en-us/articles/360012491394-Why-do-I-see-On-Behalf-Of-in-my-email-

## Types of changes

## Checklist

## Further comments
We have group leaders that author and send emails from their personal email addresses.  Mailgun handles this pretty well by relaying email "on behalf of" or "via" whatever email address is set as the "From".  Additionally the 'h:Sender' header completely hides the fact that we are using Mailgun for our internal domain.  For other mail relays such as straight SMTP it makes sense to replace the sender but for Mailgun, some organizations may elect to allow relaying of email on any domain as long as Mailgun will properly handle it (which it seems to do quite well).

## Documentation
It might be worth putting a note about this setting in the Mailgun section on step 5 in this documentation here: https://www.rockrms.com/Rock/BookContent/8#integrations  It also might be worth mentioning in the Safe Sender Domain section here: https://www.rockrms.com/Rock/BookContent/8#safesenderdomains


cabal95:
@cbump Our mailgun domain is set as `mg.shepherdchurch.com` so that incoming e-mails to `shepherdchurch.com` still go through our normal server, but any incoming messages we want handled by a workflow are passed to Rock. Which means _all_ e-mails sent by Rock do the "On behalf of" thing since our staff e-mails come from the "wrong" domain.

I thought we could fix this by adding a second domain to mailgun, which it let us do, but each domain has it's own set of API credentials so it still does the "on behalf of" thing.

As far as you know, would this PR solve our problem too?

cbump:
@cabal95 The "h:Sender" portion of this PR (probably the single most important part) should solve the problem you are seeing.

nairdo:
Hi Chuck @cbump, I spent some time trying to understand the 'what, how and why' behind this PRs  "h:Sender" mailgun trick, but it's not clear.  Can you please clarify or correct my understanding, and/or attache some examples showing the final effect it has on the client/recipient?

From what I've been able to tell so far, when h:Sender is included in the mailgun HTTP handler ([as per their documentation](https://documentation.mailgun.com/en/latest/api-sending.html#sending)), it will cause a "Sender:" header to be included which [according to IANA](https://www.iana.org/assignments/message-headers/message-headers.xhtml)'s [RFC5322] section 3.6.2, (http://www.iana.org/go/rfc5322). In that RFC, it specifies:

> If the originator of the message can be indicated by a single mailbox and the author and transmitter are identical, the "Sender:" field SHOULD NOT be used.

In this PR it looks like:

1. The Sender: header will now always be added _IF_ the sender is safe (in the "safe sender" domain list).
2. If the sender is not safe... and _IF_ the new "Replace Unsafe Sender" transport attribute is set to N/False, the from address will not be replaced with the organization's safe email address and the "Reply-To" header will not be added to the email. (...essentially bypassing the entire SafeSender check, right?)

cbump:
Hey @nairdo I basically followed this guide which uses the h:Sender header + the From header in safe sender situations in order to determine what to do with the  "On behalf of" situation with relaying email through mailgun: https://help.mailgun.com/hc/en-us/articles/360012491394-Why-Do-I-See-On-Behalf-Of-in-My-Email-

If you set "Replace Unsafe Sender" to false, then it just sends it directly any email address with the "On Behalf Of" notation.  The Reply-To is unnecessary in this situation because it's actually relaying it for the original sender.

cbump:
Here is what the email looks like if it's sent from another domain:
image
Here is what it looks like if it's sent from secc.org:
image
Prior to setting the h:Sender header, ALL emails displayed including those from secc.org displayed the "On Behalf Of" notation.

nairdo:
Hi @cbump Chuck,  Just an FYI... We will be merging this into to Rock in the next week or so and expect that it will be in either pre-alpha-release 1.11.0.14 or 1.11.0.15.

matthewbronkema:
@cbump, can you merge in the latest develop changes into your branch to resolve conflicts?

PR #3571, opened 2019-02-28
Added Content Channel Items Controller

richarddubay:
## Proposed Changes
This change add a Content Channel Item API Endpoint to pull content channel items from a comma delimited list of guids that come from data views. This would be beneficial to the Apollos Project because it would help us pull personalized content based on personas/data views. 

I'd love for this PR to be the beginning of a discussion. Does this work with the vision of Rock? Am I missing anything? What could we do to make this better or faster?

## Types of changes

_NOTE: I can't say that this has been approved by the core team. But it is a recommendation to add to the core Rock product._

## Checklist

## Further comments
I have another possible solution that I will paste here for discussion. This one uses less code but I am not sure of it's performance. I'll just drop it here and then we can talk about it.

```
RockContext rockContext = new RockContext();

List<string> guidList = ( guids ?? "" ).Split( ',' ).ToList();

IQueryable<int?> attributeValueEntityIdList = new AttributeValueService( rockContext ).Queryable()
                .Where( av => guidList.Contains( av.Value ) )
                .Select( av => av.EntityId );

List<ContentChannelItem> contentChannelItemList = new ContentChannelItemService( rockContext ).Queryable()
                .Where( cci => attributeValueEntityIdList.Contains( cci.Id ) ).ToList();

return contentChannelItemList;
```

### DIFF
--- a/Rock.Rest/Controllers/ContentChannelItemsController.Partial.cs
+++ b/Rock.Rest/Controllers/ContentChannelItemsController.Partial.cs
@@ -1,0 +1,65 @@
+﻿// <copyright>
+// Copyright by the Spark Development Network
+//
+// Licensed under the Rock Community License (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.rockrms.com/license
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+//
+using System.Collections.Generic;
+using System.Linq;
+using System.Web.Http;
+
+using Rock.Data;
+using Rock.Model;
+using Rock.Rest.Filters;
+
+namespace Rock.Rest.Controllers
+{
+    /// <summary>
+    /// ContentChannelItems REST API
+    /// </summary>
+    public partial class ContentChannelItemsController
+    {
+        #region ContentByDataViewGuids
+        /// <summary>
+        /// Returns a list of content based on list of dataview guids
+        /// </summary>
+        /// <param name="guids"></param>
+        /// <returns></returns>
+        [HttpGet]
+        [Authenticate, Secured]
+        [System.Web.Http.Route( "api/ContentChannelItems/GetFromPersonDataView" )]
+        public List<ContentChannelItem> GetFromPersonDataView( string guids )
+        {
+            var rockContext = new RockContext();
+
+            // Turn the comma separated list of guids into an array of strings
+            string[] parsedGuids = guids.Split( ',' );
+
+            // Get any attribute values with a value that matches a passed in guid
+            var attributeValuesPerGuid = new AttributeValueService( rockContext ).Queryable().Where( a => parsedGuids.Contains( a.Value ) );
+
+            // Only get the EntityIds - these are the content channel items
+            List<int?> mainEntityIds = attributeValuesPerGuid.Select( a => a.EntityId ).ToList();
+
+            // Get any content channel items with an id that matches an entity id from the mainEntityIds list
+            var contentChannelItems = rockContext.ContentChannelItems.Where( a => mainEntityIds.Contains( a.Id ) );
+
+            // Turn that Queryable into a list
+            IQueryable<ContentChannelItem> finalList = contentChannelItems.Select( a => a );
+
+            // Return list
+            return finalList.ToList();
+        }
+        #endregion
+    }
+}
--- a/Rock.Rest/Rock.Rest.csproj
+++ b/Rock.Rest/Rock.Rest.csproj
@@ -124,6 +124,7 @@
     <Compile Include="Controllers\CodeGenerated\NcoaHistoriesController.cs" />
     <Compile Include="Controllers\CodeGenerated\PageShortLinksController.cs" />
     <Compile Include="Controllers\CodeGenerated\PersonTokensController.cs" />
+    <Compile Include="Controllers\ContentChannelItemsController.Partial.cs" />
     <Compile Include="Controllers\ContentChannelItemSlugsController.Partial.cs" />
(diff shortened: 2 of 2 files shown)


edmistj:
Not sure what you're trying to achieve here. What do the GUIDs represent? 

What's here would definitely not perform well as you're searching through every attribute value on the system. You'll want to filter that down by some pattern of AttributeType, EntityType, EntityId. 

You might back up a bit and document our the problem you're trying to solve.

richarddubay:
Hey Jon ( @edmistj )! I can definitely speak to what we're attempting to do. 

We're working on trying to solve the problem of content personalization in an app (the Apollos Project) using Rock. The solution that we've come up with as an MVP is to create data views as "personas" and then tie content channel items to those data views using an item attribute. In our app, we are trying to pull only those content channel items that are associated with the data views that a particular person is a part of. 

One part of this is pulling the data views that a person is a part of. This is what we are attempting to do in our [other PR](https://github.com/SparkDevNetwork/Rock/pull/3539). From the data views that are returned, we take the `guid` because that is what we found in the AttributeValue table `Value` field that actually ties us to the content channel item.

We did attempt to take those data views and get the attribute values as one call and then do another call which got the content channel items based on the attribute values. The problems we ran into doing it this way were (1) it took too long to attempt to `foreach` through each of the attribute values and pull the content channel item one by one, and (2) we couldn't just pass all of the attribute values at one time and get back a list of all the content channel items ... we ran into a limit with "nodes."

So our solution tries to combine the attribute value and content channel item calls into one. It takes a list of `guids` that represent data views that a person is a part of and spits out the content channel items that are tagged with those data views. Specifically, we search through the AttributeValues for anything with a `Value` that matches a `guid` that we passed in. We take the `EntityId` from this list because it matches to the `Id` of the content channel item. Then we get the content channel items from the list of `Ids`. 

I definitely think if we can somehow filter the AttributeValues table before searching it, that would be beneficial. I can take a look to see if there is something we can use (maybe AttributeType or EntityType like you suggested). 

If there is a better way for us to achieve this goal, I'd love to know that too. If there is already an existing call that we could take advantage of or a better way to code what we're trying to do, we're all ears. It's been a long time since we really developed anything for Rock so we could be using bad practices and not even be aware of it. 

Thanks for all your help!

edmistj:
OK that makes sense. The key will be to filter as much as you can before you query though the Attribute Values. 

You'd could add to your query a filter on AttributeType and also EntityType. That would significantly reduce the number you'd need to filter through. It's also the contains that will show it down, but there's probably not much you can do about that. 

We have some other ideas about how to filter for personalized content coming, but it's not ready yet.

richarddubay:
@edmistj Okay, I updated the endpoint to filter by both the EntityType and the AttributeType (FieldType) before it tries to search through the AttributeValues table. This does seem to make the call faster from my testing. Now granted, I didn't do benchmarks or anything. 😂

Side Note: I'm not sure why Appveyor doesn't like this now. I merged in the latest from `develop`to try to fix the errors, but it didn't seem to fix it. 🤷‍♂️

Rock.Rest/Controllers/ContentChannelItemsController.Partial.cs
```diff
@@ -0,0 +1,80 @@
...
+            RockContext rockContext = new RockContext();
+
+            // Turn the comma separated list of guids into a list of strings.
+            List<string> guidList = ( guids ?? "" ).Split( ',' ).ToList();
+
+            // Get the Id of the Rock.Model.ContentChannelItem Entity.
+            int contentChannelItemEntityTypeId = EntityTypeCache.Get( "Rock.Model.ContentChannelItem" ).Id;
+
+            // Get the Field Type (Attribute Type) Id of the Data View Field Type.
+            int fieldTypeId = new FieldTypeService( rockContext ).GetByName( "Data Views" ).Select( ft => ft.Id ).FirstOrDefault();
```
nairdo: You might want to fetch this from cache via:
   `int fieldTypeId = FieldTypeCache.Get( Rock.SystemGuid.FieldType.DATAVIEWS.AsGuid() ).Id;`

edmistj:
I believe you could also reduce this down to one trip to the database by getting all filtering to a single queryable

Rock.Rest/Controllers/ContentChannelItemsController.Partial.cs
richarddubay: This is a great idea and it shall be done.

richarddubay:
@edmistj @nairdo I think we're close?

nairdo:
@richarddubay, I don't have a charged laptop sitting at the airport but you _might_ be able to reduce that query further using our `.WhereAttributeValue(...) method. Take a look at the code recipes at the end of this section of the 202 Ignition guide: https://www.rockrms.com/Rock/Developer/BookContent/17/17#attributeampattributevalue

richarddubay:
@nairdo So instead of 

```
// Get the list of attribute values that contain an attribute id from the attributeList.
// and that contain a value that matches one of the data view guids that were passed in.
IQueryable<int?> attributeValueEntityIdList = new AttributeValueService( rockContext ).Queryable()
    .Where( av => attributeIdList.Contains( av.AttributeId ) && guidList.Contains( av.Value ) )
    .Select( av => av.EntityId );

// Get a list of content channel items whose ids match one of the entity ids in the attributeValueEntityIdList.
List<ContentChannelItem> contentChannelItemList = new ContentChannelItemService( rockContext ).Queryable()
    .Where( cci => attributeValueEntityIdList.Contains( cci.Id ) ).ToList();
```

I could do something more along the lines of:
```
// I want a list of content channel items whose ids match up to attribute values that represent entity ids
List<ContentChannelItem> contentChannelItemList = new ContentChannelItemService( rockContext )
    .Queryable()
    .WhereAttributeValue( rockContext, av => attributeIdList.Contains( av.AttributeId ) && guidList.Contains( av.Value ) )
    .ToList();
```

Maybe? Is that what you're thinking? I didn't get a lot of time to look at it today but it seems like maybe that is working.

richarddubay:
@nairdo I went ahead and made the change I _think_ you were talking about. Let me know if there are any other changes that I would need to make in order to make this super functional. I appreciate your help!

PR #3590, opened 2019-03-17
Added Prayers By Groups API Endpoint

richarddubay:
## Proposed Changes
At NewSpring, we are creating a prayer module for our native apps. This change adds an API endpoint to the PrayerRequest controller to allow someone to get all of the prayers from group members of groups that a person belongs to.

Parameters:
`personId`: The id of the person whose groups you want to pull.
`groupTypeIds`: A comma separated list of the group types which determine the types of groups to pull prayers from.

## Types of changes

_NOTE: I can't say this has been approved by the core team. But it is a recommendation to add to the core Rock product._

## Checklist

* [x] I have added [required tests ](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests/README.md) that prove my fix is effective or that my feature works
* [x] I have included updated language for the [Rock Documentation ](https://www.rockrms.com/Learn/Documentation) (if appropriate)

## Further comments
Like the last couple of PR's that I've submitted, I'd love for this to kick off a discussion about the right way to pull off something like this. I know Groups are some of the most data intensive entities so I want to make sure that I'm pulling this information as efficiently as possible. I feel like this is pretty good right now, but would love to make it great. 

Also, my intention is that this would be something that would be helpful to core Rock. If there is anything I need to do in order to make this something that could be merged, please let me know. Also, the converse is true. If this is something that you don't see as helpful to the community, I'll be happy to remove this PR.


nairdo:
Hi @richarddubay I think you should be able to reduce that from 4 database calls to 1 or 2.  Would you mind taking a stab at whipping up some performance metrics for your current 4 db call version (on a decently populated test instance) compared with the new endpoint once you've reduced it down?  I think running it somewhere between 500-2000 times should give you a pretty good idea of the impact and difference between a 4 vs 1 db call endpoint.

richarddubay:
Hey @nairdo ... thanks for getting back with me on this! 

Quick question: When you talk about reducing this from 4 database calls to one or two, are you saying that there are other things I can call that I'm missing (functions or other LINQ calls) that do what I'm trying to do? Or are you just talking about linking these multiple calls together by actually attaching them to each other? 

For example, should I be looking for a different function call to get the group members? Or should I be attempting to link together the group members, groups, and then group member calls into one call? 

I'm guessing it's the latter, but I wanted to make sure. I looked for any "get" type functions that I might use to make these calls easier, but I didn't find anything super helpful. Of course, I could just have missed something. 

Thanks!

nairdo:
Hey @briankalwat, yeah I was really just talking about linking the calls together into something like this pseudo code:

```csharp
IQueryable<PrayerRequest> prayerRequestList = prayerRequestService.Queryable()
    .Where( pr =>
        // ( groupMemberPersonAliasList.Contains( ( int ) pr.RequestedByPersonAliasId ) ) &&
        ( pr.IsActive.HasValue && pr.IsActive.Value == true ) &&
        ( pr.IsPublic.HasValue && pr.IsPublic.Value == true ) &&
        ( pr.IsApproved.HasValue && pr.IsApproved == true ) &&
        ( !pr.ExpirationDate.HasValue || pr.ExpirationDate.Value > now )
    )
    .Where( pr => groupTypeIds.Contains( pr.Group.GroupTypeId ) &&
         pr.Group.Members.Any( m => m.PersonId == personId ) )
```

I haven't studied your endpoint code too closely -- so this snippet may not cover other things you want to do to filter down to the final result set.

Also, can you rename this new endpoint to `GetByPersonsGroupAndGroupTypes` -- I think it matches the  intent of the result set better, right?

richarddubay:
@nairdo I added a couple prayer requests to our rock instance and it did not add a group id to the request. I see a field for that in the database, but it's not being filled in. So I see what you're attempting to do in your example above, but if the group id isn't being filled in I don't think I can do this that way. I did, however, get the calls down to 2. One to get all the way down to the person alias ids of the group members and the other to get the prayers for those people. Which I feel pretty good about.

Do you see anything else I can do to make this better?

nairdo:
@richarddubay That's probably right -- I don't recall any current *core* functionality in Rock whereby a group member can enter a prayer request that would be associated with his group.  I was guessing that is something you guys might have written in-house.  _That's also new core functionality I believe we would love to get funded._

In the meantime, if this new endpoint is trying to mimic a 'group concept' by fetching all of the _other_ active group member's prayer requests for any group for which _**they**_ are also an active member of any groups of the given grouptype, then it needs to be be **renamed** more appropriate.

Perhaps the name `GetForGroupMembersOfPersonInGroupTypes` fits our naming convention better and explains what it's returning: the group member's prayer requests.

Then, a few remaining things:

1. It's getting a bit complex so could you consider either filtering out the person's own prayer requests on the client side -- or -- perhaps an optional boolean flag (excludePerson) should be passed to indicate excluding that person's own prayer requests.  This would avoid someone else asking for a second similar endpoint to also include the given person's prayer requests.
2. I think you should probably only get those people (group members) if they are **still active** (GroupMemberStatus) in that group **and** not marked IsArchived, right?
3. Lastly, I think it might perform better if you used our `AsIntegerList()` extension method (see [example of similar situation](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Rest/Controllers/GroupsController.Partial.cs#L618)) instead of ToString() each GroupTypeId in your LINQ ([here](https://github.com/SparkDevNetwork/Rock/pull/3590/files#diff-83442af11c90d55184fc0e4a796585fcR169)).

richarddubay:
@nairdo These changes are great. I didn't even think about "Oh ... maybe the groups or group members I'm looking at aren't even active in that group". Smart.

Anyway, I made most of your suggested changes except for #1 ... on which I have a question. I like the idea of passing in an `excludePerson` flag (and your reasoning behind it), but it seems to me that doing so would introduce another database call (which is something it seems we were attempting to actively avoid). 

So is it possible to handle something like an `excludePerson` flag without adding another database call? If not, do you feel like we'd be okay introducing another call for this case? FWIW, I should be attempting to load test some of these calls soon so maybe I can test both scenarios, but thought I'd get your opinion first. 

Thanks so much for helping us make these API endpoints great. We all really appreciate it.

cabal95:
@richarddubay You should be able to use the `var person = GetPerson()` method to get the currently authenticated person (this will only hit the database if the value is not already cached, which it probably is).

You can then filter them out with something like (coded here, so it may not be 100% accurate):

```
.Where( pr => !pr.RequestedByPersonAliasId.HasValue || pr.RequestedByPersonAlias.PersonId != person.Id )
```

That should include any prayer request that _doesn't_ have a RequestedByPersonAliasId value, or whose PersonId value does not equal the currently logged in person.

In most cases, this should all happen without any additional queries.

richarddubay:
@nairdo @cabal95 I love the idea of not having to pass the person in. I think that's great and made that change. I also added the `excludePerson` boolean as well. I'm going to do some load testing of this and our other API endpoint suggestions really soon so that we can have some numbers. I'm feeling pretty good about where it is _especially_ considering where it has come from.

richarddubay:
So I had to add back the `personId`. When we call the API from our app, we use a API Rest Key which, as you know, actually corresponds to a person in the database. So when we were pulling prayers, it thought that the current person was the Rest API Key person. Which won't work.

richarddubay:
@nairdo @cabal95 Thanks so much for all your help in making this API endpoint so good. I really do appreciate it! Since this is a pretty custom endpoint, I'm going to go ahead and close this PR and attempt to move this to a plugin sometime down the road. Thanks again for all your help!

sparkdevnetwork-service:
➤ grace hansen commented:

Nick Airdo ( https://app.asana.com/0/21779865363458/list ) - This PR was submitted by Richard Dubay from NS. Flagging this for you for appropriate next steps.

sparkdevnetwork-service:
➤ Nick Airdo commented:

Note from Jon: "Don’t pull in these PRs though as we discussed with them these being in a plugin."

sparkdevnetwork-service:
➤ grace hansen commented:

Nick Airdo ( https://app.asana.com/0/21779865363458/list ) - got it. So do these remain open in Github?

PR #3659, opened 2019-04-22
Adding multi-match (plus capture group) support to the Lava RegExMatchValue filter.

cbump:
## Proposed Changes
The new RegexMatchValue Lava filter in v9 is really slick but it doesn't support multiple matches using capture groups.  This PR takes all capture groups, flattens it into a simple list of strings, and then returns that instead of just a string.

## Types of changes

## Checklist

## Further comments
This will handle multiple matches on the same string as well as match groups.  A good example of using both multi-match and match groups is as follows:

```
{% assign haystack = "red car blue car orange car" %}
{% assign matches = haystack | RegExMatchValue:'(\w+)\s+(car)' %}

<pre>
Whole Match: {{matches[0] }}
Group 1: {{matches[1] }}
Group 2: {{matches[2] }}

Whole Match: {{matches[3] }}
Group 1: {{matches[4] }}
Group 2: {{matches[5] }}

Whole Match: {{matches[6] }}
Group 1: {{matches[7] }}
Group 2: {{matches[8] }}
</pre>
```

This outputs the follow:

> Whole Match: red car
> Group 1: red
> Group 2: car
> 
> Whole Match: blue car
> Group 1: blue
> Group 2: car
> 
> Whole Match: orange car
> Group 1: orange
> Group 2: car

## Documentation
The documentation for the RegexMatchValue lava filter will need to be updated here:

https://community.rockrms.com/lava/filters/text-filters#regexmatchvalue

### DIFF
--- a/Rock/Lava/RockFilters.cs
+++ b/Rock/Lava/RockFilters.cs
@@ -794,22 +794,35 @@ namespace Rock.Lava
         }
 
         /// <summary>
-        /// Returns matched RegEx string from inputted string
+        /// Returns matched RegEx string (or list of strings) from inputted string
         /// </summary>
         /// <param name="input">The input.</param>
         /// <param name="expression">The regex expression.</param>
         /// <returns></returns>
-        public static string RegExMatchValue( string input, string expression )
+        public static object RegExMatchValue( string input, string expression )
         {
-            if ( input == null )
+            if (input == null)
             {
                 return null;
             }
 
             Regex regex = new Regex( expression );
-            Match match = regex.Match( input );
+            var matches = regex.Matches( input );
 
-            return match.Success ? match.Value : null;
+            // Flatten all matches to single list
+            var captured = matches
+                .Cast<Match>()
+                .Where(m => m.Success == true)
+                .SelectMany( o =>
+                    o.Groups.Cast<Capture>()
+                        .Select( c => c.Value ) );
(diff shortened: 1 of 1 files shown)


jonedmiston:
This is a change in behavior (returns a array vs a string) but since it hasn't been deployed we could be ok with it. I wonder though if we should keep this filter `RegExMatchValue` as is (returns a single string) and add an additional filter `RegExMatchGroups` to return an array. If you were only expecting a single value it would be nice to not have to do array length checks etc.

@TimBarley, you PR'ed the original filter. Any input?

TimBarley:
I agree with Jon's assessment. I think both have excellent use cases and I could easily see both being included. If the idea would be to keep the filters at a minimum (to remove any redundant confusion), I would vote for including this code over mine since it will still provide the same feature, just a different way of handling the result.

Thanks for including me in the conversation.

cbump:
@jonedmiston Do you mind weighing in on @TimBarley's comment and just making the final decision.  Whatever you decide is fine by me and I'll update the PR accordingly.

jonedmiston:
@cbump We should move down the path of two separate filters. I believe that will provide to most power. Can you also provide the text needed documentation for the new filter?

jonedmiston:
@cbump , going to close this, but we'd love to see a new PR with the separate filter as discussed above.

Issue #3502, opened 2019-01-17
DDay.iCal schedule recurrence does not override the effective end date


dcs619:
### Description
The current DDay.iCal package appears to be setting DTEnd ~incorrectly~ [to the first instance of the recurrence], ~which~ however [Rock ](https://github.com/SparkDevNetwork/Rock/blob/develop/RockWeb/Blocks/Event/CalendarLava.ascx.cs#L402) then sets to the Schedule.EffectiveEndDate as [seen here ](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Model/Schedule.cs#L390) even though it seems that property should hold the "[date that this Schedule expires and becomes inactive. ](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock/Model/Schedule.cs#L116)".  This code could be "fixed" to include a check for `iCal.Until` instead of just `iCal.DTEnd`. However, with this library also having issues with setting StartDate (#1147) ~and EffectiveStartDate (#1350)~, it may be time to revisit #2050.

If the resulting conclusion is `Won't Fix`, is there a suggested Lava alternative to getting the true Effective End Date?

### Steps to Reproduce
1. Create new schedule on Rock demo
1. Use schedule builder to create a recurring schedule
1. Confirm schedule is set 
1. Confirm via SQL that DTEnd is set to the same as DTStart
1. Output upcoming schedules via Lava and some schedules do not appear 

![screen shot 2019-01-17 at 12 52 21 pm](https://user-images.githubusercontent.com/1210933/51338954-3085d700-1a59-11e9-9b4c-0aba4346a682.png)

![screen shot 2019-01-17 at 12 52 30 pm](https://user-images.githubusercontent.com/1210933/51338958-34195e00-1a59-11e9-8bdb-51dce250cfa1.png)

![screen shot 2019-01-17 at 12 53 07 pm](https://user-images.githubusercontent.com/1210933/51338963-38457b80-1a59-11e9-85ee-f9f013b07f3d.png)

**Expected behavior:**

Schedule.EffectiveEndDate would be set to the End Date of the recurrence.

**Actual behavior:**

Schedule.EffectiveEndDate is set to the Start Date of the recurrence.

### Versions
* **Rock Version:** 8.5
* **Client Culture Setting:** en-us

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1116730867919351">image.png</a>

nairdo:
[...]

PR #3833, opened 2019-07-31
iCal event feed now properly includes all recurring events that overlap with the requested date range.

chrismcgrath:
## Proposed Changes

The GetEventCalendarFeed class was not returning many recurring events that should have been included in the iCal feed, because it was attempting to query for events where the EffectiveStartDate and EffectiveEndDate range overlaps a requested date range, but EffectiveEndDate was populated with the date of the first occurrence.

This change will set EffectiveEndDate to be the end date of the final occurrence, or to 12/31/9999 if the recurrence has no end.

Fixes: #3502 

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality, which has been approved by the core team)

## Checklist

- [X] This is a single-commit PR. (If not, please squash your commit and re-submit it.)
- [X] I have read the [Contributing to Rock](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc

## Further comments

Some other small changes and bug fixes in this PR:

- Fixed a bug that would crash the request if it found an event with no schedule.
- Fixed a bug that would crash if the URL was not properly formatted
- Improved the HTML parsing of Paragraph tags and &nbsp; codes in the description
- Increased the default date range when none is specified from the next two months to 3 months before and 12 months after the current date because the old limit was too strict to be useful. Ideally this would be set with global attributes rather than be hardcoded but that is beyond the scope of this PR.

## Migrations

@jonedmiston said to include the Migration code here. All this does is load all the recurring events and save them again to update their EffectiveEndDate values. As far as I know, there is no need to update the values of non-recurring events.

        public override void Up()
        {
            var rockContext = new RockContext();
            var recurring = new ScheduleService( rockContext ).Queryable().Where( s => s.IsActive ).Where( s => s.iCalendarContent.Contains( "RRULE" ) || s.iCalendarContent.Contains( "RDATE" ) ).ToList();
            foreach ( var schedule in recurring )
            {
                schedule.PreSaveChanges( rockContext, EntityState.Modified );
            }

            rockContext.SaveChanges();
        }

### DIFF
--- a/RockWeb/App_Code/GetEventCalendarFeed.cs
+++ b/RockWeb/App_Code/GetEventCalendarFeed.cs
@@ -23,12 +23,14 @@ using System.Net;
 using Rock;
 using Rock.Data;
 using Rock.Model;
+using Rock.Web.Cache;
 
 using DDay.iCal;
 using DDay.iCal.Serialization.iCalendar;
 
+using RestSharp.Extensions;
+
 using System.Globalization;
-using Rock.Web.Cache;
 
 namespace RockWeb
 {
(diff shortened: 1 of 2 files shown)


RockWeb/App_Code/GetEventCalendarFeed.cs
```diff
@@ -115,70 +117,87 @@ private iCalendar CreateICalendar( CalendarProps calendarProps )
             {
                 foreach ( EventItemOccurrence occurrence in eventItem.EventItemOccurrences )
                 {
-                    iCalendarSerializer serializer = new iCalendarSerializer();
-                    iCalendarCollection ical = ( iCalendarCollection ) serializer.Deserialize( occurrence.Schedule.iCalendarContent.ToStreamReader() );
-
-                    foreach ( var icalEvent in ical[0].Events )
+                    if ( occurrence.Schedule != null )
```
nairdo: @chrismcgrath  As it currently stands, I'm having a hard time seeing what else was changed in this method.  I wonder if this was written as an 'if continue' I think it would be much easier to read the diff:

```
   if ( occurrence.Schedule == null )
   {
      continue;
   }
```

RockWeb/App_Code/GetEventCalendarFeed.cs
```diff
@@ -115,70 +117,87 @@ private iCalendar CreateICalendar( CalendarProps calendarProps )
...
+                            if ( interactionDeviceType != "Outlook" )
+                            {
+                                ievent.Description = description.ConvertBrToCrLf()
+                                                                    .Replace( "</P>", "" )
+                                                                    .Replace( "</p>", "" )
+                                                                    .Replace( "<P>", Environment.NewLine )
+                                                                    .Replace( "<p>", Environment.NewLine )
+                                                                    .Replace( "&nbsp;", " " )
+                                                                    .SanitizeHtml();
+                            }
```
chrismcgrath: The description formatting changed to remove HTML code that was showing up when viewing the feed in a Google calendar.

RockWeb/App_Code/GetEventCalendarFeed.cs
```diff
@@ -115,70 +117,87 @@ private iCalendar CreateICalendar( CalendarProps calendarProps )
...
+                                Uri result;
+                                if ( Uri.TryCreate( eventItem.DetailsUrl, UriKind.Absolute, out result ) )
+                                {
+                                    ievent.Url = result;
+                                }
+                                else if ( Uri.TryCreate( "http://" + eventItem.DetailsUrl, UriKind.Absolute, out result ) )
+                                {
+                                    ievent.Url = result;
+                                }
+                            }
```
chrismcgrath: The above code checks for a valid URL, attempts to fix the URL if it isn't valid, and ignores an unfixable invalid URL. Previously a url like www.rockrms.com would crash the feed.

RockWeb/App_Code/GetEventCalendarFeed.cs
chrismcgrath: Agreed, unfortunate that it displayed the diff this way.  I commented the other changes below.

RockWeb/App_Code/GetEventCalendarFeed.cs
nairdo: @chrismcgrath Ok.  My only last concern is the number of schedules that would be updated in that loop.  I've checked a few large systems and have seen between 2287 and 16380  schedules (`SELECT COUNT(1) FROM [Schedule] WHERE [IsActive] = 1`).  Can that Queryable in the migration Up() be tightened/target any further (to only the necessary schedules)?

nairdo:
For reference, here is the best way to view the **diff** between the before and after (including the final migration): https://github.com/SparkDevNetwork/Rock/compare/006b1a50be18ce76ddcf9c56ff52b611779bba52...359e07afa5563d641368af1ac64e0a6eb0294de3

nairdo:
@chrismcgrath  Just a quick note -- the changes from this PR seem to have severely impacted the performance of the Event Calendar, so we've been working various fixes to try to address each uncovered issue (https://github.com/SparkDevNetwork/Rock/commit/a86b3c54c10e5fce6a1009bdfd6e73eb2dead06c#diff-bc523c91b13cfbde2de37bff923fa41e and https://github.com/SparkDevNetwork/Rock/commit/95628e1076338575dc52685eed825a94253dd35e;  We haven't proven this last commit in a real, data heavy environment yet).

PR #3842, opened 2019-08-06
Update Email Preference Entry

engineeringdrew:
## Proposed Changes

We wanted an ability to launch a workflow whenever a person unsubscribes from a list.  We wanted to be able to alert the staff and take the appropriate actions automatically whenever a person unsubscribes.  This PR will add a workflow type field to the block settings that isn't required.  After the communication lists to unsubscribe from are chosen and the submit button is clicked on, the workflow type if chosen will be used to get the workflow and launch it.

**Here's what I did:**
I updated the EmailPreferenceEntry.ascx.cs code to include a workflow type field block attribute and used its guid in the click event of the submit button only if the unsubscribe radio button is checked.  The guid is used to get the workflow type cache that is passed to a method I called StartWorkflow that launches the workflow associated with the workflow type cache.  The method StartWorkflow uses a workflow service, the workflow type cache, a dictionary of attributes, and a workflow name suffix.

This worked, and here's a snippet of some of the C# code I added:

                Guid? workflowGuid = GetAttributeValue( "UnsubscribeWorkflow" ).AsGuidOrNull();
                WorkflowTypeCache workflowType = null;
                var workflowService = new WorkflowService( rockContext );

                if (workflowGuid != null)
                {
                    workflowType = WorkflowTypeCache.Get( workflowGuid.Value );
                }

                // Start workflow for this person
                if (workflowType != null)
                {
                    Dictionary<string, string> attributes = new Dictionary<string, string>();
                    StartWorkflow( workflowService, workflowType, attributes, string.Format( "{0}", _person.FullName ), _person );
                }
## Types of changes

New feature (non-breaking change which adds functionality, which has been approved by the core team)

## Checklist

- [x] I have read the Contributing to Rock doc
- [x] By contributing code, I agree to license my contribution under the Rock Community License Agreement

## Further Comments

Tested to make sure it launches a workflow without messing up when a workflow type is chosen.

-----
Assigned to JacobM for QA Testing

### DIFF
--- a/RockWeb/Blocks/Communication/EmailPreferenceEntry.ascx.cs
+++ b/RockWeb/Blocks/Communication/EmailPreferenceEntry.ascx.cs
@@ -45,6 +45,7 @@ namespace RockWeb.Blocks.Communication
     [MemoField( "No Mass Emails Text", "Text to display for the 'No Mass Emails' option.", false, "I am still involved with {{ 'Global' | Attribute:'OrganizationName' }}, but do not wish to receive mass emails (personal emails are fine).", "", 3, null, 3, true )]
     [MemoField( "No Emails Text", "Text to display for the 'No Emails' option.", false, "I am still involved with {{ 'Global' | Attribute:'OrganizationName' }}, but do not want to receive emails of ANY kind.", "", 4, null, 3, true )]
     [MemoField( "Not Involved Text", "Text to display for the 'Not Involved' option.", false, " I am no longer involved with {{ 'Global' | Attribute:'OrganizationName' }}.", "", 5, null, 3, true )]
+    [WorkflowTypeField( "Unsubscribe Workflow", "The workflow type to launch for person who wants to unsubscribe.", false, required: false )]
     [MemoField( "Success Text", "Text to display after user submits selection.", false, "<h4>Thank You</h4>We have saved your email preference.", "", 6, null, 3, true )]
     [CodeEditorField( "Unsubscribe Success Text", "Text to display after user unsubscribes from communication lists.", CodeEditorMode.Lava, CodeEditorTheme.Rock, 200, false, UNSUBSCRIBE_SUCCESS_TEXT_DEFAULT_VALUE, order: 7 )]
     [TextField( "Reasons to Exclude", "A delimited list of the Inactive Reasons to exclude from Reason list", false, "No Activity,Deceased", "", 8 )]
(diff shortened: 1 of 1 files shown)


jonedmiston:
Hey @engineeringdrew could you add a little background about this PR and maybe which organization it's for.

RockWeb/Blocks/Communication/EmailPreferenceEntry.ascx.cs
```diff
@@ -227,11 +228,27 @@ protected override void OnLoad( EventArgs e )
...
+
+                if (workflowGuid != null)
+                {
+                    workflowType = WorkflowTypeCache.Get( workflowGuid.Value );
+                }
+
+                // Start workflow for this person
+                if (workflowType != null)
+                {
+                    Dictionary<string, string> attributes = new Dictionary<string, string>();
```
nairdo: Hi @engineeringdrew I see this `attributes` list you pass to the `StartWorkflow` method _would_ be used by setting them into the launched workflow (`workflow.SetAttributeValue(...)`) but I don't see you populating this list anywhere.  Did you miss this? (Or am I missing something?)

engineeringdrew: I just left that in there in case any developer wanted to use it.  Right now it doesn't do anything.

jwmoses8995:
@nairdo  I tested this and it works!

RockWeb/Blocks/Communication/EmailPreferenceEntry.ascx.cs
nairdo: @engineeringdrew I guess what I should have said is -- you're passing in a completely _empty_ dictionary.  So the `foreach` loop in the `StartWorkflow` method will never do anything.  It's just extra code and should be removed -- because otherwise, glancing at the `StartWorkflow` method, a developer might not even realize there are _no_ attributes in the dictionary.

On the other hand, if you know there's something in the block's context-data that should be passed in (for an example, see https://github.com/SparkDevNetwork/Rock/commit/ae912c29e44ef7c9df0428a4b01ad3f8b037c5d2#diff-589989fd7f3358571965f779a866aaf3R381) then it could make sense to add them to the dictionary and leave the code in the StartWorkflow method as is.

Let me know what you think about this.

Thanks for looking at this again -- and for helping make Rock better!

nairdo:
@engineeringdrew I'll pull this in and make those minor adjustments.  Thanks for your help with making Rock better!

nairdo:
@engineeringdrew  I forgot to reference this PR when I made one more change after the one marked above so I thought I'd reference it here: https://github.com/SparkDevNetwork/Rock/commit/4e6d891505f45b14e8e43e01b65671790ae6d2ef

PR #3966, opened 2019-10-24
Group Schedule Toolbox Added Ability

TheArkChurch:
## Proposed Changes

Added the ability to have Schedule names displayed to the user via a block setting.

In most of the group scheduling process you deal with schedules by name this is also how we deal with them internally. A person rarely sees the actual time a schedule starts instead they see Sunday 1st, Sunday 2nd, Sunday 3rd etc... All of these schedules start at 5:30am, for us and for multiple reasons, which might be unique to us. This is the only place I have found where it's an actual time displayed and not the schedule name, and for us it's causing issues. 

This is Not a Bug, but when a volunteer looks at their schedule, after they've confirmed, knowing that they've accepted for Sunday 3rd service and then they only see 5:30am they freak out. No where does it show the schedule name. Now you can change the behavior in the block settings.

The top image shows the before and the bottom shows the after (if the block is configured with the new setting enabled):
Times
Names

## Types of changes

- [ x] New feature (non-breaking change which adds functionality, ~~which has been approved by the core team)~~

## Checklist

- [ x] This is a single-commit PR. (If not, please squash your commit and re-submit it.)
- [ x] I verified my PR does not include whitespace/formatting changes -- because if it does it will be closed without merging.	
- [ x] I have read the [Contributing to Rock](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc
- [ x] By contributing code, I agree to license my contribution under the [Rock Community License Agreement](https://www.rockrms.com/license)

## Migrations

### DIFF
--- a/RockWeb/Blocks/GroupScheduling/GroupScheduleToolbox.ascx.cs
+++ b/RockWeb/Blocks/GroupScheduling/GroupScheduleToolbox.ascx.cs
@@ -80,6 +80,13 @@ namespace RockWeb.Blocks.GroupScheduling
 </div>",
         Order = 2
          )]
+         
+    [BooleanField(
+        "Show Names on schedules instead of Times?",
+        Key = AttributeKey.ShowScheduleNames,
+        Description = "Set to true to show the Names's.",
+        DefaultBooleanValue = false,
+        Order = 3 )]
     public partial class GroupScheduleToolbox : RockBlock
     {
         protected class AttributeKey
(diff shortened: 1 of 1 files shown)


TheArkChurch:
Added images to show how this looks on our production system. We let volunteers check-in for all 3 services at the same time if they need to.

nairdo:
@TheArkChurch I accidentally referenced the wrong (older) PR #3965 when I made these recent commits (https://github.com/sparkdevnetwork/rock/commit/bfcdb6ecfc9a7d6b094608b1e2b7112357739713 https://github.com/sparkdevnetwork/rock/commit/40578f891ed92a89837faf4d68093c183ef67370) after we decided to display the schedule's "name" the _default_ behavior.

Therefore this PR is also good to close now.  Thanks again for pointing this issue out.

TheArkChurch:
Glad to hear it. There is one other block, the Group Schedule Confirmation block, that has the same behavior and should be changed. It wasn't in this pr because I didn't find it until later and hadn't had the time to update it.

PR #4096, opened 2020-02-05
+ Added new EntityFromCacheObject Lava filter. (Fixes #4089)

### LINKED ISSUE #4089: [Proposal] New Lava Filter EntityFromCachedObject
## Proposal

This proposal is for a new Lava filter that would allow the user to turn a Cached object into a real Entity object.

In v10, a bug was fixed that caused the `Attribute` filter to not return cached objects. While this is a good thing, it does make some things harder to do in Lava. Suppose we have this Lava:

```
{% assign Campus = Workflow | Attribute:'Campus','Object' %}
{{ Campus.Location.FormattedAddress }}
```

On v10 it will not work because the campus object returned is actually a cached object that only has a bare-bones `Location` property. Currently the way around this would be to use Entity commands to load the campus and then traverse it's properties, but that is a rather unfriendly way to do things.

The proposal would be that this new Lava filter would take in an object and if it is an `IEntityCache` object then it would use a little magical Reflection to determine the actual entity type and then load and return that entity from the database.  If the incoming object is not `IEntityCache` then the original object would be returned. In other words, if you pass in a real `IEntity` object then you will get back a real `IEntity object still. But if you pass in an `IEntityCache` object, you will get a real `IEntity` object back.

```
{% assign Campus = Workflow | Attribute:'Campus','Object' | EntityFromCachedObject %}
{{ Campus.Location.FormattedAddress }}
```

## Pull Request

If this proposal is approved, I can build up and submit a pull request for this functionality. Just want to verify the concept before I start coding something.

## Input Needed

I don't really care for the filter name, but haven't come up with a better one yet.

cabal95:
## Proposed Changes
A recent change to Rock had a side affect of causing some issues when accessing objects via Lava. If you get the object value of certain attributes, they now return a cached object instead of an entity object. Normally this is fine, but there are a few cases where you need the real entity object. This PR adds a filter that allows the user to easily convert from a cached object to an entity object.

Fixes: #4089 

## Types of changes

## Checklist

* [x] This is a single-commit PR. (If not, please squash your commit and re-submit it.)

## Further comments
New Extension method and Unit Tests were added.

## Documentation
Documentation for new filter has been added.

https://community.rockrms.com/lava/filters/other-filters#entityfromcachedobject

## Migrations
If your pull request requires a migration, please _exclude the migration from the Rock.Migration project_, but submit it with your pull request. Please add a note to your pull request that provides a heads up that a migration file is present.


nairdo:
@cabal95 Major bonus points for already adding the Lava documentation!

PR #4334, opened 2020-08-06
+ Fixes incorrect sorting on Attribute Values in LinqExtensions (Fixes #4332)

### LINKED ISSUE #4332: Registrant List Not Sorting on Person Attribute
### Prerequisites
* [x] Put an X between the brackets on this line if you have done all of the following:
    * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    * Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

When you have a Person Attribute that is shown in the Registrant Grid and you try to sort by that field the sorting is incorrect. I have spent some time debugging this and it looks like the issue is in Rock.ExtensionMethods.LinqExtensions.cs in the Sort function (Line 386). What is happening is that when it pulls the model to use for the custom sort value it is matching on the attributeValue.EntityId being equal to the model.Id, however the model is a registrant entity so matching on those fields is pulling the incorrect entity, the match when the attribute is a person attribute should actually be on attributeValue.EntityId = model.Person.Id. It seems like the attributeCache.EntityTypeId needs to be checked before the value for model is set. 

### Steps to Reproduce

1. Create a new Registration Template where the Registrant Form has a Person Attribute that will display in the grid
2. Create a Registration Instance
3. Populate the Registration Instance
4. Attempt to sort on the Person Attribute in the Registrant Grid

**Expected behavior:**

The values display in the desired sorted order.

**Actual behavior:**

The values are incorrectly sorted.

### Versions

* **Rock Version:** Rock McKinley 10.2 (1.10.2.3)

courtneycooksey:
## Notice

## Proposed Changes

Solution for #4332 where the attributeValue EntityId was trying to match on the wrong entity type. I added a check to see if the source entity and attribute entity type were the same, if they are not I search the source entity for property of the same type as the attribute entity type and same the property information as well as the property's primary key field. When matching the attributeValue.EntityId I use the correct property and primary key field to match against. 

Fixes: #

## Types of changes

## Checklist

## Migrations

### DIFF
--- a/Rock/Utility/ExtensionMethods/LinqExtensions.cs
+++ b/Rock/Utility/ExtensionMethods/LinqExtensions.cs
@@ -16,7 +16,6 @@
 //
 using System;
 using System.Collections.Generic;
-using System.ComponentModel.DataAnnotations;
 using System.Data.Entity;
 using System.Linq;
 using System.Linq.Expressions;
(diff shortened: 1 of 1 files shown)


MikeDPeterson:
@courtneycooksey 
**Feedback Comments**

* This is pretty cool! So it looks like it works like this
  * When sorting an IQueryable of RegistrationRegistrant (for example) by one of the attribute values, it detects if the Attribute is really on some other entityType. For example, it could be an attribute on RegistrationRegistrant.GroupMember or RegistrationRegistrant.Person  instead of just Registration Registrant
  * If  the attribute's entityType is really Person (for example),  it finds the Matching Property on RegistrationRegistration, which would be RegistrationRegistrant.Person, then sorts by that instead
  * Great job of figuring out how the guts of Event Registration, EntityType, Attributes, EntityId and all that works! 
  * There was some tricky stuff in there. _Pretty impressive_.
  * The comments do a good job of explaining each section

**Other Comments**
* This is fairly specific to RegistrationInstanceRegistrantList, and it would have been tempting to just deal with the issue there, but a more general approach was taken -- which is nice. (as a result it should be tested thoroughly)

**Suggestions**
* EntityTypeCache could be used here instead https://github.com/SparkDevNetwork/Rock/pull/4334/files?w=1#diff-153ecfca12d309cf3b61ee13991a964dR377
* These sections could be simplified since the PK on any IEntity/IModel/IHasAttributes is [Id]. 
https://github.com/SparkDevNetwork/Rock/pull/4334/files?w=1#diff-153ecfca12d309cf3b61ee13991a964dR397
https://github.com/SparkDevNetwork/Rock/pull/4334/files?w=1#diff-153ecfca12d309cf3b61ee13991a964dR428

**Issues**
* An null exception will occur if there are no items in the grid https://github.com/SparkDevNetwork/Rock/pull/4334/files?w=1#diff-153ecfca12d309cf3b61ee13991a964dR376

**Minor stuff**
  * Whitespace -- doesn't match [our normal white-spacing](https://github.com/SparkDevNetwork/Rock/wiki/Coding-standards)


@courtneycooksey  If you are willing to make those corrections to this PR, please make an updated commit or re-submit a clean commit. Great Job!

courtneycooksey:
@MikeDPeterson I believe I have made all the appropriate changes, please let me know if I missed something.

Issue #3893, opened 2019-09-11
API Docs Page hardcodes port number in api requests


jeffmikels:
### Description

In the file served from `domain/api/docs/index.html` the port number is incorrectly generated for the backend requests. This is only a problem when running Rock behind a proxy. The request to the proxy will need to be on one port, but the request to the actual Rock server will be on another port.

For example, if the browser accesses the site over https, but the site proxies to Rock over regular http, Rock will generate a url for the api resources that honors the specified protocol (such as https) but will specify the port as 80.

As a result, the generated JavaScript will look like this (line 169 of the generated html):

```javascript
$.get(': https://: domain.tld:80' + '/api/RestControllers/EnsureRestControllers', function () ...
```

### Steps to Reproduce

1. Install Rock
2. Put Rock behind a reverse proxy handling https but forwarding requests over http
3. Visit https://YOURROCKSERVER/api/docs/
4. Resources will not load

**Expected behavior:**

Rock should not specify the port number in those urls. Or Rock should allow the user to customize the protocol and port settings it uses for generated urls.

**Actual behavior:**

Urls are generated inconsistently with https as the protocol but 80 as the port number.

### Versions

* **Rock Version:** [8.8]
* **Client Culture Setting:** [Unknown]

jeffmikels:
It's still not fixed in 10.x. I think this should be an easy edit.

jeffmikels:
It appears this can be fixed with a change to `SwaggerConfig.cs` to redefine the RootUrl... and/or a change to `RockRequestContext.cs` I'm issuing a pull request to see if it would help.

PR #4336, opened 2020-08-07
Fix for #3893

jeffmikels:
closing #3893 

Swagger RootUrl seems to pick up the scheme from the request url, but then generates a port based on the internal (behind the proxy) url resulting in urls like this:

`https://rockserver.example.com:80`

### DIFF
--- a/Rock.Rest/Swagger/SwaggerConfig.cs
+++ b/Rock.Rest/Swagger/SwaggerConfig.cs
@@ -74,7 +74,18 @@ namespace Rock.Rest.Swagger
                          // However, there may be situations (e.g. proxy and load-balanced environments) where this does not
                          // resolve correctly. You can workaround this by providing your own code to determine the root URL.
                          //
-                         //c.RootUrl(req =>
+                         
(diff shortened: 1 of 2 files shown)


cabal95:
Was the change to RockRequestContext intentional? It's only used by mobile blocks, and the change doesn't seem to make sense the way it's written since the variable  is being overwritten by the next line of code anyway.

nairdo:
Hi @jeffmikels  I'd have to concur with Daniel's question/statement.  I don't think the change to RockRequestContext was needed/valid, right? https://github.com/SparkDevNetwork/Rock/pull/4336/files#diff-ab15647cbb77ee652aa00ede7e261f03R129

nairdo:
Closing until Jeff's had a chance to look this over and possibly resubmit.

treusch:
@jeffmikels @nairdo - was this ever fully fixed. We are getting an error trying to access the api docs that seems to be this issue. Thanks!

jeffmikels:
I gave up on Rock partially because of this. That was years ago. The church is using Planning Center now.

PR #4357, opened 2020-08-20
Added ContentItemById and ContentItemByGuid lava filters

briankalwat:
## Proposed Changes

This PR adds two new lava filters for use around Content Channel Items. The `ContentItemByGuid` filter is especially useful when used in conjunction with the Content Channel Item attribute, as that attribute does not currently support returning the entire content item object.

The `ContentItemById` filter can be helpful when wanting to reference or use data from a specific content item within the static content of a page, without having to write out or use the full entity command.

Ex: 
`{{ item | Attribute:'ContentChannelItem','Object' }}`
does not work (currently).

So instead...
`{{ item | Attribute:'ContentChannelItem','RawValue' | ContentItemByGuid }}`
or

```
{% assign item = 1234 | ContentItemById %}
Check out this story from {{ item | Attribute:'Author' }}, called "{{ item.Title }}".
```

## Types of changes

## Checklist

## Documentation
Assuming we could add a "Content Channel Item" section under filters in the lava docs, or just list these under "Other".

**ContentItemById**
Returns a full content channel item object from the Id of the content channel item.
Example: 
```
{% assign item = 1234 | ContentItemById %}
Check out this story from {{ item | Attribute:'Author' }}, called "{{ item.Title }}".
```

**ContentItemByGuid**
Returns a full content channel item object from the Guid of the content channel item.
Example: 
```
{% assign relatedItem = item | Attribute:'RelatedContentChannelItem','Guid' | ContentItemByGuid %}
Check out this related story, called "{{ relatedItem.Title }}".
```

### DIFF
--- a/Rock/Lava/RockFilters.cs
+++ b/Rock/Lava/RockFilters.cs
@@ -2441,6 +2441,58 @@ namespace Rock.Lava
                 PersonService.DeleteUserPreference( person, settingKey );
             }
         }
+        
+        /// <summary>
+        /// Content Channel Item the by identifier.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="input">The input.</param>
+        /// <returns></returns>
+        public static object ContentItemById( DotLiquid.Context context, object input )
+        {
+            if ( input == null )
+            {
+                return null;
+            }
+
+            int contentChannelItemId = -1;
+
+            if ( !Int32.TryParse( input.ToString(), out contentChannelItemId ) )
+            {
+                return null;
+            }
+
+            var rockContext = new RockContext();
+
+            return new ContentChannelItemService( rockContext ).Get( contentChannelItemId );
+        }
+
+        /// <summary>
+        /// Content Channel Item the by unique identifier.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="input">The input.</param>
+        /// <returns></returns>
+        public static object ContentItemByGuid( DotLiquid.Context context, object input )
+        {
+            if ( input == null )
+            {
+                return null;
+            }
+
+            Guid? contentChannelItemGuid = input.ToString().AsGuidOrNull();
+
+            if ( contentChannelItemGuid.HasValue )
+            {
+                var rockContext = new RockContext();
+
+                return new ContentChannelItemService( rockContext ).Get( contentChannelItemGuid.Value );
+            }
+            else
+            {
+                return null;
+            }
+        }
 
         /// <summary>
         /// Persons the by identifier.


jonedmiston:
@briankalwat thanks for this. We can see the value in this for sure. Nick and I were talking today and wanted to run an idea past you. This could turn into a slippery slope where the community wants more and more of these (like the person ones). What if we made the filter more generic like this:

`{% assign item = '1354812B-6B61-3CB5-42CA-6B25157D318D'| EntityByGuid:'ContentChannelItem' %}`
and
`{% assign item = 12| EntityById:'ContentChannelItem' %}`

The filter takes a parameter that tells it what Entity Type to send back. Wanted to get your input. We can code it.

cabal95:
Would the “ById” version be worrisome? I think that kind of replaces entity commands which can be disabled. But this filter couldn’t be disabled so you are giving somebody a lot of access they Potentially shouldn’t have just because they can type some lava.

briankalwat:
@jonedmiston even better! I think it accomplishes the same goal and makes it a whole lot more useful at the same time. 

@cabal95 i'm not sure I follow how this is any less secure than the entity command since entity has the `securityenabled` parameter which can be used to bypass. Are you saying because the entity command technically has to be an enabled command for it to work that it's more secure? Curious to hear your thoughts.

cabal95:
@briankalwat yes. So for example, if I create an HTML block and give a staff/user permission to edit the content; I as the administrator can choose to not give them access to not run entity commands. With the “ById” filter, I no longer have that ability to deny them access. The ByGuid filter is less of a concern to me because you would have to know the specific Guid. You couldn’t just loop through numbers looking for valid ids.

Would just fixing the Content Channel Item attribute/field to support the “Object” access mode work? That is an easy fix and I think it’s a bug that it doesn’t in the first place. Then the “ByGuid” generic filter could still be implemented without opening a huge potential security concern (IMHO).

jonedmiston:
Looking into the history of the ContentChannelItem fieldtype it looks like the implementation to allow for 'Object' was missed. We are going to fix that and spin through the other fieldtypes looking for others (seems like more than one was missed).

briankalwat:
@jonedmiston that sounds great - that will be very helpful, thank you!

briankalwat:
@jonedmiston revisiting this again with another use case for an `EntityByGuid` lava filter - it looks like the "Custom Grid Options" tab on blocks don't typically make an entire entity accessible (unless i'm missing something), so the only way to get to attributes is via entity command, using the Row Id that is available.

image

```
{% eventitem where:'Id == {{ Row.Id }}' iterator:'EventItems' %}
{{ EventItems | First | Attribute:'Childcare' }}
{% endeventitem %}
```

With the filter, it'd just be `{{ Row.Guid | EntityByGuid:'EventItem' | Attribute:'Childcare' }}`.

This would be a super powerful addition, but I have no idea what it would take to add it. Thanks for reading and considering - i'm super grateful for all that y'all do!

jonedmiston:
@briankalwat Just want to make sure I understand. How would this be different than an `entity` command? 

On a separate note, you want to be _very_ careful about performance when adding 'Custom Grid Options'. The example above would add at least two database calls per row. This would make the page above very slow if you have more than a 20-30 records on your grid.

jonedmiston:
@briankalwat the custom grid columns are a good way to add links to other pages for reporting, launching workflows, etc. They don't work well if you're trying to load additional data from the database if you have lots of rows in your grid. We put a warning at the top to reflect that (we have an internal task to make the message stronger). 

Like everything in technology the more power the more responsibility. Not sure of the use-case you're trying to solve but you might need to consider a dynamic data block with custom SQL to get the Entity Data and Attribute Data in one call. SQL is my tool of last resort, but it often is the answer when you run into performance issues (that or custom C#).

PR #4362, opened 2020-08-20
Added ability to specify Entity Context URL values by Guid.

cabal95:
## Notice

## Proposed Changes

With security becoming more and more a concern these days, we wanted to stop using Id numbers for our Context URL values in certain places. For example, in our Small Group Locator pages, we want to use an HTML block to display additional information about the group passed in the URL. This is currently passed by Id number as that is what Context uses. But in this case, we don't want the user to be able to start passing in random Group Ids and potentially find sensitive information (since groups are used everywhere).

We could get around this by using entity commands as well, but thought this would be a feature that others would benefit from as well.

This is a bit security by obscurity since the same query string parameter accepts both Guid values as well as integer values. But our intention is to actually change the key name  to imply a Guid to make the user think they can't just easily change the value:

```
http://www.example.org/page/123?GroupGuid=07f063c0-0661-462e-971e-d67cd574d1a9
```

Instead of:

```
http://www.example.org/page/123?GroupId=47
```

To be clear about the above, if we decided to just use `Group` as the parameter name, then both of these would be valid:

```
http://www.example.org/page/123?Group=07f063c0-0661-462e-971e-d67cd574d1a9
http://www.example.org/page/123?Group=47
```

But as I said, in our hope is that the user sees a Guid value and doesn't bother to try changing it to an integer value.

Fixes: #

## Types of changes

- [x] New feature (non-breaking change which adds functionality, ~which has been approved by the core team~)

## Checklist

## Further comments

Security could be improved by checking if the context key name ends in `Guid` and if so don't allow integer values - but that does have the potential of breaking existing pages where they accidentally used a query string parameter name ending in `Guid` even though they are passing integer values.

Another option would be to modify the Context Parameters UI and stored data to allow the user to select if they want to allow Integer Id values, Guid values, or both and then enforce those choices.

## Documentation

https://community.rockrms.com/documentation/bookcontent/14/206#usingcontext should probably be updated to mention that Guid values are allowed in addition to integer values.

## Migrations

### DIFF
--- a/Rock/Data/Entity.cs
+++ b/Rock/Data/Entity.cs
@@ -576,12 +576,21 @@ namespace Rock.Data
     internal class KeyEntity
     {
         public string Key { get; set; }
-        public int Id { get; set; }
+
+        public int? Id { get; set; }
+
+        public Guid? Guid { get; set; }
+
         public IEntity Entity { get; set; }
 
         public KeyEntity( int id )
         {
             Id = id;
+        }
+
+        public KeyEntity( Guid guid )
+        {
+            Guid = guid;
         }
 
         public KeyEntity( string key )
--- a/Rock/Web/UI/RockPage.cs
+++ b/Rock/Web/UI/RockPage.cs
@@ -1037,6 +1037,12 @@ namespace Rock.Web.UI
                                 {
                                     ModelContext.AddOrReplace( modelContextName, new Data.KeyEntity( contextId.Value ) );
                                 }
+
+                                Guid? contextGuid = PageParameter( type.Name + "Guid" ).AsGuidOrNull();
+                                if ( contextGuid.HasValue )
+                                {
+                                    ModelContext.AddOrReplace( modelContextName, new Data.KeyEntity( contextGuid.Value ) );
+                                }
                             }
                         }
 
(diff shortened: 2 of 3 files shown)


briankalwat:
This is great, @cabal95! 🚀

matthewbronkema:
@cabal95, this is cool change! However it introduces an issue into the person profile page. Go to Ted Decker's profile Step tab using his Guid (/Person/8FEDC6EE-8630-41ED-9FC5-C7157FD1EAA4/Steps). When you click Noah the link should go to Noah's step tab, but instead it goes to Noah's Person Profile tab.

nairdo:
So, can you also handle this case so that the 'sticky' tab feature (see [video example](http://somup.com/cYjw2HXiNE)) is maintained with your enhancement?  (It has to do with how the FormatPersonLink in the `RockWeb\Blocks\Crm\PersonDetail\GroupMembers.ascx.cs` is parsed/created.)

cabal95:
Fixed. The FormatPersonLink method now defaults to checking against Person.Id, but if the page has a Person page context parameter (which is always should) then it will whatever value is specified in the URL for it's string replacement test. This should ensure compatibility with future changes that may provide yet other ways to specify the query string parameter.

nairdo:
Thanks @cabal95 Daniel!

PR #4365, opened 2020-08-21
Added support for web blocks to support API actions.

cabal95:
## Notice

## Proposed Changes

This is a developer level feature. When Rock added support for Mobile App blocks, one of the features of those blocks was to allow direct API calls to the blocks. This provided blocks with a way for the mobile app to communicate with the block to load data asynchronously. This was necessary since we didn't have the option for webforms postback. This feature was implemented via an interface.

This PR adds support to the base `RockBlock` to implement this interface. This allows any web-based block to implement API actions that return custom data (for example, an object that would be encoded as JSON and sent to the client). An example usage would be:

```cs
[BlockAction]
public object AddIntegers( int a, int b )
{
    return a + b;
}
```

This action could be called via a URL such as `http://localhost:6229/api/Blocks/Action/12/6141/AddIntegers?a=10&b=15` and would return the value `25`.

You could also send the request as a POST (required to send complex parameters), `http://localhost:6229/api/Blocks/Action/12/6141/AddIntegers` with the body data: `{"a": 10, "b": 15}`

Fixes: #

## Types of changes

## Checklist

## Migrations

### DIFF
--- a/Rock/CheckIn/CheckInBlock.cs
+++ b/Rock/CheckIn/CheckInBlock.cs
@@ -812,6 +812,7 @@ namespace Rock.CheckIn
                             if ( checkinBlock != null )
                             {
                                 checkinBlock.SetBlock( page, block, true, true );
+                                checkinBlock.RequestContext = RequestContext;
                                 checkinBlock.GetState();
                                 return checkinBlock;
                             }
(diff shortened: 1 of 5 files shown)


cabal95:
Closing until further discussion happens.

PR #4496, opened 2020-11-19
Added field type for email address that requires retyping for confirmation

MichaelAllen:
## Notice

## Proposed Changes

This PR adds a new field type `EmailConfirmField` that prompts for an email address and a confirmation. It will only validate if both emails match. It also updates most public facing blocks that were asking for an email address to also ask for a confirmation.

See [this idea](https://community.rockrms.com/ideas/1467/prompt-to-confirm-email-address-on-public-facing-blocks) for more explanation of the issue this is trying to solve.

>We are seeing a constant (now magnified since people are interacting with us more online) issue where people enter their email address wrong, causing all sorts of data integrity issues.
>
>The core issue is that people often do not enter their email address correctly (especially on mobile) and while mis-spelling a name or phone might not be ideal, misspelling an *email* address means we cannot contact them at all (especially if we're not getting a phone number, too). For years the industry standard to work around this has been to solicit the email address TWICE, and only let the form submit if they match. This gives the user a chance to fix a mistake they didn't realize they made.

*Note: If this isn't something that we want on all the public blocks, I can resubmit with just the new field type. I think it still would be valuable to have this as an option for workflows.*

Fixes: #N/A

## Types of changes

- [x] New feature (non-breaking change which adds functionality, ~which has been approved by the core team~)

## Checklist

## Further comments

I went through and tested every block that is modified by this PR and they all correctly reject bad emails, and accept matching emails. The only block I was not able to test is `Transaction Entry v2` since it doesn't work with the Test Gatweay. However, I don't see a reason it should have an issue based on reading through the code.

## Documentation
N/A

## Migrations
N/A

### DIFF
--- a/Rock/Field/Types/EmailConfirmFieldType.cs
+++ b/Rock/Field/Types/EmailConfirmFieldType.cs
@@ -1,0 +1,236 @@
+﻿// <copyright>
+// Copyright by the Spark Development Network
+//
+// Licensed under the Rock Community License (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.rockrms.com/license
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+//
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using System.Web.UI;
+
+using Rock.Reporting;
+using Rock.Web.UI.Controls;
+
+namespace Rock.Field.Types
+{
+    /// <summary>
+    /// Field used to save and display an email address with required confirmation
+    /// </summary>
+    [Serializable]
+    public class EmailConfirmFieldType : FieldType
+    {
+        #region Formatting
+
+        /// <summary>
+        /// Formats the value as HTML.
+        /// </summary>
+        /// <param name="parentControl">The parent control.</param>
+        /// <param name="value">The value.</param>
+        /// <param name="configurationValues">The configuration values.</param>
+        /// <param name="condensed">if set to <c>true</c> [condensed].</param>
+        /// <returns></returns>
+        public override string FormatValueAsHtml( Control parentControl, string value, Dictionary<string, ConfigurationValue> configurationValues, bool condensed = false )
+        {
+            return HtmlFormat( FormatValue( parentControl, value, configurationValues, condensed ) );
+        }
+
+        /// <summary>
+        /// Formats the value as HTML.
+        /// </summary>
+        /// <param name="parentControl">The parent control.</param>
+        /// <param name="entityTypeId">The entity type identifier.</param>
+        /// <param name="entityId">The entity identifier.</param>
+        /// <param name="value">The value.</param>
+        /// <param name="configurationValues">The configuration values.</param>
+        /// <param name="condensed">if set to <c>true</c> [condensed].</param>
+        /// <returns></returns>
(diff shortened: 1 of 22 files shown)


jonedmiston:
@MichaelAllen I wish we'd have talked before you put all of the effort into this. We can't be heavy handed in forcing this feature on the whole community. While many would like to have emails be double entered there will be an equal (or more) that would not want to force this from a UX experience (especially on mobile). 

We have ideas on how this could be done differently though. We would recommend that we keep the current email field and add a System Setting that would 'Enable Email Confirmation'. The field would also have a new property that would opt into confirmation also so it would only apply to public blocks. We'd need to check that it did not compromise the UI (might need some adjustments).

If you'd like to work down that path we should talk to iron out any details.

MichaelAllen:
I am definitely willing to modify this to better fit with your vision for the feature. 

I had originally created this control/field type to solve a problem we were having with people misspelling their email on workflow forms. The additional time/effort to test it on some of the public blocks was minimal.

I like your idea of having a single field with a property for showing the confirmation. I think it will make for a better user experience too, since someone could easily switch an existing attribute to show a confirmation or not without having to make any other changes to their block / workflow.

Since some blocks are used both internally and publicly, would it be better to have a block setting for requiring email confirmation as opposed to a global one? That would allow the user more flexibility to decide which email inputs need confirmation.

With regards to layout, most of the blocks I looked at looked fine having the two fields stacked. It may be nice to add a property to display the fields side-by-side (`.col-md-6`) for the few cases where it would look better to have them on the same row. Is that OK to do from a UI perspective, or should I just always keep them full width and stacked?

I'm sure there are some additional details that I am not thinking of.

jonedmiston:
@MichaelAllen Let me discuss with Nick some of the options and we'll circle back in a couple of days.

jonedmiston:
@MichaelAllen sorry for the delay in getting back to you. Below is what we were thinking.

Below is our proposal for this feature. Feel free to comment.
In general, we believe that for as many people would like this feature to improve data quality there would be equal amount who would not want it for adding friction to the UX. Therefore, we want to ensure that both camps can have their cake and eat it too. 

First, we see adding a global kill switch to disable the feature at the system level. This would be a system setting in the ‘General Configuration’ under ‘Admin Tools > System Settings > System Configuration’ called ‘Disable Email Confirmation’ (help text: “This setting will disable the email confirmation feature at a system level.”). Note that this setting is not a ‘Global Enable’ switch but rather a ‘Global Disable’.
Next, the email server control would need to be updated with an additional setting ‘Enabled Confirmation’ that when enabled would show two email entries and confirm that they match on validation. This control would also consider the system setting above. If ‘Disable Email Confirmation’ were true it would ignore the setting on the control. 

The field type and attribute type would also then be updated to allow for configuration of this setting.
Finally, settings would be needed in places where an administrator would like to selectively enable the feature. This would include:

1. Setting on the Email Attribute (field type) called ‘Enable Confirmation’ (help text: “Request that the person have to re-type the email to ensure that it is correct.”)

1. A new Event Registration Person Field Type of ‘Email with Confirmation’. This would in a sense be a clone of the current Email field but would enable the confirmation option.

1. Block settings on the following blocks to “Enable Email Confirmation”:

	1. Family Pre-Registration
	2. Attendance Self Entry
	3. Connection Opportunity Sign-up
	4. Person Update Kiosk
	5. Event Calendar Item Personalized Registration
	6. Pledge Entry
	7. Transaction Entry
	8. Transaction Entry v2
	9. Group Registration
	10. Groups Simple Register
	11. Prayer Request Entry
	12. Account Entry

MichaelAllen:
Would you be open to a PR that only partially implements this feature? 

Here is what I have completed so far:
- Added global killswitch system setting
- Updated Server control with a setting to enable validation
- Updated Field type
- Updated Attribute type

I haven't added this to any core blocks yet, but it is available for use on workflow forms. I think the point I am at now would be beneficial for people that are building public-facing workflow forms.

I still plan to implement the remaining requirements eventually, but I have a lot of internal projects that are currently taking all of my time.

jonedmiston:
@MichaelAllen  might need some time to re-establish context on this after the conference.

Issue #4123, opened 2020-02-24
Twilio API Error


morganw93:
### Prerequisites
* [No because demo can't send texts ]Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
    [X] Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?
    [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ASparkDevNetwork%2FSlack) to see if your bug or enhancement is already reported?

### Description

All incoming messages are causing Twilio to throw this error: 
"Warning - 14103

Message Invalid Body

The Message body does not contain valid text or a media URL"

Twilio support says this is because our API is not reacting as intended. 

Despite this error everything works in Rock as intended. 

But we have thousands of bugs in Twilio. 

I have attached the full error details in screenshots.

### Steps to Reproduce

1. Send a text to a number set up in Rock/Twilio with a rock web hook.

**Expected behavior:**

No errors in Twilio

**Actual behavior:**

A 14103 error is thrown.

### Versions

* **Rock Version:** v10.1
* **Client Culture Setting:** [En-us]

morganw93:
![Screen Shot 2020-02-24 at 10 54 27 AM](https://user-images.githubusercontent.com/30188136/75173181-1b678700-56f4-11ea-9aa7-894581398d2f.png)

morganw93:
![Screen Shot 2020-02-24 at 10 54 43 AM](https://user-images.githubusercontent.com/30188136/75173194-228e9500-56f4-11ea-95d9-4b83ae6ad50a.png)

morganw93:
![Screen Shot 2020-02-24 at 10 54 49 AM](https://user-images.githubusercontent.com/30188136/75173210-26bab280-56f4-11ea-91b2-d435c857d799.png)

nairdo:
@morganw93  I believe that bug was already smashed by the Pull Request from @MichaelAllen (https://github.com/SparkDevNetwork/Rock/pull/4107) a week or so ago.  Unfortunately, I don't think anyone had previously opened an official issue (until yours) so you would not have seen it when doing an Issue Search.  So, unless I'm mistaken, I'll close this out now. :)

PR #4554, opened 2021-01-19
Update Twilio webhooks to avoid "Invalid Body" warnings from Twilio

MichaelAllen:
## Proposed Changes

This commit (https://github.com/SparkDevNetwork/Rock/commit/35415fe4a45612ae8ad4b5288e1fbfe2a77249e5) reintroduced an old bug that causes Twilio to log an error every time a message Is received if Rock doesn't send an immediate reply. Messages still send and receive fine, but the constant errors logged in Twilio console make troubleshooting anything else more difficult.

Here is the full text of the error that gets logged:
```
Warning - 14103
Message Invalid Body
The Message body does not contain valid text or a media URL
```

According to Twilio's docs, they expect an empty response element (`<Response />`) from the webhook when there is no response to send, but Rock was sending an empty message body instead (`<Response><Message /></Response>`). 

This PR updates both Twilio webhooks to return null instead of an empty message object, when there is nothing to send. We have been running this change in production since Jan 19th, and haven't noticed any issues.

Fixes: #4123 

## Types of changes

## Checklist

## Further comments
N/A

## Documentation
N/A

## Migrations
N/A

### DIFF
--- a/RockWeb/Webhooks/Twilio.ashx
+++ b/RockWeb/Webhooks/Twilio.ashx
@@ -81,6 +81,11 @@ class TwilioResponseAsync : TwilioDefaultResponseAsync
 
         new Rock.Communication.Medium.Sms().ProcessResponse( toPhone, fromPhone, body, out errorMessage );
 
+        if ( errorMessage.IsNullOrWhiteSpace() )
+        {
+            return null;
+        }
+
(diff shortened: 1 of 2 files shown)


nairdo:
@MichaelAllen  We're looking into this now -- and thanks for catching this.

PR #4596, opened 2021-02-12
Wumc 1.10.3 pastoral care plugin feature request

rutledgek:
Added to the Pastoral Care Plugin care item detail block attributes to set care item to current time and allow requestor self selection in the person picker.


rutledgek:
Wrong Branch

PR #4886, opened 2022-01-24
Modification to allow for virtual directories on themes

cfunkdev:
## Notice

## Proposed Changes
In our multi-instance environment, SECC needed to be to be able to have a single source for our themes instead of having to copy them to each of our 6 Rock Instances.  To do so we implemented the Themes folder as a virtual directory, and had our theme assets in a file share so that we could pull from that single source.  In order to do that we had to update line 35 of RockTheme.cs to use a MapPath.  This MapPath functionality will work in single Rock Instance implementations and in multi instance/multi server implementations.  

We have been running this current version of this update since August 2021 (with Rock 12.6) and a previous version of it in Rock 10.2

## Checklist

- [X] This is a single-commit PR. (If not, please squash your commit and re-submit it.)
- [X] I have read the [Contributing to Rock](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc
- [X] Unit tests pass locally with my changes
- [X] I have added any required [unit tests](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests.UnitTests/README.md) or [integration tests](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests.Integration/README.md) that prove my fix is effective or that my feature works
- [X] I have included updated language for the [Rock Documentation](https://www.rockrms.com/Learn/Documentation) (if appropriate)

## Migrations
No migration necessary.

### DIFF
--- a/Rock/Web/UI/RockTheme.cs
+++ b/Rock/Web/UI/RockTheme.cs
@@ -32,7 +32,7 @@ namespace Rock.Web.UI
     /// </summary>
     public class RockTheme
     {
-        static private string _themeDirectory = HttpRuntime.AppDomainAppPath + "Themes";
+        static private string _themeDirectory = System.Web.Hosting.HostingEnvironment.MapPath( "~/Themes" );
 
         /// <summary>
         /// Gets or sets the name.


crayzd92 (approved):

crayzd92:
Manually picked change in v13.3 via https://github.com/SparkDevNetwork/Rock/commit/21807cde1fc49447e19d4fd728b12f96dbd8ac31

Issue #5127, opened 2022-08-21
Campus Schedule Type dropdown not loading options


MichaelAllen:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
  * Did you perform a search at https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ARock to see if your bug or enhancement is already reported?
  * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
  * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?

#### A Picture Is worth a Thousand Words
![DefinedValue](https://user-images.githubusercontent.com/282879/185811046-63eef8c4-0d05-46af-90a7-96f91b9f2ebd.png)
![CampusEdit](https://user-images.githubusercontent.com/282879/185811047-3642cd4f-e4dd-4699-a0fd-283624cd08e5.png)

### Description
The campus editor block is not correctly loading the values for the defined values picker for campus schedule types. I am only seeing the default "Weekend Service" even though I've added an additional option to the defined type.

### Steps to Reproduce
1. Add a new item to the "Schedule Type" defined type
1. Edit a campus under `General Settings > Campuses`
1. Add a new item to the "Campus Schedules" list
1. Try to select your newly added schedule type

**Expected behavior:**

The "Schedule Type" dropdown correctly loads the values from the defined type.

**Actual behavior:**

No values are loaded. Only the default value is present in the list.

### Proposed Solution
It looks like the pickers are loaded in `LoadDropDowns` which is called by `ShowEditDetails`but the picker isn't on the screen at that point.

I think that it needs to instead be loaded in `ShowCampusScheduleEdit`. Similar to how the topics are loaded in `ShowCampusTopicEdit`

Schedule:
https://github.com/SparkDevNetwork/Rock/blob/f317c12f0944bdb79f694e1e2fa940579a3f0ac2/RockWeb/Blocks/Core/CampusDetail.ascx.cs#L1027-L1030

Topic:
https://github.com/SparkDevNetwork/Rock/blob/f317c12f0944bdb79f694e1e2fa940579a3f0ac2/RockWeb/Blocks/Core/CampusDetail.ascx.cs#L1050-L1054

I'll test this proposed fix this afternoon and send a PR if it works.

### Versions
* **Rock Version:** v1.13.6
[...]

PR #5141, opened 2022-08-24
Fixed Schedule Type Dropdown not loading values on Campus Detail Block

MichaelAllen:
## Notice

## Proposed Changes

Fixes: #5127 

The campus editor block was not correctly loading the values for the defined values picker for campus schedule types. The only option available was the default "Weekend Service" even though there were additional defined values.

The issue was because the pickers are assigned a DefinedTypeId in `LoadDropDowns` which is called by `ShowEditDetails` but the Schedule Type picker is in a modal that isn't on the screen at that point. When it gets added to the screen later, it didn't have a DefinedTypeId assigned and only showed the default value as an option.

I added assigning the DefinedTypeId to `ShowCampusScheduleEdit`; similar to how the topics were being loaded for `ShowCampusTopicEdit`. That causes it to load the values for the picker after it gets added to the screen.

## Types of changes

## Checklist

## Migrations

### DIFF
--- a/RockWeb/Blocks/Core/CampusDetail.ascx.cs
+++ b/RockWeb/Blocks/Core/CampusDetail.ascx.cs
@@ -1026,6 +1026,7 @@ namespace RockWeb.Blocks.Core
         private void ShowCampusScheduleEdit( Guid campusScheduleGuid )
         {
             hfCampusScheduleGuid.Value = campusScheduleGuid.ToString();
+            dvpScheduleType.DefinedTypeId = DefinedTypeCache.Get(Rock.SystemGuid.DefinedType.SCHEDULE_TYPE.AsGuid()).Id;
             dlgSchedule.Show();
 
             if ( campusScheduleGuid == Guid.Empty )


nairdo:
@MichaelAllen  Thanks for the fix.  I'll make sure you get the PR credit in our community points.  

Also...  although with your fix now you can choose your own  DefinedType DefinedValue, the main block that uses campus-schedules ([Weekly Metrics](https://prealpha.rocksolidchurchdemo.com/intranet/tools/weekly-metrics)) really does not do anything with the CampusSchedule.ScheduleTypeValueId yet.  That block really only has a block setting that allows the admin to filter out all but the configured campus-schedules:

image

MichaelAllen:
@nairdo I can add a feature to that block to let you select the "Schedule Type" to use for filtering the available schedules. 

Here is what I am thinking:

1) Rename "Filter Schedules By Campus" to "Limit Schedule Selection to Campus Schedules". 
Help Text: "When enabled, this would limit the schedules shown to only those assigned to the selected campus' 'Campus Schedules'."
(To better clarify what it does, and to match with the campus teams option)

2) Add a new multi-select: "Campus Schedule Types"
Help Text: "When 'Limit Schedule Selection to Campus Schedules' is enabled, only campus schedules of the selected types will be shown."

Issue #5399, opened 2023-04-14
Obsidian Registration Pre/Post HTML is not Hidden when Control is


Jon-Corey:
## Please go through all the tasks below

* [x] Check this box only after you have successfully completed both the above tasks

## Please provide a brief description of the problem. Please do not forget to attach the relevant screenshots from your side.
Attributes in registrations can have Pre-HTML and Post-HTML. Attributes can also be filtered so that they only show on the form when a condition is met. The Obsidian Registration Entry block does not hide the attribute's Pre/Post HTML when the attribute is hidden. It always shows the Pre/Post HTML. This is inconsistent because both the C# Registration Entry block and the C# Workflow Entry block hide the attribute's Pre/Post HTML when the attribute is hidden.Pre and Post HTML on the Obsidian Registration Entry block (the Pre/Post HTML is shown even though the attribute they are attached to is hidden):

![PrePostHtmlObsidianBlock](https://user-images.githubusercontent.com/110615344/231819542-bdcb294e-8a3f-4327-9264-bea807e50ef8.png)Pre and Post HTML on the C# Registration Entry block for comparison (the Pre/Post HTML is hidden because the attribute they are attached to is hidden):

![PrePostHtmlC#Block](https://user-images.githubusercontent.com/110615344/231819526-57555801-f1c7-4a92-9f1c-a567d0f3ddd1.png)## Expected Behavior
On the Obsidian Registration Entry block, the visiblility of an attribute's Pre/Post HTML should be dependent on the visibility of the attribute itself.## Actual Behavior
On the Obsidian Registration Entry block, all attributes' Pre/Post HTML is always visible, regardless of if the attribute they are attached to is visible.## Steps to Reproduce

1. Create a registration template
1. On the first form of that template, add a registrant attribute with the following configurations:
  * Field Type: Boolean
1. Add another registrant attribute with the following configurations:
  * Field Type: Text
  * Pre-HTML: `<h3>Pre HTML</h3>`
  * Post-HTML: `<h3>Post HTML</h3>`
1. On the text attribute, add a field filter that shows the Text field if the Boolean field is "Yes"
1. Create an instance of the registration
1. Go to a page with the Obsidian Registration Entry block with the registration instance id in the URL as a page parameter
1. Select any number of registrants and hit "Next"
1. Notice how the "Pre HTML" heading and the "Post HTML" heading are shown even when the Text field is not

## Rock Version
[...]

PR #5402, opened 2023-04-14
Fixed an issue in Obsidian Registration Entry where Pre/Post HTML was not hidden when filtered

Jon-Corey:
## Notice

## Proposed Changes

This fixes an issue where the Obsidian Registration Entry block was not hiding Pre/Post HTML when its associated attribute was hidden. This is inconsistent with both the C# Registration Entry block and the C# Workflow Entry block. 

This fixes this issue by only sending Pre/Post HTML to the ItemsWithPreAndPostHtml component if the field is visible. If the field is not visible, the Registrant component gives the ItemsWithPreAndPostHtml component empty strings for the Pre/Post HTML.

Screenshot of before this fix (the filtered (hidden) attribute's Pre/Post HTML is shown, even though the attribute is hidden):
PrePostHtmlObsidianBlock

Screenshot of after this fix (the filtered (hidden) attribute's Pre/Post HTML is hidden because the attribute is hidden):
PrePostHtmlFixed

Fixes: #5399 

## Types of changes

## Checklist

## Migrations


nairdo:
> I know Spark's guidelines require a single-commit PR, so should I close this PR and create a new one with a single commit that includes the above changes plus this new bugfix? Or should I amend this commit and force-push to this branch?

@Jon-Corey  Since GitHub makes 'squash merges' quite simple now, you can just add that additional commit to this PR and then we'll eval it.  [Note: _Since we're still adjusting the Obsidian pattern in Rock, I still can't promise we would be able to accept this PR  (or possibly without additional changes) but we will examine it closely once you've submitted the fix._]

Issue #5430, opened 2023-05-02
Obsidian Registration Expired Discount Codes Throwing Exception


Jon-Corey:
## Please go through all the tasks below

* [x] Check this box only after you have successfully completed both the above tasks

## Please provide a brief description of the problem. Please do not forget to attach the relevant screenshots from your side.
When creating registrations for camps, we often have a series of discount codes for early registrations that expire before we expect final payment. On these registrations we do not require the total cost to be payed as soon as the registration is completed. This means that we often have people that used a discount code on their registration go back to make an additional payment on their registration after the discount code has expired.Previously, on the C# block, it would allow them to make the additional payment and it would keep the existing discount, even though the discount code is now expired. This is the expected result.The Obsidian block does not seem to have anticipated needing to be able to use expired discount codes and so when someone goes to make a payment on an existing registration that has an expired discount code, the block throws an unhandled NullReferenceException and redirects the user to an error page.The error page that is shown:

![ObsidianRegistrationExpiredDiscountCodesThrowingExceptionError](https://user-images.githubusercontent.com/110615344/235771830-4ab9271d-0518-4674-9d8c-44349b675a95.png)## Expected Behavior
Trying to make a payment on a registration with a discount code that was valid at the time, but has since expired should present the user with a payment screen that allows them to make additional payments. The existing discount code should still apply, even if it has expired since the registration was initially filled out.## Actual Behavior
Trying to make a payment on a registration with a discount code that was valid at the time, but has since expired presents the user with an error page.## Steps to Reproduce

1. Create a registration template
1. In the Details section, set the Financial Gateway to the "Test Gateway"
1. In the Details section, set the Cost to $10 and set the Minimum Initial Payment to $1
1. In the Discounts section, create a new discount with the following configuration:
 - Discount Code: "test"
 - Discount Percentage: 50%
 - Effective Dates:
 - Start: Any time in the past
 - End: Any time in the future
1. Create an instance of the registration
1. Go to a page with the Obsidian Registration Entry block with the registration instance id in the URL as a page parameter
1. Select any number of registrants and hit "Next"
1. Fill in the First Name and Last Name fields and hit "Next"
1. Fill in the Discount Code field with "test" and hit "Apply" to apply the "test" discount
1. Fill in the Amount To Pay Today field with "$1" and hit "Next"
1. Fill in the Credit Card field with "0000 0000 0000 0000"
[...]

PR #5431, opened 2023-05-02
Fixed an issue in Obsidian Registration Entry where expired discount codes were throwing exception

Jon-Corey:
## Notice

## Proposed Changes

This change fixes an issue in the Obsidian Registration Entry block where when a user goes to make a payment on an existing registration that has a discount code that was valid at the time, but has since expired, the block throws an unhandled NullReferenceException.

This fixes the issue by disabling the discount code validator when the registration is an existing registration and the discount code was already filled in. This is the same approach that the C# block takes.

Screenshot of before this fix (an unhandled NullReferenceException is thrown when trying to make a payment on a registration that uses an expired discount code):

ObsidianRegistrationExpiredDiscountCodesThrowingExceptionError

Screenshot of after this fix (the block doesn't validate the existing discount code and allows the user to make a payment):

ObsidianRegistrationExpiredDiscountCodesThrowingExceptionFix

Fixes: #5430 

## Types of changes

## Checklist

## Migrations

### DIFF
--- a/Rock.Blocks/Event/RegistrationEntry.cs
+++ b/Rock.Blocks/Event/RegistrationEntry.cs
@@ -198,9 +198,10 @@ namespace Rock.Blocks.Event
         /// Checks the discount code provided. If a null/blank string is used then checks for AutoApplied discounts.
         /// </summary>
         /// <param name="code">The code.</param>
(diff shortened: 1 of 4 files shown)


ethan-sparkdevnetwork:
@Jon-Corey I did give you the PR points for this one. Thanks for your hard work!

Issue #5446, opened 2023-05-16
Check-in Error When Clicking the "Back" Button After Not Being Able to Check In


Jon-Corey:
### Please go through all the tasks below

- [X] Check this box only after you have successfully completed both the above tasks

### Please provide a brief description of the problem. Please do not forget to attach the relevant screenshots from your side.

When a person is unable to check in, but tries to anyway, they reach a screen that says "Sorry, there are currently not any available options to check into." (or whatever custom messaging was added to the Ability Level Select block). If they click the "Back" button while on this screen, a NullReferenceException is caused and the user is shown an error page.

This seems to be caused by people that can't check in being removed from the CurrentCheckInState.CheckIn.Families.People list. When going to the previous page, the Time Select  block tries to access the contents of this list without checking if it is null first.

Here is the screen that has the "Back" button that causes the issue:
![The page that tells the user that they cannot check in](https://github.com/SparkDevNetwork/Rock/assets/110615344/a00c44e8-59ec-4e86-9f60-4eef11f530c2)

Here is the error screen that happens after clicking the "Back" button pictured above:
![CheckInBackButtonErrorPage](https://github.com/SparkDevNetwork/Rock/assets/110615344/d5489944-acd0-43bf-9712-b97004cf16ab)

### Expected Behavior

When a person tries to check in, but cannot check in, either:

1. The "Back" button should be hidden if it will break stuff when it is clicked.
2. The "Back" button should go to the previous page without causing an error.

### Actual Behavior

When a person tries to check in, but cannot check in, the "Back" button causes an error page to be shown.

### Steps to Reproduce

1. Configure a check-in configuration so that it has a currently active schedule and so that it prevents duplicate check-in
2. Check someone in
3. Check that person in again to get to the screen that says "Sorry, there are currently not any available options to check into."
4. Click the "Back" button
5. Notice the NullReferenceException at RockWeb.Blocks.CheckIn.TimeSelect.OnLoad

### Rock Version

14.3 (Although this is also an issue in 15)

### Client Culture Setting

en-US

ethan-sparkdevnetwork:
@Jon-Corey AbilityLevelSelect has already been fixed in commit 6d6e4967. Your fix for TimeSelect looks good. Can you create another PR for branch "hotfix-1.14.3" with just that fix? We are going to be building 14.3 and 15.1 alpha very soon. Thanks for your help!

Jon-Corey:
[...]

PR #5447, opened 2023-05-16
Fixed an issue where clicking the back button was throwing an exception

Jon-Corey:
## Notice

## Proposed Changes

This change fixes an issue with check-in where a NullReferenceException is thrown and an error page is shown when a user is unable to checkin and clicks the "Back" button that is displayed on that page.

This change fixes that issue by doing two things. First, this change hides the "Back" button on the Ability Level Select block when the user is not able to check in. Second, this change also adds a null check to the Time Select block so that if somehow someone gets there with a null People list, the block will not break.

The following screenshot is before the change. Clicking the "Back" button would show an error page.
A page that tells the user that they cannot check in with a "Back" button

The following screenshot is after the change. Notice that there is no longer a back button when the "Sorry, there are currently not any available options to check into." message is shown.
A page that tells the user that they cannot check in without a "Back" button

Fixes: #5446

## Types of changes

## Checklist

## Migrations


Jon-Corey:
This Pull Request contains a fix that had already been implemented by another commit. Another Pull Request has been created that omits this duplication here: #5452.

Issue #5465, opened 2023-06-06
v1.15.1.1 Alpha - Tooltip Typo on Family Attendance Badge Setting


mikedotmundy:
## Please go through all the tasks below

* [x] Check this box only after you have successfully completed both the above tasks

## Please provide a brief description of the problem. Please do not forget to attach the relevant screenshots from your side.
The new `Adult Attendance Type` setting for the Family Attendance Badge has 2 typos in it.Currently reads: _Determine if the individual attendance be shown instead of..._

Should read: _Determine**s** if the individual attendance **should** be shown instead of..._![Screenshot 2023-06-06 at 08 58 40](https://github.com/SparkDevNetwork/Rock/assets/23226974/24beacad-d800-4626-b635-305cbb110777)## Expected Behavior

1. Change _Determin_ to _Determines_
1. Add the word _should_

## Actual Behavior
Typos## Steps to Reproduce
See above## Rock Version
v15.1, v16.0## Client Culture Setting
en-US

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1204827440121158">image.png</a>

nairdo:
via https://github.com/SparkDevNetwork/Rock/commit/24059794acc3f02ebc6ac2d83882b4428cab819d and https://github.com/SparkDevNetwork/Rock/commit/131f2ce0663024988df1ff13f58c61dc9189ce8e

PR #5466, opened 2023-06-06
Fix misspelling and added missing word for Adult Attendance Type

SkellyRen:
Fixed Adult Attendance Type Description (Fixes [#5465](https://github.com/SparkDevNetwork/Rock/issues/5465))

## Notice

## Migrations


nairdo:
Hi @SkellyRen  Sorry about this -- when we saw that 'typo' in the new setting, we realized the text was _very_ different from what was written in the specification.  Because of that, we ended up correcting it in a different way than this Pull Request did -- (again, so we could follow more explicitly what was in our original design documents) via https://github.com/SparkDevNetwork/Rock/commit/24059794acc3f02ebc6ac2d83882b4428cab819d and https://github.com/SparkDevNetwork/Rock/commit/131f2ce0663024988df1ff13f58c61dc9189ce8e

So, I'm sorry we won't be able to use your PR on this one, but thank you again and please keep helping make Rock better with us!

Issue #5480, opened 2023-06-14
Typo in tooltip for Giving Alert Launch Workflow of Type setting


SkellyRen:
### Please go through all the tasks below

- [X] Check this box only after you have successfully completed both the above tasks

### Please provide a brief description of the problem. Please do not forget to attach the relevant screenshots from your side.

In the Tooltip for the Giving Alert Launch Workflow of Type it reads "...setting the attribute with key of 'FinancialTransactionId' _**is**_ it exists.'

![image](https://github.com/SparkDevNetwork/Rock/assets/48486355/07b743f1-4a58-4520-b24c-5eb6d258ca17)

### Expected Behavior

The tooltip should read "...setting the attribute with key of 'FinancialTransactionId' _**if**_ it exists.

### Actual Behavior

See above

### Steps to Reproduce

- Go to 'Giving Alerts'
- Click on 'Setting Cog'
- Scroll down to 'Giving Alerts section'
- Click on '+ button to add new Giving Alert'
- Scroll down to 'Alert Actions' and hover over tooltip for 'Launch Workflow of Type'

### Rock Version

Tested on v14.2 and v16.0 pre alpha site 

### Client Culture Setting

en-US

PR #5481, opened 2023-06-14
Fixed Typo for Launch Workflow of Type Tooltip

SkellyRen:
## Notice

Fixed Typo in tooltip for Giving Alert Launch Workflow of Type setting #5480 

## Migrations


nairdo:
@SkellyRen Thanks for jumping in there to help make Rock better!  Before I did the squash merge, I also slightly adjusted your commit message so that it would refer back to the underlying reported issue as `Fixes #5480` so it would auto close that issue with a reference in the comments.

PR #5493, opened 2023-06-23
Fixed an issue in Obsidian Registration Entry where Pre/Post HTML was not hidden when filtered

Jon-Corey:
## Notice

## Proposed Changes

This fixes an issue where the Obsidian Registration Entry block was not hiding Pre/Post HTML when its associated attribute was hidden. This is inconsistent with both the C# Registration Entry block and the C# Workflow Entry block. 

This fixes this issue by only sending Pre/Post HTML to the ItemsWithPreAndPostHtml component if the field is visible. If the field is not visible, the Registrant component gives the ItemsWithPreAndPostHtml component empty strings for the Pre/Post HTML.

Screenshot of before this fix (the filtered (hidden) attribute's Pre/Post HTML is shown, even though the attribute is hidden):
PrePostHtmlObsidianBlock

Screenshot of after this fix (the filtered (hidden) attribute's Pre/Post HTML is hidden because the attribute is hidden):
PrePostHtmlFixed

Fixes: #5399 

## Types of changes

## Checklist

## Further comments

This Pull Request is a duplicate of PR #5402. This one was created at the request of @nairdo in order to target v1.15.2 instead of v1.14.3. This Pull Request also includes the page scrolling bugifx mentioned in that PR.

## Migrations

### DIFF
--- a/Rock.JavaScript.Obsidian.Blocks/src/Event/RegistrationEntry/registrant.partial.ts
+++ b/Rock.JavaScript.Obsidian.Blocks/src/Event/RegistrationEntry/registrant.partial.ts
@@ -26,7 +26,7 @@ import RockButton from "@Obsidian/Controls/rockButton";
 import RegistrantPersonField from "./registrantPersonField.partial";
 import RegistrantAttributeField from "./registrantAttributeField.partial";
 import NotificationBox from "@Obsidian/Controls/notificationBox.obs";
-import { RegistrantInfo, RegistrantsSameFamily, RegistrationEntryBlockFamilyMemberViewModel, RegistrationEntryBlockFormFieldViewModel, RegistrationEntryBlockFormViewModel, RegistrationEntryBlockViewModel, RegistrationFieldSource, RegistrationEntryState, RegistrationEntryBlockArgs } from "./types.partial";
+import { RegistrantInfo, RegistrantsSameFamily, RegistrationEntryBlockFamilyMemberViewModel, RegistrationEntryBlockFormFieldViewModel, RegistrationEntryBlockFormFieldRuleViewModel, RegistrationEntryBlockFormViewModel, RegistrationEntryBlockViewModel, RegistrationFieldSource, RegistrationEntryState, RegistrationEntryBlockArgs } from "./types.partial";
 import { areEqual, newGuid } from "@Obsidian/Utility/guid";
 import RockForm from "@Obsidian/Controls/rockForm";
 import FeeField from "./feeField.partial";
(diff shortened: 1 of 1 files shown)


joezimjs:
@Jon-Corey 
Thanks for working on this. The changes you made are close. I see you noticed some scroll jank when updating a value than can affect the visibility of other fields because you tried to work around it with some scroll trickery. Sadly, that workaround isn't enough. Check out my example:

chrome_Yw4GGsrpd5

My "Cabin Preference" is based on the value of "Leader Preference", which is a text field, not a toggle or anything you can simply click on to change the value. As you can see in the video below, after every keypress, the field gets unfocused.

https://github.com/SparkDevNetwork/Rock/assets/1313898/c61cc1bd-2a24-4b10-8523-f98f4c83d218

In addition, if you did this with a dropdown or toggle like you did and you're navigating through the form with a keyboard you're still losing the focus on that field and when you hit `tab` to go to the next field, it'll actually focus the first field in the form.

So, do you wanna take another swing at finding a way to fix this? If, not, we'll make sure it gets in our backlog to work through it some more. If you do decide to take on the challenge, please remember that there are still a lot of Obsidian architectural questions that haven't been answered and this could be changing before Obsidian is ready to go live, so don't make any big architectural changes or anything. Thanks!

Jon-Corey:
@joezimjs

Thanks for catching that! From what I can tell, I would not be able to fix that focus problem without either a really janky solution or a large architectural change.

The problem happens when I change the value of the `ItemsWithPreAndPostHtml` component's `items` at runtime. This triggers the `ItemsWithPreAndPostHtml` component to re-render all of the controls by deleting all existing controls and re-creating them. This causes the page to scroll to the top and for focus to be lost (since the focused element was deleted).

I worked around the scroll issue by saving the scroll position before the re-render and then setting the scroll position to that saved value after the re-render.

I didn't notice the focus issue until you pointed it out, but I've just looked at doing something similar to the scroll where I save what field is currently focused and then just set that field as focused after the re-render. The problem with that is that since all of the fields are deleted and re-created, the element that was focused no longer exists. I don't see a good way to tie the old element to the new one.

Unless you have an idea for how to track the focused HTML element through the re-render for me to try, I will have to let this go into the backlog.

Thanks for your help!

joezimjs:
Thanks for looking into that more. That's what I was assuming was the problem, but I hadn't had much time to investigate more deeply. We've been deliberating about what to do with `ItemsWithPreAndPostHtml` because it's kinda hacked together and requires the runtime compiler to be loaded, which we're hoping to remove.

Thanks again for your help. We'll take it from here.

frogydiak:
I think it is related to this issue but this only mentioned the pre and post HTML. Please try to logout of the external site if you are logged in. You'll notice that the attribute (field) that is default to show up will not appear. It will only show if you are logged in.

joezimjs:
@frogydiak That is an unrelated issue. Can you please create a new issue?

frogydiak:
@joezimjs Just posted a [new issue](https://github.com/SparkDevNetwork/Rock/issues/5547) as you asked.

Issue #5249, opened 2022-12-14
Changes to included .lava files aren't picked up without a global cache clear (Fluid engine)


MichaelAllen:
### Prerequisites ###

* [x] Put an X between the brackets on this line if you have done all of the following:
  * Did you perform a search at https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ARock to see if your bug or enhancement is already reported?
  * Can you reproduce the problem on a fresh install or the [demo site](http://rock.rocksolidchurchdemo.com/)?
  * Did you include your Rock version number and [client culture](https://github.com/SparkDevNetwork/Rock/wiki/Environment-and-Diagnostics-Information) setting?

### Description ###

When referencing a lava file using an `{% include %}` tag, changes to the file are not reflected on the page unless I clear the global cache. Changing the block/page cache settings doesn't change the outcome, and the "red running man" icon that bypasses most other caching also doesn't seem to have an effect on this.

### Steps to Reproduce ###

1. [Set Lava engine to "Fluid"]
1. Create a file `~~/Assets/Lava/test.lava` with the contents "Hello"
1. Add an HTML block to a page with these contents: `{% include '~~/Assets/Lava/test.lava' %}`
1. Notice that the page shows "Hello"
1. Edit the file and change the contents to "Goodbye"
1. Notice that the page still shows "Hello"
1. Clear the global cache
1. Notice that the page now (correctly) shows "Goodbye"

**Expected behavior:**

If the cache is bypassed (red running man icon), the page should immediately reflect any changes made to an included lava file.

If the cache is enabled (green running man icon), I'm not sure if I should expect external changes to be automatically picked up immediately, or if it would be equivalent to making a change with SQL where Rock doesn't know that it happened without a manual cache clear.

**Actual behavior:**

The lava template cache prevents the changes from being picked up without a global cache clear regardless of the cache settings.

Note: This template cache is also not clearable from the cache manager. (See #5250)

### Versions ###

* **Rock Version:** 1.13.7 and 1.15.0.8 (pre-alpha)
* **Client Culture Setting:** en-US

frogydiak:
Hello! Any update on this issue? We're on v15.3 and clearing the Global Cache every time I made a change in a lava file is making development slower. 

Dev team, please please include a fix on this if possible with v15.4 or 16.2. Pretty please!

jonedmiston:
[...]

PR #5502, opened 2023-07-05
Update WebsiteLavaTemplateCache to respect the Cache Control setting on the admin toolbar

MichaelAllen:
## Notice

## Proposed Changes

Fixes: #5249 

Currently, when you are trying to change the contents of a lava include file (using Magnus or editing the file directly on the server), the changes can take several minutes to show up unless you manually clear the cache. This is because the WebsiteLavaTemplateCache is [using a hash of the content as the cache key](https://github.com/SparkDevNetwork/Rock/blob/6176894de4a2ed5495fad3efa22036617b908803/Rock/Lava/WebsiteLavaTemplateCacheService.cs#L254). 

Normally this is fine because a change to the lava would invalidate the cache. However, with an `{% include %}` tag, there is no change to the lava that is being processed (It's hashing the include statement, not the contents of the included file).

In this PR, I've modified the WebsiteLavaTemplateCache to look at the "cache bypass" setting similar to how this setting affects the RockCache. This will allow an admin to bypass the template cache while developing.

## Types of changes

## Checklist

## Further comments

I originally wanted to find a way to have the cache look at the contents of the included file. However, the processing for the `{% include %}` tag is implemented in the Fluid library. 

Without making changes to the Fluid library, the best solution I could find was to bypass the lava template caching when the cache bypass is enabled.

I copied basically the same logic that was used in RockCache to bypass the cache: https://github.com/SparkDevNetwork/Rock/blob/7e8513ff4d670e8640d1ec215068cfc630fda8c1/Rock/Web/Cache/RockCache.cs#L330-L337

## Migrations

### DIFF
--- a/Rock/Lava/WebsiteLavaTemplateCache.cs
+++ b/Rock/Lava/WebsiteLavaTemplateCache.cs
@@ -14,6 +14,8 @@
 // limitations under the License.
 // </copyright>
 //
+using System;
+using System.Collections.Generic;
 using Rock.Web.Cache;
 
 namespace Rock.Lava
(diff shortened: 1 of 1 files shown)


MrUpsideDown:
@MichaelAllen - thanks for taking the time to submit this PR. I've modified your implementation in the final commit to consolidate the caching logic in a single component, but the end result is the same.
As you have noted, there is an issue in the Fluid library that is the underlying cause of this caching problem with the `{% include %}` tag. Your change will be a great help to those who need an immediate workaround while developing new Lava templates. We are working on a comprehensive fix for #5249 that should be available in a later update.

pbassham:
Would love to see this in 15.2 if possible.

Issue #5669, opened 2023-11-09
Member Added to Group Workflow is Sometimes Missing Entity


Jon-Corey:
### Description

When a group member is added to a group from the group member detail block, sometimes any "Member Added to Group" workflows will run before the group member has been added to the database. This means that the workflows will not be passed the group member as the entity (the entity will be null).

Whether or not the workflows fire before the group member is added seems to be affected by the number of group member attributes and group requirements. The more attributes and requirements, the more likely the workflows will run before the group member is saved to the database.

### Actual Behavior

"Member Added to Group" group member workflows only sometimes have the group member as the entity passed to the workflow.

### Expected Behavior

"Member Added to Group" group member workflows will always have the group member as the entity passed to the workflow.

### Steps to Reproduce

This is a highly timing-dependent issue to try to reproduce. I was unable to reproduce this on the demo site, but I was able to reproduce this with a stock rock instance hooked up to a large database.

1. Get a Rock instance that is attached to a database with a somewhat large amount of data
2. Create a group with a lot of group member attributes and/or group requirements
3. Create a workflow that uses the Attribute Set From Entity action to get the group member from the entity and assign it to a group member attribute
4. Add the new workflow as a "Member Added to Group" group member workflow on the group
5. Add a few people to the group
6. Notice how some or all of the workflows don't have the group member from the entity

### Issue Confirmation

- [X] Perform a search on the [Github Issues](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ARock) to see if your bug or enhancement is already reported.
- [X] Try to reproduce the problem on a fresh install or on the [demo site](https://rock.rocksolidchurchdemo.com/).

### Rock Version

v15.1

### Client Culture Setting

en-US

PR #5670, opened 2023-11-09
- (Workflow) Fixed an issue where group member add workflows wouldn't always be passed an entity

Jon-Corey:
## Proposed Changes
This fixes an issue where "Member Added to Group" group member workflows were not guaranteed to happen after the group member was added to the database when the group member was added from the Group Member Detail block. This would lead to "Member Added to Group" group member workflows sometimes not having the group member in the entity that was passed to the workflow.This issue was caused by the Group Member Detail block wrapping the save methods in a transaction that was tasked to save the group member, then save the group member's attribute values, and then calculate the group member's requirements. If saving the attributes and calculating the requirements took too long, the group member workflows that were triggered by the save event would run before the transaction was closed.This fixes that issue by mimicking the pattern of other post-save actions and waiting for 1 second before continuing. This ensures that the transaction has enough time to complete before the entity is retrieved and the workflows are triggered.Fixes: #5669

## Types of changes

## Checklist

* [x] This is a single-commit PR. (If not, please squash your commit and re-submit it.)
* [x] I have read the [Contributing to Rock](https://github.com/SparkDevNetwork/Rock/blob/master/.github/CONTRIBUTING.md) doc
* [x] By contributing code, I agree to license my contribution under the [Rock Community License Agreement](https://www.rockrms.com/license)
* [x] I have added any required [unit tests](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests.UnitTests/README.md) or [integration tests](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests.Integration/README.md) that prove my fix is effective or that my feature works
* [x] I have included updated language for the [Rock Documentation](https://www.rockrms.com/Learn/Documentation) (if appropriate)

### DIFF
--- a/Rock/Tasks/UpdateGroupMember.cs
+++ b/Rock/Tasks/UpdateGroupMember.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.Data.Entity;
 using System.Linq;
+using System.Threading.Tasks;
 
 using Rock.Data;
 using Rock.Model;
@@ -126,7 +127,7 @@ namespace Rock.Tasks
                                             {
                                                 if ( message.State == EntityContextState.Added && QualifiersMatch( rockContext, trigger, message.GroupMemberStatus, message.GroupMemberStatus, message.GroupMemberRoleId, message.GroupMemberRoleId ) )
                                                 {
-                                                    LaunchWorkflow( rockContext, trigger, message );
+                                                    LaunchWorkflow( trigger, message );
                                                 }
 
                                                 break;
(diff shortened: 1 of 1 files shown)


PraveenMathew92:
Hi Jon,
 
Thank you for the PR and your contribution to making Rock better.
 
I see you have come up with an approach similar to what is already being done in Rock. For instance, in the same [GroupMemeberSaveHooks](https://github.com/SparkDevNetwork/Rock/blob/5c130bae49463e82b04bf672fcab126c73f595c8/Rock/Model/Group/GroupMember/GroupMember.SaveHook.cs#L295) class, we use the delay of 1 second to let the updates be committed to the database.

However, after a lot of internal discussions, we have decided to eliminate the "Delays" and have the PostSaveHooks be run only after the changes to the model have been committed to the database. We feel this would be a better way to ensure that the workflows which are run after the Model update like the Member Added to Group workflow does cause race conditions.
 
So, while we would not be able to include this PR, it did help us fix up other areas of Rock and make it more robust. Once again thank you for your contribution.

Jon-Corey:
@PraveenMathew92, that is a much better solution. Thank you for taking the time to make sure this was completed the right way.

Issue #5723, opened 2024-01-23
Workflow Forms Show Error When Double Clicking Buttons


Jon-Corey:
### Description

Workflow forms have configurable action buttons at the bottom of the form. When double-clicking any of these action buttons, the workflow advances as expected, but all the user sees is an error message. This means that if there is another form or a confirmation message, the user will not see it. They will only see the error.

Here is a screenshot where I double-clicked the Submit button. Notice that I still see the form and that I also see an error message. I do not see the action button's Response Text.

![WorkflowDoubleClickError](https://github.com/SparkDevNetwork/Rock/assets/110615344/8daedb4b-cbf1-4cef-aaf0-62c35fd1fe9e)

### Actual Behavior

Clicking an action button multiple times causes an error that prevents the user's view from advancing to the next item.

### Expected Behavior

The workflow action buttons should behave the same regardless of if they are clicked once or multiple times. After clicking an action button, the user's view should be advanced to the next item. They should not still see the form.

### Steps to Reproduce

1. Create a new workflow type
2. Add a Form action to the Start activity (Workflow Control > Form)
3. Add an additional Cancel button
4. Save the workflow
5. Click the Launch Workflow button
6. Double-click either the "Submit" or "Cancel" button
7. Notice the "Object reference not set to an instance of an object" error

### Issue Confirmation

- [X] Perform a search on the [Github Issues](https://github.com/issues?q=is%3Aissue+user%3ASparkDevNetwork+-repo%3ARock) to see if your bug or enhancement is already reported.
- [X] Try to reproduce the problem on a fresh install or on the [demo site](https://rock.rocksolidchurchdemo.com/).

### Rock Version

v15.1, v17.0

### Client Culture Setting

en-US

nairdo:
[_We are in communication with Jon Corey and also let him know about this via his Pull Request._] This is a mission critical block that is rather complex and has been in use for 8+ years.  We're going to defer this 'double click bug' to fix it in the Next Gen WorkflowEntry block.

chead4:
As mentioned above, this is a critical and complex block that will be fixed in the Next Gen Workflow Entry block.  At this time, we're going to close this issue and track internally.

PR #5724, opened 2024-01-23
Workflow - Fixed Issue Caused By Double Clicking Action Buttons

Jon-Corey:
## Proposed Changes

This fixes an issue where double-clicking the action buttons on a workflow form would cause an error. This fixes that issue by putting some client-side JavaScript on the page that disables all action buttons after one has been clicked as long as the form is valid. This prevents subsequent clicks from registering before the postback has been completed.

One result of this fix is that the Cancel button and other buttons that skip validation will need to have the `handleWorkflowCancelButtonClick()` JavaScript function in their `onclick` event instead of having "`return true;`". The new JavaScript function returns true after it handles disabling the action buttons so that they can't be clicked before the postback completes. I've put this new function in the ButtonCancel merge field so that if someone wants to change a normal button to a button that ignores validation, they just have to change "`{{ ButtonClick }}`" to "`{{ ButtonCancel }}`" in the button's HTML.

If a Rock instance ends up with "`return true;`" instead of "`{{ ButtonCancel }}`", the Cancel buttons on their workflow forms will have the same bug that they currently have. To get this fix, they will need to use "`{{ ButtonCancel }}`".

Fixes: #5723

## Types of changes

## Checklist

## Documentation

There is a line in the Blasting Off With Workflows book that will need to be updated. The section that will need updating is Working With Entry Forms > Buttons > Cancel Buttons. Here, the last sentence will need to be changed from:

> To keep the validation from occurring simply use the Cancel button type or change the {{ ButtonClick }} merge field to be "return true;".

To:

> To keep the validation from occurring simply use the _Cancel_ button type or change the {{ ButtonClick }} merge field to be "{{ ButtonCancel }}".

## Migrations

A migration will need to be built that looks at the [Button HTML defined type](https://rocksolidchurchdemo.com/admin/general/defined-types/44) and replace "`onclick="return true;"`" with "`onclick="{{ ButtonCancel }}"`" in the Button HTML field for each of the defined values.

I would be willing to write the migration for this if someone would be able to point me in the direction of some documentation for how to write migrations.

### DIFF
--- a/Rock/Model/Workflow/WorkflowActionForm/WorkflowActionFormUserAction.cs
+++ b/Rock/Model/Workflow/WorkflowActionForm/WorkflowActionFormUserAction.cs
@@ -113,7 +113,7 @@ namespace Rock.Model
 
             // Add other buttons that have been configured to disable validation.
             // In accordance with the Rock documentation, button validation can be disabled
-            // by replacing the text "{{ ButtonClick }}" with "return true;" in the Button HTML Attribute.
+            // by replacing the text "{{ ButtonClick }}" with "{{ ButtonCancel }}" in the Button HTML Attribute.
             // Add any buttons with this configuration to the collection of Cancel buttons.
             var buttonType = DefinedTypeCache.Get( Rock.SystemGuid.DefinedType.BUTTON_HTML );
 
@@ -126,7 +126,7 @@ namespace Rock.Model
                     buttonHtml = buttonHtml.ToLower().Replace( " ", "" );
 
                     if ( !buttonHtml.Contains( "{{buttonclick}}" )
-                         && buttonHtml.Contains( "onclick=\"returntrue;\"" ) )
+                         && buttonHtml.Contains( "onclick=\"{{buttoncancel}}\"" ) )
                     {
                         cancelButtonList.Add( buttonDefinedValue.Guid );
                     }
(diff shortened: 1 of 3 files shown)


nairdo:
@Jon-Corey  We've been looking at this change and the implications.  The WorkflowEntry block is quite the mission critical block, so we need to take extra care when changing its behavior.  That block has essentially been working this way for 8+ years.

Initially we though it might be safer to still work with the older `return true;` buttons by amending your change with `||` _the old way in_ ([line 129](https://github.com/SparkDevNetwork/Rock/pull/5724/commits/ff68a69094fd1ca647ffee11ece20a527cd58582#diff-b9656793fe3ba40b713b7d70902892823a563cad0672c81fbd189fbd3abbdf29R129))

`&& ( buttonHtml.Contains( "onclick=\"{{buttoncancel}}\"" ) || buttonHtml.Contains( "onclick=\"returntrue;\"" ) )`
`                                                           ^^^^^^^^^^...`
...but we've decided it would be best to defer your changes until the Next-Gen re-write of the WorkflowEntry block.  I will also be marking the corresponding issue with the `Will Fix In Next Gen` label.

Sorry about this.  We do appreciate your assistance to help making Rock better!  If you do end up using your changes in your production Rock instance, let us know how it goes.

PR #6592, opened 2025-11-22
Fix Export crash from leftover OWIN dependency

BStoneATX:
Changes to be committed:
modified: Rock/Web/UI/Controls/Grid/Grid.cs

Proposed Changes

This PR removes a leftover OWIN dependency inside the Actions_ExcelExportClick method in Rock/Web/UI/Controls/Grid/Grid.cs.

Starting in Rock McKinley v17, the OWIN hosting pipeline was removed. However, the Excel export logic still attempts to call:

Context?.GetOwinContext()?.Request?.CallCancelled

On any Rock instance running 17.6+, this results in the following runtime exception whenever a grid is exported to Excel:

InvalidOperationException: No owin.Environment item was found in the context.

This prevents all Excel exports system-wide.

Fix:
The OWIN call has been replaced with a safe fallback:

var cancellationToken = CancellationToken.None;

This removes the dependency on OWIN entirely and restores Excel export functionality on all affected grids.

Why this fix is needed:
	•	OWIN was removed in Rock 17+
	•	The old GetOwinContext() call now always throws
	•	Excel exports crash in all grids
	•	Replacing the call restores expected behavior with no side effects

Fixes: #

Types of changes
	•	Bugfix (non-breaking change which fixes an issue)
	•	New feature
	•	Breaking change

Checklist
	•	This is a single-commit PR
	•	Verified no whitespace/formatting changes
	•	I have read the Contributing to Rock documentation
	•	I agree to license my contribution under the Rock Community License Agreement
	•	Unit tests pass locally
	•	Added unit/integration tests if appropriate
	•	Documentation updated if appropriate

Further comments

This issue affects all Excel export functionality on any installation upgraded to Rock 17.6+, including Early Access builds. Since the OWIN pipeline no longer exists, any call to GetOwinContext() will always fail. This minimal fix removes the OWIN dependency and restores normal grid export behavior.

Documentation

No end-user documentation changes required.

Migrations

No migration is required.

### DIFF
--- a/Rock/Web/UI/Controls/Grid/Grid.cs
+++ b/Rock/Web/UI/Controls/Grid/Grid.cs
@@ -2184,7 +2184,7 @@ $('#{this.ClientID} .{GRID_SELECT_CELL_CSS_CLASS}').on( 'click', function (event
             // Safely retrieve the cancellation token so Rock can cancel the export if the request is aborted.
             // Checking if the request is cancelled is cheap but it makes the code less readable,
             // thus cancellation checks should only be added before performance-heavy regions.
-            var cancellationToken = Context?.GetOwinContext()?.Request?.CallCancelled ?? CancellationToken.None;
+            var cancellationToken = CancellationToken.None;
 
             // Create the excel file.
             using ( var excel = new ExcelPackage() )


cabal95:
We are going to close this PR as the issue is with Owin not working.  Owin is an integral part of Rock and IIS, finding the root issue would be better. As-is, this PR would remove functionality that is working correctly on every other Rock instance.

PR #6633, opened 2026-01-02
Improved the RenderLavaEndpoint command to parse the querystring in the route parameter

MichaelAllen:
## Notice

## Proposed Changes

When using the Render Lava Endpoint command, if you add a query string parameter to the route - it never calls the endpoint. As soon as the query string is removed - the endpoint is called.

This PR updates the RenderLavaEndpoint lava command to allow for query string parameters.

Example:
```
//- This works
{% renderlavaendpoint route:'^/group-finder/showgroups' %}

//- This does not currently work
{% renderlavaendpoint route:'^/group-finder/showgroups?GroupKey=17MNOZN7yQ' %}
```

If this PR is accepted, both of the above examples would work, and the query string parameter from the second one would be available to the lava app in the `QueryString` merge field, just as it would have been if the endpoint were called from an `hx-get` or similar tag on the page.

Fixes: https://github.com/Triumph-Tech/Triumph-Helix/issues/17

## Types of changes

## Checklist

- [ ] I have added any required [unit tests](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests/README.md) or [integration tests](https://github.com/SparkDevNetwork/Rock/blob/develop/Rock.Tests.Integration/README.md) that prove my fix is effective or that my feature works

## Further comments
[This thread in Rocket.Chat](https://chat.rockrms.com/channel/lava?msg=7gJPGE9LCRFMuxfLZ) does a good job of explaining one possible use case for this.

> In one GET endpoint (/my-app/groups), I'm trying to call another endpoint (/my-app/group). In groups endpoint, I loop on the actual set of groups, and have this in the for loop {% renderlavaendpoint route:'^/my-app/group?GroupGuid={{ groupGuid }}' %}
> 
> To not repeat code, I was hoping that the groups endpoint could loop on all the guids and use the group endpoint for each.
Doing the "call" to the group endpoint via an hx-get attribute of an element, results in the browser doing all the Ajax calls.
> 
> If, instead, the {% renderlaveendpoint %} could be used in the groups endpoint, then there is no API calls from the browser, because it would all occur internal, server-side...
> 
> If I had 100 groups to show, using the hx-get approach, results in 1+100 API calls from the browser.
If renderlavaendpoint worked [with querystring params], that is 1 call, because groups returns the content from all the group calls in one shot.

## Documentation
In [the documentation](https://community.rockrms.com/page/3761) for this lava command, we could add an example showing a querystring. Maybe something like this:

> If you include query string parameters in the route, they will be made available to your lava application in the `QueryString` merge field.
> ```
> {% renderlavaendpoint route:'^/group-finder/showgroups?GroupKey=17MNOZN7yQ' %}
> ```

## Migrations

### DIFF
--- a/Rock/Lava/Blocks/RenderLavaEndpoint.cs
+++ b/Rock/Lava/Blocks/RenderLavaEndpoint.cs
@@ -17,7 +17,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.IO;
+using System.Linq;
 using System.Web;
 
 using Rock.Cms;
@@ -63,6 +65,16 @@ namespace Rock.Lava.Blocks
             }
 
             var route = settings["route"].TrimStart( "^/".ToCharArray() );
+            NameValueCollection queryParams = null;
+
+            // Split out query string
+            var splitRoute = route.Split( new[] { '?' }, 2 );
+            if ( splitRoute.Length == 2 )
+            {
+                route = splitRoute[0];
+                queryParams = splitRoute[1].ParseQueryString();
+            }
+
             var routeComponents = route.Split( new[] { '/' }, 2 );
 
             // Check for invalid route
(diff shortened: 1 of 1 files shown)


nairdo:
This has also been tested and cherry picked into v18.2

PR #6716, opened 2026-02-27
+ (Core) Fixed steps authorization in Bulk Update to correctly inherit permissions from the parent Step Program when no explicit permissions are set on the Step Type. (Fixes #6714)

### LINKED ISSUE #6714: Cannot Add/Update Steps from Bulk Update
### Description

A user cannot add/update/remove steps from Bulk Update even though they should have permission to.

Image

Image

Image

### Actual Behavior

"You are not authorized to add Starting Point Class steps." even though the user is in a security role that has edit permission to the "Step Program". Technically if you define the security all the way at the "Step Type" level it does end up working, but inherited permissions from the "Step Program" are not working. (It has to do with using a StepTypeCache object in the bulk update block, but I am unsure if there are any further repercussions.)

### Expected Behavior

Permissions to be inherited properly on the Step Type from the Step Program and be able to add/update/remove steps from the bulk update block.

### Steps to Reproduce

1. Add "Staff Workers" security role to a Step Program's "Edit" permission. (ex. Discipleship Path)
2. Login as a user with "Staff Workers" security role, such as Pete Foster
3. Visit /BulkUpdate/0 or attempt to bulk update some users from a Grid (ex. https://rock.rocksolidchurchdemo.com/BulkUpdate/0) 
4. Expand Steps
5. Choose **Action:** Add Step, **Step Program:** Discipleship Path, **Step Type:** Baptism
6. Notice error "You are not authorized to add Baptism steps."

### Issue Confirmation

### Rock Version

17.6-18.2

### Client Culture Setting

en-US

jonedmiston:


nairdo:
This was very helpful, but in the course of testing I uncovered a related bug in the BulkUpdate block that it was not OR checking the Authorization.MANAGE_STEPS so this fix in the v18.3 branch fixes both issues. https://github.com/SparkDevNetwork/Rock/commit/148ecd1eb0a60f8a08d8a936d39a4e44de715637

PR #6720, opened 2026-03-02
Fix penny drift in fee coverage distribution across accounts

jonedmiston:
## Proposed Changes

This PR fixes a rounding/penny drift issue in the Transaction Entry V2 block's fee coverage distribution logic. When multiple accounts receive fee coverage amounts, the previous implementation rounded each account's proportional fee independently, which could cause the sum of rounded fees to differ from the target total by a penny or more.

The fix implements the **largest-remainder method** (also known as penny reconciliation) to distribute fee coverage amounts. This ensures that:
1. Each account's fee is calculated proportionally based on its gift amount
2. Per-account fees are rounded to the nearest cent
3. The sum of all per-account fees equals exactly the target total fee coverage amount

The new `DistributeFeeCoverage()` method:
- Calculates each account's proportional fee
- Floors each to whole cents and tracks fractional remainders
- Distributes remaining cents to accounts with the largest remainders (breaking ties by gift amount for deterministic results)
- Handles edge cases where rounding might result in over-allocation

## Types of changes

## Checklist

- [x] This is a single-commit PR
- [x] I verified my PR does not include whitespace/formatting changes
- [x] By contributing code, I agree to license my contribution under the Rock Community License Agreement

## Further comments

The change is backward compatible and only affects the internal calculation of per-account fee coverage amounts. The logic is deterministic and handles edge cases (zero amounts, negative remainders) appropriately.

https://claude.ai/code/session_01AovbKrSUUHrq2RKNT9aJRd

### DIFF
--- a/RockWeb/Blocks/Finance/TransactionEntryV2.ascx.cs
+++ b/RockWeb/Blocks/Finance/TransactionEntryV2.ascx.cs
@@ -3307,19 +3307,22 @@ mission. We are so grateful for your commitment.</p>
             var selectedAccountAmounts = caapPromptForAccountAmounts.AccountAmounts.Where( a => a.Amount.HasValue && a.Amount != 0 ).ToArray();
 
             var totalFeeCoverageAmount = GetSelectedFeeCoverageAmount();
-            var totalSelectedAmounts = selectedAccountAmounts.Sum( a => a.Amount.Value );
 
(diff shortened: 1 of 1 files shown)


nairdo:
@claude While this fix was good, I ended up copying it to the v18.3 branch so I could test it there.  Then I ended up just committing it straight away.  Closing as fixed in v18.3

PR #6744, opened 2026-03-20
Fix Attribute Matrix Items timeout on save

### LINKED ISSUE #6743: Attribute Matrix Items Sometimes Fail to Save
### Description

Attribute Matrix Items can sometimes fail to save due to SQL query plan regression causing a full table scan of AttributeValue.Value on save. When SQL executes a poorly optimized plan, saving the attribute matrix item fails with a timeout error.

The relevant query is shown below:

```
/*
This query text was retrieved from showplan XML, and may be truncated.
*/

SELECT TOP (1) 
    [Extent1].[Id] AS [Id], 
    [Extent1].[IsSystem] AS [IsSystem], 
    [Extent1].[AttributeId] AS [AttributeId], 
    [Extent1].[EntityId] AS [EntityId], 
    [Extent1].[Value] AS [Value], 
    [Extent1].[PersistedTextValue] AS [PersistedTextValue], 
    [Extent1].[PersistedHtmlValue] AS [PersistedHtmlValue], 
    [Extent1].[PersistedCondensedTextValue] AS [PersistedCondensedTextValue], 
    [Extent1].[PersistedCondensedHtmlValue] AS [PersistedCondensedHtmlValue], 
    [Extent1].[IsPersistedValueDirty] AS [IsPersistedValueDirty], 
    [Extent1].[ValueAsDateTime] AS [ValueAsDateTime], 
    [Extent1].[ValueAsNumeric] AS [ValueAsNumeric], 
    [Extent1].[ValueAsPersonId] AS [ValueAsPersonId], 
    [Extent1].[ValueAsBoolean] AS [ValueAsBoolean], 
    [Extent1].[ValueChecksum] AS [ValueChecksum], 
    [Extent1].[CreatedDateTime] AS [CreatedDateTime], 
    [Extent1].[ModifiedDateTime] AS [ModifiedDateTime], 
    [Extent1].[CreatedByPersonAliasId] AS [CreatedByPersonAliasId], 
    [Extent1].[ModifiedByPersonAliasId] AS [ModifiedByPersonAliasId], 
    [Extent1].[Guid] AS [Guid], 
    [Extent1].[ForeignId] AS [ForeignId], 
    [Extent1].[ForeignGuid] AS [ForeignGuid], 
    [Extent1].[ForeignKey] AS [ForeignKey]
    FROM [dbo].[AttributeValue] AS [Extent1]
    WHERE ( EXISTS (SELECT 
        1 AS [C1]
        FROM [dbo].[Attribute] AS [Extent2]
        WHERE ([Extent2].[FieldTypeId] = @p__linq__0) AND ([Extent2].[Id] = [Extent1].[AttributeId])
    )) AND ( EXISTS (SELECT 
        1 AS [C1]
        FROM [dbo].[AttributeMatrix] AS [Extent3]
        WHERE ( EXISTS (SELECT 
            1 AS [C1]
            FROM [dbo].[AttributeMatrixItem] AS [Extent4]
            WHERE ([Extent3].[Id] = [Extent4].[AttributeMatrixId]) AND ([Extent4].[Id] = @p__linq__1)
        )) AND (((LOWER( CAST( [Extent3].[Guid] AS nvarchar(max)))) = [Extent1].[Value]) OR ((LOWER( CAST( [Extent3].[Guid] AS nvarchar(max))) IS NULL) AND ([Extent1].[Value] IS NULL)))
    ))
```

This query appears to be generated by [AttributeValue.GetRootMatrixAttributeValue](https://github.com/SparkDevNetwork/Rock/blob/f1803bd60017a902fea3783b8d2e251933927a91/Rock/Model/Core/AttributeValue/AttributeValue.Logic.cs#L467).

A good query plan looks something like this, where the query leverages indexes on AttributeValue and does a key lookup back to the clustered index:

Image

However, SQL Server will sometimes regress to a query plan like the one shown below where we're doing a full table scan on AttributeValue.Value:

Image

On a busy system with many AttributeValue records (even with the database maintenance job running daily), this problem can come and go seemingly at random due to the query plan SQL chooses. Rebuilds or statistics updates  on AttributeValue table indexes can alleviate the problem temporarily but do not provide a permanent fix.

Perhaps the checksum column on AttributeValue could be leveraged in `GetRootMatrixAttributeValue` to generate a more performant query?

Note: This was not reproduced on the demo system due to the complexity of doing so, but has been observed in multiple Rock instances running v16.x and up.

### Actual Behavior

Saving attribute matrix items fails in a transient manner.

### Expected Behavior

Saving attribute matrix items works consistently.

### Steps to Reproduce

- Edit an attribute matrix instance configured as an attribute on an entity
- Make changes to attribute matrix item and click "Save"
- Sometimes, saving the attribute matrix item times out and the item fails to save. Subsequent attempts may or may not be successful. Updating statistics or rebuilding indexes on the AttributeValue table might help.

### Issue Confirmation

### Rock Version

17.6

### Client Culture Setting

en-US

jonedmiston:
## Summary
- **Fixes #6743** — Attribute Matrix Items sometimes fail to save due to SQL Server choosing a bad query plan that full-table-scans `AttributeValue.Value`.
- Adds a `ValueChecksum` pre-filter to `GetRootMatrixAttributeValue()` using `SqlFunctions.Checksum()`, allowing SQL Server to leverage the indexed `ValueChecksum` column before doing the expensive string comparison.
- The `AttributeValueExpressionVisitor` only optimizes `==`/`!=` expressions, not `.Contains()`, so this query needed the checksum filter added manually.

## Test plan
- [ ] Edit an attribute matrix item on an entity and save — verify it saves successfully
- [ ] Confirm the generated SQL includes a `ValueChecksum IN (SELECT CHECKSUM(...))` clause via SQL Profiler
- [ ] Verify no regression on systems with large `AttributeValue` tables

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Rock/Model/Core/AttributeValue/AttributeValue.Logic.cs
+++ b/Rock/Model/Core/AttributeValue/AttributeValue.Logic.cs
@@ -20,6 +20,7 @@ using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations.Schema;
 using System.Data.Entity;
 using System.Data.Entity.Infrastructure;
+using System.Data.Entity.SqlServer;
 using System.Data.SqlTypes;
 using System.Linq;
 
@@ -471,9 +472,17 @@ namespace Rock.Model
             var attributeService = new AttributeService( rockContext );
             var attributeValueService = new AttributeValueService( rockContext );
 
-            var matrixGuidQuery = attributeMatrixService.Queryable().AsNoTracking().Where( am =>
-                am.AttributeMatrixItems.Any( ami => ami.Id == EntityId ) )
-                .Select( am => am.Guid.ToString() );
+            // Get the matrix that contains the current entity.
+            var matrixQuery = attributeMatrixService.Queryable().AsNoTracking().Where( am =>
+                am.AttributeMatrixItems.Any( ami => ami.Id == EntityId ) );
+
+            var matrixGuidQuery = matrixQuery.Select( am => am.Guid.ToString() );
+
+            // Build a checksum subquery to leverage the indexed ValueChecksum
+            // column, preventing SQL Server from choosing a full table scan on
+            // AttributeValue.Value. See issue #6743.
+            var matrixGuidChecksumQuery = matrixQuery
+                .Select( am => SqlFunctions.Checksum( am.Guid.ToString() ) );
 
             var matrixFieldType = FieldTypeCache.Get( SystemGuid.FieldType.MATRIX );
             var attributeIdQuery = attributeService.Queryable().AsNoTracking().Where( a =>
@@ -481,7 +490,9 @@ namespace Rock.Model
                 .Select( a => a.Id );
 
             var attributeValue = attributeValueService.Queryable().AsNoTracking().FirstOrDefault( av =>
-                 attributeIdQuery.Contains( av.AttributeId ) && matrixGuidQuery.Contains( av.Value ) );
+                 attributeIdQuery.Contains( av.AttributeId )
+                 && matrixGuidChecksumQuery.Contains( ( int? ) av.ValueChecksum )
+                 && matrixGuidQuery.Contains( av.Value ) );
 
             return attributeValue;
         }


nairdo:
I did end up combining these two solutions and committed it to the v18.3 branch via
 https://github.com/SparkDevNetwork/Rock/commit/79c3e3f2f649fd3fa9875717e01cbe853e997cb8

PR #6968, opened 2026-08-12
Fixed Electronic Signature action erroring with "Captcha was not valid" on Obsidian Workflow Entry block (Fixes #6967)

### LINKED ISSUE #6967: Electronic Signature Action errors with "Captcha was not valid" on Obsidian Workflow Entry Block.
### Description

Submitting a signature document through the Electronic Signature workflow action fails when the Obsidian Workflow Entry block has CAPTCHA enabled. The request returns the error "Captcha was not valid." 

Image

### Actual Behavior

Unable to complete the signature document when CAPTCHA is enabled. 

### Expected Behavior

Able to complete the signature document without an error when CAPTCHA is enabled. 

### Steps to Reproduce

1. Create a workflow that prompts the current person to sign the Photo Release signature document using the Signature Document workflow action. 
2. Launch the workflow using the Obsidian Workflow Entry block and configure block where CAPTCHA is not disabled. 
3. Attempt to complete the signature document. 
4.See error. 

### Issue Confirmation

### Rock Version

v19.3 and v19.2

### Client Culture Setting

en-US

YesuCS:
## Description
When a workflow with a Signature Document (Electronic Signature) action runs on the Obsidian Workflow Entry block with CAPTCHA support enabled, completing the signature fails with "Captcha was not valid" (#6967, affects v19.2/v19.3, still present in v19.4).

The block's `GetNextInteractiveAction` action validates CAPTCHA on every invocation. The entry form component acquires a token and sends it in the block action context, but the electronic signature component submitted without one, so the server always rejected the request.

## Fix
`electronicSignature.obs` now follows the same pattern as `entryForm.obs`: it renders the Captcha control when CAPTCHA is enabled, waits for a token on mount, keeps the signature panel hidden until the captcha completes, and passes the token in the `BlockActionContextBag` when submitting the signed document.

Verified with ESLint and a full Obsidian Blocks type-check build.

### DIFF
--- a/Rock.JavaScript.Obsidian.Blocks/src/WorkFlow/WorkflowEntry/Actions/electronicSignature.obs
+++ b/Rock.JavaScript.Obsidian.Blocks/src/WorkFlow/WorkflowEntry/Actions/electronicSignature.obs
@@ -12,7 +12,11 @@
             </div>
         </div>
 
-        <div class="well">
+        <div v-if="isCaptchaEnabled" v-show="!isCaptchaComplete">
+            <Captcha ref="captchaElement"></Captcha>
+        </div>
+
+        <div v-if="!isCaptchaEnabled || isCaptchaComplete" class="well">
             <ElectronicSignature v-model="signature"
                                  :documentTerm="documentTerm"
                                  :isDrawn="isDrawn"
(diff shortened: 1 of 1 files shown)


nairdo:
@YesuCS Thanks for putting this together, the shape of the fix looks right and matches the pattern established in [2ed8857a47](https://github.com/SparkDevNetwork/Rock/commit/2ed8857a47#diff-8e23c5b77cd179eac35671d142f82e265c5b2a07f8a992df80b1ff782cffc084) from earlier this year for the entry form and the other captcha-protected blocks. We also had an internal report of the same symptom on the interactive Payment step of the Workflow Entry block earlier, which is the same root cause. Since it's an identical fix, we paired it into a follow-up commit on top of yours so both action components get patched together.

See https://github.com/SparkDevNetwork/Rock/commit/429bac2e1146d6e53c6c449d83a0fe452d2ac15a

